PR #7, opened 2021-08-25
fix: button links in L2 section

JChiaramonte7:


kmeraz:
image
do we want to change the button up or the section up once we've switched to L2?
Maybe we can change the button to instead be a button-sized box saying "You're on L2! ✅"

JChiaramonte7:
> image
> do we want to change the button up or the section up once we've switched to L2?
> Maybe we can change the button to instead be a button-sized box saying "You're on L2! ✅"

It currently switches to our "Start Trading" CTA button when you're on L2.

https://github.com/Kwenta/kwenta/pull/7/files#diff-7281795361d165520226fb3121882c778fd17cc10f68740adac5b319d01be03eR83

kmeraz (approved):
ah, i see that now. user error on my part. 🚀

PR #34, opened 2021-09-09
feat: (connector) handle unsupported networks

snxkm:
Currently if you connect to the app with an unsupported network like Matic the app will break. This handles that.

## Description
Update Connector to check for supported networks and show a `console.error`

## Motivation and Context
We want to provide a good experience to users even if they are on the wrong network.

To be more useful this probably needs more design to show a modal or similar telling the user to switch networks, and/or a a MetaMask prompt to switch to another network? This change is not included.

## How Has This Been Tested?
Manual testing with Chrome and MetaMask 10.0.3

image

### DIFF
--- a/containers/Connector/Connector.tsx
+++ b/containers/Connector/Connector.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useMemo } from 'react';
 import { createContainer } from 'unstated-next';
 import {
 	TransactionNotifier,
@@ -15,6 +15,8 @@ import {
 	synthetix,
 } from '@synthetixio/contracts-interface';
 import { ethers } from 'ethers';
+// @ts-ignore
+import snx from 'synthetix';
 
 import { ordersState } from 'store/orders';
 import { hasOrdersNotificationState } from 'store/ui';
@@ -30,7 +32,6 @@ import { LOCAL_STORAGE_KEYS } from 'constants/storage';
 import { CRYPTO_CURRENCY_MAP, CurrencyKey, ETH_ADDRESS } from 'constants/currency';
 import { synthToContractName } from 'utils/currencies';
 import { invert, keyBy } from 'lodash';
-import { useMemo } from 'react';
 
 const useConnector = () => {
 	const [network, setNetwork] = useRecoilState(networkState);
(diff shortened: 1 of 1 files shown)


asherism:
https://www.figma.com/file/sghReDviy1lcr6kkNqlc0e/Kwenta-User-Flows?node-id=5688%3A232352

Link to Figma designs for the 'Unsupported Network' banner.

containers/Connector/Connector.tsx
```diff
@@ -15,6 +15,8 @@ import {
 	synthetix,
 } from '@synthetixio/contracts-interface';
 import { ethers } from 'ethers';
+// @ts-ignore
+import snx from 'synthetix';
```
JChiaramonte7, line 19: I would try to pull everything from [@synthetixio/contracts-interface](https://github.com/Synthetixio/js-monorepo/tree/master/packages/contracts-interface) if possible. In this case we may have to expose new functionality.

containers/Connector/Connector.tsx
```diff
@@ -65,22 +66,29 @@ const useConnector = () => {
...
-			const provider = loadProvider({
+			const _provider = loadProvider({
 				networkId,
 				infuraId: process.env.NEXT_PUBLIC_INFURA_PROJECT_ID,
 				provider: window.ethereum,
 			});
 			const useOvm = getIsOVM(networkId);
 
-			const snxjs = synthetix({ provider, networkId, useOvm });
+			const supportedNetworks = await snx.networks;
```
JChiaramonte7, line 78: If you use `@synthetixio/contracts-interface` you should be able to get a list of networks from: 

https://github.com/Synthetixio/js-monorepo/blob/4e76071b7b63ad112281899c2224b78cc46ff41d/packages/contracts-interface/src/types.ts#L5-L14

Only one that's missing is `local`, but we can add that in the future.

containers/Connector/Connector.tsx
```diff
@@ -65,22 +66,29 @@ const useConnector = () => {
 
 	useEffect(() => {
 		const init = async () => {
-			// TODO: need to verify we support the network
 			const networkId = await getDefaultNetworkId();
 
-			const provider = loadProvider({
+			const _provider = loadProvider({
```
JChiaramonte7, line 71: Internal variable syntax probably more used in solidity than here.

JChiaramonte7 (changes requested):
It looks like the changes are not working for me. I'm getting the console error of network unsupported on Mainnet (and optimistic Mainnet).

https://kwenta-git-fork-snxkm-snxkm-handle-unsupported-networks-kwenta.vercel.app

asherism:
https://www.figma.com/file/R9dygkd4HBA7PdxtE3iLis/External?node-id=0%3A1

Updated Figma link!

containers/Connector/Connector.tsx
snxkm: `provider` is already defined in the outer scope which is why I changed it to `_provider`. WDYT?

snxkm:
@asherism looking great! Thanks.

containers/Connector/Connector.tsx
JChiaramonte7: hmm interesting it was working before, is `provider` coming from `synthetix` ?

Issue #36, opened 2021-09-09
Synthetix/Kwenta build IDs on UI


JChiaramonte7:
## Feature Proposal
We should add build IDs to the bottom of the Kwenta page to help us as the community narrow down help-desk issues. This can be either tied to a Kwenta commit hash or [Synthetix version](https://github.com/Synthetixio/synthetix/releases). Or both.

### Rationale
Someone posts an issue to the help-desk and we inquire which version of Kwenta they're on. 

Having a build ID that links us to either a commit and/or Synthetix can help us eliminate potential causes. Typically issues arise because someone has an old build cached on their machine and the version of Synthetix is outdated. (Context: sometimes Synthetix contract addresses change with each deployment.)

### Implementation
- Some research has to be done into how to get commit hashes and/or Synthetix versions.
- Note: [@synthetixio/contracts-interfaces](https://www.npmjs.com/package/@synthetixio/contracts-interface) version is a good proxy to the Synthetix version and might have the granularity we want.

### Design Spec

Link to Figma (Hard to see, but small and at the bottom of the page):
https://www.figma.com/file/sghReDviy1lcr6kkNqlc0e/Kwenta-User-Flows?node-id=5267%3A228978

Screenshot from Figma:

![image](https://user-images.githubusercontent.com/4015977/132720453-3e3b54d8-da0c-456b-975c-7a9700064cee.png)

ZanePseudo:
Sorry for the two PRs, first one couldn't get my verified commit working properly, but sorted out for the second one that's open

PR #41, opened 2021-09-12
branch for adding git hash to pages

ZanePseudo:
## Description
Made the git hash a global variable in next.config.js, added a GitIDFooter section that could be pulled into any page in Footer.tsx, and threaded that through the dashboard, exchange, and shorting pages at the bottom of each page.

## Related issue
https://github.com/Kwenta/kwenta/issues/36

## Motivation and Context
https://github.com/Kwenta/kwenta/issues/36

## How Has This Been Tested?
Visual testing in Chrome

Really faint at the bottom per figma spec

image

### DIFF
--- a/sections/shared/Layout/AppLayout/Footer/Footer.tsx
+++ b/sections/shared/Layout/AppLayout/Footer/Footer.tsx
@@ -1,0 +1,32 @@
+import { FC } from 'react';
+import styled from 'styled-components';
+
+const Footer: FC = () => {
+    const gitID = process.env.GIT_HASH_ID!.toString();
+
+    return (
+        <GitIDFooter>
+            {gitID}
+        </GitIDFooter>
+    );
+};
+
+const GitIDFooter = styled.div`
+	font-family: Akkurat LL TT;
+    font-style: normal;
+    font-weight: normal;
+    font-size: 10px;
+    line-height: 140%;
+    
+    display: flex;
+    justify-items: center;
+    position: absolute;
+    bottom: 0px;
+    display: grid;
+    width: 100%;
+    margin-bottom: 5px;
+        
+    color: #2B3035;
+`;
+
+export default Footer;
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,10 @@
 // next.config.js
+const gitRevision = require('child_process').execSync('git rev-parse --short HEAD').toString().trim();
 
 module.exports = {
+	env: {
+		GIT_HASH_ID: gitRevision,
+	},
 	webpack: (config, options) => {
 
 		config.resolve.mainFields = ['module', 'browser', 'main'];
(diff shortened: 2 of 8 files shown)


sections/shared/Layout/AppLayout/Footer/Footer.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC } from 'react';
+import styled from 'styled-components';
+
+const Footer: FC = () => {
```
JChiaramonte7, line 4: I think it might be better to call this `GitID` and respectively `GitID.tsx`. It's a bit confusing since there is already another Footer component.

sections/shared/Layout/AppLayout/Footer/Footer.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    return (
+        <GitIDFooter>
+            {gitID}
+        </GitIDFooter>
+    );
+};
+
+const GitIDFooter = styled.div`
+	font-family: Akkurat LL TT;
```
JChiaramonte7, line 15: Pull this from our [fonts theme file](https://github.com/Kwenta/kwenta/blob/dev/styles/theme/fonts.ts) using `${props => props.theme.fonts.mono}`

You won't need the following `font-style` / `font-weight` as the resulting font should override.

Also I know mono font is a little different from the Figma designs, but I feel like it should be used for things like git hashes. cc @asherism

sections/shared/Layout/AppLayout/Footer/Footer.tsx
```diff
@@ -0,0 +1,32 @@
...
+    font-style: normal;
+    font-weight: normal;
+    font-size: 10px;
+    line-height: 140%;
```
JChiaramonte7, line 19: `line-height: 140` is already inherited from the body

sections/shared/Layout/AppLayout/Footer/Footer.tsx
```diff
@@ -0,0 +1,32 @@
...
+    
+    display: flex;
+    justify-items: center;
+    position: absolute;
+    bottom: 0px;
+    display: grid;
+    width: 100%;
```
JChiaramonte7, line 26: The build ID is still slightly off-center.

Remove
```
display: flex
display: grid
justify-items: center;
width: 100%;
```
and replace with
```
left: 50%;
transform: translate(-50%, -50%);
```

sections/shared/Layout/AppLayout/Footer/Footer.tsx
```diff
@@ -0,0 +1,32 @@
...
+    margin-bottom: 5px;
```
JChiaramonte7, line 27: Delete `margin-bottom: 5px` and set your `bottom: 5px `

JChiaramonte7 (changes requested):
Tested and it works well. Just need some CSS cleanup and then the PR should be good to go.

JChiaramonte7:
Great work. Thank you @ZanePseudo!

image

JChiaramonte7 (approved):

PR #77, opened 2021-09-20
fix: update due to some deprecated commodities and equities

vbstreetz:
## Description
Minor update to remove the recent deprecated synths.


constants/currency.ts
```diff
@@ -75,7 +66,7 @@ export const MARKET_HOURS_SYNTHS = new Set([
 	...AFTER_HOURS_SYNTHS,
 ]);
 
-export const COMMODITY_SYNTHS = new Set<CurrencyKey>([Synths.sXAU, Synths.sXAG, Synths.sOIL]);
```
justinjmoses, line 78: There should be nothing in the dapp that hardcodes anything like this. If you need it, ask the synthetix protocol and it can be added to the core functionality exported by the `synthetix` module.

JChiaramonte7: Yeah good call. In the future if we add synths that require special functionality (market hours) we should define it somewhere in `synthetix`

package.json
```diff
@@ -25,13 +25,13 @@
...
-		"@synthetixio/contracts-interface": "2.48.7",
-		"@synthetixio/data": "2.48.7",
-		"@synthetixio/optimism-networks": "2.48.4",
-		"@synthetixio/providers": "2.48.4",
-		"@synthetixio/queries": "2.48.8",
-		"@synthetixio/transaction-notifier": "2.48.4",
-		"@synthetixio/wei": "2.48.4",
+		"@synthetixio/contracts-interface": "2.49.0",
+		"@synthetixio/data": "2.49.0",
+		"@synthetixio/optimism-networks": "2.49.0",
```
JChiaramonte7, line 30: Upgrading `@synthetixio/optimism-networks` introduces too many bugs. Is there a reason this needs to be upgraded?

kmeraz (approved):

PR #88, opened 2021-09-21
feat: point 1inch referrerAddress to kwenta

JChiaramonte7:


constants/address.ts
```diff
@@ -0,0 +1 @@
+export const KWENTA_REFERRAL_ADDRESS = '0x08e30BFEE9B73c18F9770288DDd13203A4887460';
```
kmeraz, line 1: maybe this should eventually make its way into the monorepo - thoughts?

kmeraz (approved):
🧠

PR #106, opened 2021-10-06
Hook up open interest and distribution chart data

avclarke:
## Description
Query data for futures open interest and synths distribution and format for statistics charts


JChiaramonte7 (approved):
Looks great! Merge when ready

PR #153, opened 2021-10-21
Upgrade immer

bachstatter:
Immer has a vulnerability npm audit consider critical and failing CI. Going from v8 -> v9 is a breaking change. 
You can see release notes here: https://github.com/immerjs/immer/releases/tag/v9.0.0.
No type errors and I was able to do trades without any problem. Not sure how I should be testing it more. I haven't used immer too much so a bit unclear to me what the impact of that breaking change is.

## Motivation and Context
CI failing

## How Has This Been Tested?
Tested locally

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
 				"framer-motion": "4.1.17",
 				"graphql-request": "3.4.0",
 				"i18next": "19.7.0",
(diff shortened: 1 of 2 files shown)


JChiaramonte7 (approved):
Approved. Thanks for looking into this

PR #209, opened 2021-12-13
Replace charts with Trading View

avclarke:
Removes the old charts and multiple chart options and replaces them with trading view candlestick chart only.


.prettierignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 build
+public
```
JChiaramonte7, line 3: Is there a reason why we're committing the public folder now? PR seems to add a lot of noise files.

avclarke: That's where the trading view lib lives, but I can scope this to just trading view folder

JChiaramonte7: I see, I guess I just assumed TV was a plug and play npm package.

avclarke: I presume they keep it out of npm as it's kind of private, have to go through that process to get access

JChiaramonte7 (approved):

Issue #231, opened 2022-01-03
Debugging Memory Leaks on L1 Shorts


JChiaramonte7:
Seems to come from an issue infinite `estimateGas` calls

Deliverables
---
- [ ] Make sure there are no extraneous `estimateGas` gas calls hogging up system resources and provider bandwidth

PR #235, opened 2022-01-03
prevent infinite gasPrice fetching in L1 Shorting Rewards

kmeraz:
## Description

The Shorting Rewards section on L1 is doing a fetch for each currency, then updating the gasInfo based on the requests.

I'm not sure why this was done this way, since each result of the API call is overwritten by the subsequent result. e.g. the call for sETH overwrites the result from the previous call to sBTC.

Removing these fetches altogether takes care of the memory leak issue.

## Related issue

https://github.com/Kwenta/kwenta/issues/231

## How Has This Been Tested?

If you visit L1 shorts and pop open the `Performance Monitor` tab in Chrome Dev Tools, you'll now see the CPU usage steadily in the single digits (as opposed to 100+, as it's currently behaving in production)
image


JChiaramonte7 (approved):

JChiaramonte7:
Make sure you update dev too

PR #242, opened 2022-01-07
[UI Refresh] - Button component

koredefashokun:
## Description
This PR creates a new Button component for the Kwenta UI, based on the new designs.

It is important to note that this new Button does not have feature-parity with the prior version yet. It includes styles for `primary` (regular), `success`* and `danger` variants, but doesn't have styles for the other variants in the previous version. These will be added in the future. Some notable change include:
- Addition of a `mono` prop, which uses monospaced text in the button, and
- Addition of a `fullWidth` prop, which makes the button extend to the width of the containing view.

The attached screenshots show the button in different sizes, variants and states, as well as the current future market screen with this Button component in use.

(*) - The `success` variant did not exist in the previous version of the button.

## Related issue
N/A

## Motivation and Context
This PR is part of the UI revamp for Kwenta.

## How Has This Been Tested?
N/A

image
image
image
image


JChiaramonte7 (approved):
We should ask @asherism how to handle text buttons, like "Max" when selecting max wallet balance.

PR #245, opened 2022-01-08
[UI Refresh] - Add Input component

koredefashokun:
## Description
This PR updates the existing `Input` component, according to the new design specifications. It also adds a new component `OrderSizingInput`, which is used on the futures screen to set the order size. Attached is a screenshot of the futures market screen with the updated inputs.

## Related issue
N/A

## Motivation and Context
This is part of the Kwenta UI refresh.

## How Has This Been Tested?
N/A

image

### DIFF
--- a/components/Input/Input.tsx
+++ b/components/Input/Input.tsx
@@ -4,17 +4,18 @@ export const inputCSS = css`
 	width: 100%;
 	min-width: 0;
 	font-family: ${(props) => props.theme.fonts.regular};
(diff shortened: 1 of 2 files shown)


JChiaramonte7 (commented):
Noticed the screenshot the width doesn't scale 100%?

koredefashokun:
@JChiaramonte7 It does. That's the available width from the containing view. The screenshot below shows the current previous input component for reference:

image

JChiaramonte7 (approved):

PR #247, opened 2022-01-10
update OVM Shorts' subgraph endpoint

kmeraz:
## Description
We were using my own subgraph that we deployed while the synthetix subgraph synced. This PR switches away from the former to now use the latter.


JChiaramonte7 (approved):
LGTM

PR #257, opened 2022-01-13
[UI Refresh] - Add LeverageSlider component

koredefashokun:
## Description
This PR adds a `LeverageSlider` component, for use on the futures screen. The screenshot below shows the 

It should be noted that this component is only intended for use in inputting leverage on the futures screen. The component uses an SVG background (with a fixed with and height, as well as marks in specific spots) and very exact CSS values that makes it near impossible for use in another context.

## Related issue
N/A

## Motivation and Context
This PR is part of the Kwenta UI refresh.

## How Has This Been Tested?
N/A

image


JChiaramonte7 (approved):

sections/futures/LeverageSlider/LeverageSlider.tsx
```diff
@@ -0,0 +1,143 @@
...
+		margin-top: -11px;
+		background-color: transparent;
+		height: 24px;
+		left: 0;
+		right: 0;
+	}
+
+	.MuiSlider-track {
+		height: 4px;
+		background-color: #7d6b54;
```
JChiaramonte7, line 71: Not sure if we wanted to make these colors reuse-able in the theme file?

koredefashokun: Yeah. The plan is to move all the hardcoded colors out in a separate PR. I didn't want to affect other components dependent on some of the colors.

PR #288, opened 2022-01-25
[UI refresh] - Move colors out of components

koredefashokun:
## Description
This PR moves hardcoded colors out of individual components, to make future theming endeavours much easier to undertake. It is important to note that this PR simply hoists these colors from the component level to a theme file, but all work related to actually switching themes is not included in this PR, as it is considered beyond its scope (the details of these other themes have also not been finalized).

It is expected that once the other themes are completed, there will be some changes, to make sure that colors, gradients and other motifs that overlap between them can be reused accordingly. For example the `props.theme.colors` object might be akin to this:

```js
colors: {
  common: {
    // Colors that are reused across all themes.
  },
  defaultTheme: {
    border: 'white'
  },
  lightTheme: {
    border: 'black'
  },
  ...
}
```

Since the themes will have similar keys, we would be able to access the colors using syntax similar to this `box-shadow: ${props => props.theme.colors[themeName].button.shadow}`, where `themeName` is a variable equivalent to the value of the name of the currently selected theme.

## Related issue
N/A

## Motivation and Context
N/A

## How Has This Been Tested?
N/A

N/A


JChiaramonte7 (approved):
Approval for after you update the theme name

Issue #303, opened 2022-01-31
Disable Shorting on L1


platschi:
### Subject of the issue
Shorting on Ethereum mainnet (L1) is [deprecated (SCCP-146)](https://sips.synthetix.io/sccp/sccp-146/) in favor of L2 Shorts. Kwenta should reduce user confusion for those wanting to open news shorts on L1.

### Your environment
* Ethereum Mainnet
* Firefox

### Actual behaviour
Shorts on L1 still active

### Expected behaviour
Shorts on L1 should not be available given attempting to open a new short simply fails; user should be directed to L2 Shorting.

### Steps to reproduce
- Set network to Ethereum L1
- Visit https://kwenta.io/shorting

### Deliverables
- Show a clear deprecation message instead of the `ShortingCard` modal **if on L1.** 
- Instead inform the user that shorts on L1 are deprecated in favor of L2 shorts, include a button to "Switch to L2"

**Important**: Make sure to keep the "Your Position" UI for those with existing shorts on L1 to be able to manage their positions!

### Rewards
50 SNX

platschi:
Assigned to @kethcode

PR #314, opened 2022-02-01
Added graceful redirect message from L1 to L2 for shorting

kethcode:
## Description

Added MessageContainer and Message to shorting/common and ShortingCard
Display message and "Switch to L2' button when L1 ShortingCard visible
Added new redirect message to translations/en.json
 - "Shorting on L1 is no longer available; please switch to L2, or manage your L1 positions below."

## Related issue
https://github.com/Kwenta/kwenta/issues/303

## Motivation and Context
Shorting on Ethereum mainnet (L1) is deprecated (SCCP-146)
This improves the user experience, compared to attempted L1 shorting transactions failing.

## How Has This Been Tested?
Tested on Ubuntu 21.04 and Firefox and Chrome.
Switching between L1 and L2 Shorting tabs from App Header Layout

image

### DIFF
--- a/sections/shorting/ShortingCard/ShortingCard.tsx
+++ b/sections/shorting/ShortingCard/ShortingCard.tsx
@@ -15,8 +15,8 @@ import { useTranslation, Trans } from 'react-i18next';
 import { useRecoilValue } from 'recoil';
 import { isL2State } from 'store/wallet';
 
-import { MessageContainer, Message } from '../common';
-import NetworksSwitcher from 'sections/shared/Layout/AppLayout/Header/NetworksSwitcher';
+import { MessageContainer, Message, MessageButton } from '../common';
+import useNetworkSwitcher from 'hooks/useNetworkSwitcher';
 
 const ShortingCard: FC = () => {
 	const { quoteCurrencyCard, baseCurrencyCard, footerCard } = useShort({
@@ -26,6 +26,7 @@ const ShortingCard: FC = () => {
 
 	const { t } = useTranslation();
 	const isL2 = useRecoilValue(isL2State);
+	const { switchToL1, switchToL2 } = useNetworkSwitcher();
 
 	return (
 		<Container>
@@ -45,7 +46,9 @@ const ShortingCard: FC = () => {
 					<Message>
 						<Trans t={t} i18nKey="shorting.l1-deprecated" />
 					</Message>
-					<NetworksSwitcher />
+					<MessageButton size="lg" variant="primary" isRounded={true} onClick={switchToL2}>
+						{t('header.networks-switcher.l2')}
+					</MessageButton>
 				</MessageContainer>
 			)}
 		</Container>
--- a/sections/shorting/ShortingCard/ShortingCard.tsx
+++ b/sections/shorting/ShortingCard/ShortingCard.tsx
@@ -11,24 +11,43 @@ import useShort from '../hooks/useShort';
 
 import { CurrencyCardsSelector, ExchangeCardsWithSelector } from 'styles/common';
 
+import { useTranslation, Trans } from 'react-i18next';
+import { useRecoilValue } from 'recoil';
+import { isL2State } from 'store/wallet';
+
+import { MessageContainer, Message } from '../common';
+import NetworksSwitcher from 'sections/shared/Layout/AppLayout/Header/NetworksSwitcher';
+
 const ShortingCard: FC = () => {
 	const { quoteCurrencyCard, baseCurrencyCard, footerCard } = useShort({
 		defaultBaseCurrencyKey: Synths.sETH,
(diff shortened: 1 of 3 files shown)


sections/shorting/common.tsx
```diff
@@ -7,3 +12,48 @@ export const Title = styled.div`
...
+
+	/*
+	width: 100%;
+	border-radius: 1000px;
+	grid-template-columns: 1fr auto;
+	background-color: ${(props) => props.theme.colors.elderberry};
+	padding: 16px 32px;
+	max-width: 750px;
+	margin: 0 auto;
+	*/
```
platschi, line 46: CSS in comments obsolete?

kethcode:
Sounds good, I'll make those changes, and look for anything else.

kethcode:
So, a couple thoughts: 

Removing the stale CSS is fine, but that code is from sections/exchange/common.  I considered including that directly instead, but copy-pasted to keep local common components local, in case they needed customization.  One the one hand, I would prefer to use globally common components for style.  On the other, makes regression testing more complex.

I'll remove the dead CSS, but leave the duplicate code in sections/exchange/common for now.

The NetworksSwitcher component discussion is much the same, with a subtle difference in my opinion.  Swapping for the Hook, and keeping local changes local scoped, is a good point.  However, I think the Switch the L2 functionality should remain centralized, so changes in one place are reflected everywhere.  There's also a lot of formatting inherited from the component that I would have to duplicate to make it look right in the MessageContainer.

I'm inclined to leave the NetworksSwitcher component as is.

kethcode:
removed dead css.  left NetworksSwitcher alone.

translations/en.json
```diff
@@ -510,7 +510,9 @@
 				"no-results": "No shorts found",
 				"accrued-interest": "Accrued Interest"
 			}
-		}
+		},
+		"l1-deprecated": "Shorting on L1 is no longer available; please switch to L2, or manage your L1 positions below."
```
JChiaramonte7, line 514: Can we change "is no longer available" to "is deprecated"?

sections/shorting/ShortingCard/ShortingCard.tsx
```diff
@@ -11,24 +11,43 @@ import useShort from '../hooks/useShort';
...
+						</StyledCurrencyCardsSelector>
+					</ExchangeCardsWithSelector>
+					<ExchangeFooter>{footerCard}</ExchangeFooter>
+				</ConvertContainer>
+			) : (
+				<MessageContainer>
+					<Message>
+						<Trans t={t} i18nKey="shorting.l1-deprecated" />
+					</Message>
+					<NetworksSwitcher />
```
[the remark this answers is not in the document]
JChiaramonte7: Thinking this approach might be better in the event that there's regression in the NetworkSwitcher component.

Also might make more sense to use our default button style (the gold one) since that's typically used for messages in this format.

JChiaramonte7 (changes requested):

translations/en.json
kethcode: done

kethcode:
NetworksSwitcher component doesn't have an alternate style for gold.  Do you want me to build out an alternate?

JChiaramonte7:
> To be clear, do you want the component or the hook for network switcher?

The hook. Sorry was referencing @platschi 's previous comment.

kethcode:
All requested changes made.  Committed again.

platschi (approved):

sections/shorting/ShortingCard/ShortingCard.tsx
```diff
@@ -11,24 +11,46 @@ import useShort from '../hooks/useShort';
...
+
 const ShortingCard: FC = () => {
 	const { quoteCurrencyCard, baseCurrencyCard, footerCard } = useShort({
 		defaultBaseCurrencyKey: Synths.sETH,
 		defaultQuoteCurrencyKey: Synths.sUSD,
 	});
 
+	const { t } = useTranslation();
+	const isL2 = useRecoilValue(isL2State);
+	const { switchToL1, switchToL2 } = useNetworkSwitcher();
```
JChiaramonte7, line 29: Is switchToL1 used?

kethcode: it is not.  I saw the warning, but don't know how to handle it properly.  We explicitly return two values, so just omitting one of them in the deconstruction throws an error on execution.  I could alias it to _ or {} if you want.  Or create an alternate switcher function that only returns L2, and not L1.  Or whatever else you want.  Your call.

kethcode: Edit, of course, now that I try omitting it again, it works fine.  Sigh.  

Recommitting now.

JChiaramonte7 (approved):

Issue #341, opened 2022-02-08
Exchange Fee Cost not calculated correctly


platschi:
### Subject of the issue
Currently, exchange fees are wrongly calculated based on the market rate. Fee costs should be calculated based on the displayed `USD value` (at $1).

In the example below (L1 Kovan), exchanging sUSD should calculate the 0.5% fee costs based of the displayed `USD Value` of $1,000.00, which should be $5.00 and not $4.98.

![Screenshot 2022-02-08 at 16-35-28 Kwenta](https://user-images.githubusercontent.com/548702/153062593-2e02ff18-b681-433b-8411-ad230d264372.png)

### Your environment
* L1 and L2

### Deliverables

- [ ] Calculate exchange fee based on shown USD Value, not market rate.

### Steps to reproduce
1. Visit [Kwenta Exchange](https://kwenta.io/exchange/sUSD-sETH) on L1 or L2
2. Set sUSD to any other synthetic asset
3. Notice how the fee cost is not calculated based on the USD value displayed (input >350 sUSD to notice the difference due to rounding to two decimals)

### Rewards
25 SNX

PR #342, opened 2022-02-09
issue #341 - fix Exchange Fee Cost for sUSD

brossetti1:
## Description
the core issue comes from the `feeCost` calculation in the `useExchange` hook. The current implementation is calculating fees based off of `baseCurrencyAmount`, which already has the fees subtracted. Instead, we should use the `quoteCurrencyAmount` too calculate the fees. solution is as follows:

1. added a calculation called `feeAmountInQuoteCurrency` that gets memoized on its own
2. exchange `feeCost` calculation from baseCurrency and baseRate to quoteCurrency and quoteRate.

```
const feeCost = useMemo(() => {
  if (feeAmountInBaseCurrency != null) {
    return feeAmountInBaseCurrency.mul(basePriceRate);
  }
  return null;
}, [feeAmountInBaseCurrency, basePriceRate]);
```

**vs**

```
const feeCost = useMemo(() => {
	if (feeAmountInQuoteCurrency != null) {
		return feeAmountInQuoteCurrency.mul(quotePriceRate);
	}
	return null;
}, [feeAmountInQuoteCurrency, quotePriceRate]);
```

**This fix applies the correct calculation across all synths, inlcuding sUSD**

## Related issue
https://github.com/Kwenta/kwenta/issues/341

## Motivation and Context
- when trading against the synth pair, the fee cost should be applied to the `quoteCurrencyAmount` and `quoteRate` 
- when trading against the synth pair, the formatting of different values across the exchange only displays 1 decimal places when the decimals result in a `.00` value.

## How Has This Been Tested?

google chrome on a mac OS laptop
node v14.17.1
npm v8.4.0

1. I tested this by loading the app as per the README:

```bash
> npm i
...
> npm run dev
...
```

From the localhost setup I connected to `optimism mainnet` and tested against my own wallet balances (tried to grab some from the optimism kovan faucet but was having trouble there). I was able to confirm the fix by comparing my localhost connected to `optimism mainnet` with https://kwenta.io/exchange/sETH-sUSD and differentiating the formatting and fee cost amounts.

2. I have noticed a formatting issue that is applied to several spots when sUSD (and sETH) are selected, see screen shots for references there.

There are some e2e tests for the exchange, I have not tried those as yet.

##### after changes:
Screen Shot 2022-02-08 at 3 51 59 PM

##### comparison of formatting against sUSD

Screen Shot 2022-02-08 at 2 53 38 PM
Screen Shot 2022-02-08 at 4 01 11 PM

### DIFF
--- a/sections/exchange/hooks/useExchange.tsx
+++ b/sections/exchange/hooks/useExchange.tsx
@@ -554,12 +554,24 @@ const useExchange = ({
 		return null;
 	}, [baseCurrencyAmount, exchangeFeeRate]);
 
+	const feeAmountInQuoteCurrency = useMemo(() => {
+		if (exchangeFeeRate != null && quoteCurrencyAmount) {
+			return wei(quoteCurrencyAmount).mul(exchangeFeeRate);
+		}
+
+		return null;
+	}, [quoteCurrencyAmount, exchangeFeeRate]);
+
 	const feeCost = useMemo(() => {
+		if (quoteCurrencyKey?.toString() === 'sUSD') {
+			return feeAmountInQuoteCurrency;
+		}
+
 		if (feeAmountInBaseCurrency != null) {
 			return feeAmountInBaseCurrency.mul(basePriceRate);
 		}
 		return null;
-	}, [feeAmountInBaseCurrency, basePriceRate]);
+	}, [feeAmountInBaseCurrency, basePriceRate, feeAmountInQuoteCurrency, quoteCurrencyKey]);
 
 	useEffect(() => {
 		setCurrencyPair({
--- a/sections/exchange/hooks/useExchange.tsx
+++ b/sections/exchange/hooks/useExchange.tsx
@@ -547,13 +547,6 @@ const useExchange = ({
 		[gasPrice, gasInfo?.limit, ethPriceRate, gasInfo?.l1Fee]
 	);
 
-	const feeAmountInBaseCurrency = useMemo(() => {
-		if (exchangeFeeRate != null && baseCurrencyAmount) {
-			return wei(baseCurrencyAmount).mul(exchangeFeeRate);
-		}
-		return null;
-	}, [baseCurrencyAmount, exchangeFeeRate]);
-
 	const feeAmountInQuoteCurrency = useMemo(() => {
 		if (exchangeFeeRate != null && quoteCurrencyAmount) {
 			return wei(quoteCurrencyAmount).mul(exchangeFeeRate);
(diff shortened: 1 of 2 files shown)


brossetti1:
so i ive tracked down the "issue" with only 1 trailing 0 - this does not have to do with sUSD being selected or any other synth - its the default behavior of `utils.comify`  - if the decimals are `.00` it will trim it to `.0` and join it with the comified whole number

https://github.com/ethers-io/ethers.js/blob/master/packages/units/src.ts/index.ts#L23-L62

specifically right here:
https://github.com/ethers-io/ethers.js/blob/master/packages/units/src.ts/index.ts#L45-L47

you can also see this in the ethers docs here https://docs.ethers.io/v3/api-utils.html?highlight=pad#id1

I added a wrap function in `numbers.ts` around `commify` to reverse the commify single zero behavior - if thats not the desired behavior I can revert the [931d5b3](https://github.com/Kwenta/kwenta/pull/342/commits/931d5b3253a1334738218edb1723af9bfe4eb597)

brossetti1:
on another note - I applied the fix against sUSD `useShort` and the `ManageShortAction` on the final [464695e](https://github.com/Kwenta/kwenta/pull/342/commits/464695e9cda39e19dcbc37c247941f305a4d8c74) commit

looks like the interest rate is off potentially?


Screen Shot 2022-02-08 at 9 55 45 PM

I am unable to test this against the `ManageShortAction` as im using `mainnet optimism` to test

ill wait for a reply to finalize these changes.

JChiaramonte7:
So while reviewing your PR @platschi and I realized that the real problem here was that the "fee cost" was calculated from the base asset (output currency), but the base asset amount we receive from the smart contracts already have the fee taken out before returning**. Your solution actually does somewhat fix the issue for cases when sUSD is the quote asset, but it still exists if any other synth is a quote asset.

What we should do here is actually calculate the "fee cost" from the **sUSD value** *of* quote asset. That would give us the most accurate cost.

Also thanks for the two digit formatting fix!

** Helpful documentation under *Events Emitted* here: https://docs.synthetix.io/contracts/transactions/#trading-exchanging-synths

brossetti1:
@JChiaramonte7 @platschi ive got the exchange fees calculating correctly based on the `QuoteCurrencyFeeAmount`, can you confirm I should do the same for `useShort` and `ManageShortAction`?

JChiaramonte7 (approved):
Beautifully concise solution. This is how all future PRs should be. Thank you @brossetti1

PR #339, opened 2022-02-08
Fix the missing currency, ordering of trx and hide reclaim fee on l2

LeifuChen:
## Description
The following improvements have been made to this pull request.
1.  In the [transactions table](https://kwenta.io/dashboard/transactions), the synth symbol has been added to the ```To``` column.

2.  In the [transactions table](https://kwenta.io/dashboard/transactions), transactions are listed chronologically (latest tx on top) regardless of the swapped synth.

3. In the [transactions table](https://kwenta.io/dashboard/transactions),  the "Rebate/Reclamation" column is hidden if network is L2.

## Related issue
#336 

## Motivation and Context
Let's make the transaction history more user-friendly.

## How Has This Been Tested?
I made three transactions on Optimism Kovan and another three transactions on Kovan Test Network.
The transaction history shows everything as we expected in the above [Description](##description).

## Screenshots (if appropriate):
Transactions on Kovan Test Network
kovan_test
Transactions on Optimism Kovan
op_kovan

PR #348, opened 2022-02-15
Revert "Fix the missing currency, ordering of trx and hide reclaim fee on l2 (#339)"

kmeraz:
This reverts commit 93dc18791288826d0accb637d7937d5d9328d723.

### DIFF
--- a/components/Table/Table.tsx
+++ b/components/Table/Table.tsx
@@ -33,7 +33,6 @@ type TableProps = {
 	noResultsMessage?: React.ReactNode;
 	showPagination?: boolean;
 	pageSize?: number | null;
-	hiddenColumns?: string[];
 };
 
 export const Table: FC<TableProps> = ({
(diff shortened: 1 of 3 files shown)


JChiaramonte7 (approved):

JChiaramonte7:
@LeifuChen We had a bug from your PR that caused the shorts page to crash when a wallet was connected and we will have to temporarily revert the commit

image

LeifuChen:
Let me investigate this issue.

components/Table/Table.tsx
```diff
@@ -48,7 +47,6 @@ export const Table: FC<TableProps> = ({
 	className,
 	showPagination = false,
 	pageSize = null,
-	hiddenColumns = undefined,
```
LeifuChen, line 51: @JChiaramonte7 I checked the shorting page. There is a shorting history table using this component and undefined is not an acceptable data type. We have to pass an empty array here, like this ```hiddenColumns = []```. How do you want to fix this issue?

Issue #343, opened 2022-02-11
Kwenta/GnosisSafe wallet connection issue on Mobile


platschi:
### Subject of the issue
Using the Kwenta dApp within Gnosis Safe on **mobile viewport** doesn't automatically connect to the Gnosis Safe. It does work totally fine on desktop, though.

### Your environment
* Optimism
* Chrome

Tested on desktop with Chrome Inspector on "iPad Air" mobile viewport.

### Expected behaviour
Using Gnosis Safe on Mobile, the Gnosis SDK should automatically connect Kwenta to the connected Safe.

### Steps to reproduce
1. Visit [GnosisSafe](https://gnosis-safe.io/app), make sure it is set to **Optimism**.
2. Add an existing safe (can be read-only). Contact the community PM for a test safe wallet.
3. Visit Apps > Add Custom App > Add [Kwenta](https://kwenta.io). Reload the Apps page and it should appear on the top.
4. Visit Kwenta under Apps. The Kwenta UI should automatically connect to the Gnosis Safe wallet.

This works fine on Desktop. Switching to mobile viewport and trying to visit Apps > Kwenta again, Kwenta does not connect to Gnosis Safe.

### Deliverables
- [ ] Find out what causes the issue
- [ ] Fix the bug and make it work on mobile viewport

### Rewards
25 SNX

### Examples
1. Working Gnosis Safe with Kwenta on Desktop
2. Same set-up with mobile viewport (not connecting)

<img width="1442" alt="Screen Shot 2022-02-11 at 12 35 29" src="https://user-images.githubusercontent.com/548702/153620992-37159133-c897-46f9-87ad-6a36555de28d.png">

<img width="533" alt="Screen Shot 2022-02-11 at 12 34 54" src="https://user-images.githubusercontent.com/548702/153621103-934b1d68-1ee4-49c4-9396-c5a2412db92b.png">

platschi:
Assigned to @burtrico

PR #363, opened 2022-02-19
Fixed gnosisSafe wallet connect option for mobile

burtrico:
## Description
The gnosis wallet connect option was not working for mobile, it would automatically disconnect a user. I noticed that the wallet connect modal actually didn't even show gnosis when you switched to mobile. I went to the wallet connector config.ts and found that 'gnosis' was missing the `rpcUrl: infuraRpc` option. After adding this, Gnosis then appeared in the modal when on mobile. I was able to connect to the wallet and/or stay connected after that. 

## Related issue
(https://github.com/Kwenta/kwenta/issues/343)

## Motivation and Context
GnosisSafe is a popular, more secure, wallet that any major application should support. 

## How Has This Been Tested?
On gnosisSafe.io, I used ngrok.io to forward the provided URL to my local host 3000. I used npm run build and run dev, then copied the ngrok URL into gnosisSafe's custom app (example = https://4286-104-54-248-129.ngrok.io ). I then switched the chrome dev tool inspector to mobile. 
Ngrok.io loaded slow and often failed to load in gnosisSafe, but I was able to notice that added the `rpcUrl: infuraRpc` to the gnosis wallet config, solved the problem.
I did not notice any changes to others areas of the code since this was a minor change.

BEFORE the fix:
Screen Shot 2022-02-19 at 3 49 24 PM

AFTER the fix:
Screen Shot 2022-02-19 at 3 48 33 PM

### DIFF
--- a/containers/Connector/config.ts
+++ b/containers/Connector/config.ts
@@ -59,7 +59,7 @@ export const initOnboard = (network: Network, subscriptions: Subscriptions) => {
 					walletName: 'portis',
 					apiKey: process.env.NEXT_PUBLIC_PORTIS_APP_ID,
 				},
-				{ walletName: 'gnosis' },
+				{ walletName: 'gnosis', rpcUrl: infuraRpc },
 				{ walletName: 'trust', rpcUrl: infuraRpc },
 				{ walletName: 'walletLink', rpcUrl: infuraRpc, preferred: true },
 				{ walletName: 'torus' },
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
 	"requires": true,
 	"packages": {
 		"": {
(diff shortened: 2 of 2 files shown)


platschi (commented):
Auditor should check `package-lock.json` change, `name` field has been added. Wonder why it wasn't there before, assuming that it's usually required.

burtrico:
> I think there is a slight misunderstanding with this one. It should **auto-connect** to the safe even when in the mobile viewport like it does in desktop mode. Safe app user shouldn't have to manually connect their wallet.

Ok, I'll take a look

platschi (approved):
Works on mobile (now shows Gnosis Safe as an option). Auto-connect does not work due to constrains by blocknative.

PR #366, opened 2022-02-21
Feature: ens support

0xwontonsoup:
## Description
Fixes #356 
* Adds ENS support to `Connector.tsx` component

## Related issue
Related to #356, i just finished where this user left off.

## Motivation and Context
Needed ENS support on Optimism

## How Has This Been Tested?
* I used my own infura URL and my own private ENS name to test that the `getENSNameAndAvatarUrl()` method paired with the new `useEffect()` hook, which is inside of the `useConnector()` function, worked to fetch and display an ENS name on the UserMenu's WalletButton component
* By default, the ENS name text was capitalized. To fix this, I added `text-transform` to the `WalletButton` styled-component and added a prop to it with a ternary.
    * The ternary is used to only change the text to lowerccase when the `WalletButton` has its className attribute set to `lowercase`, otherwise, nothing is affected.

none


0xwontonsoup:
# Note
The WalletButton's `truncatedAddress` will update to the ENS name (if available) after ~2 seconds.

Let me know if this is an issue 🙂

containers/Connector/Connector.tsx
```diff
@@ -160,13 +163,31 @@ const useConnector = () => {
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [isAppReady]);
 
+	useEffect(() => {
```
JChiaramonte7, line 166: Not sure if this belongs in the `connector.tsx`. Should be local to wherever we render wallet addresses and logic should probably be in a reusable hook.

containers/Connector/Connector.tsx
```diff
@@ -160,13 +163,31 @@ const useConnector = () => {
...
+		if (signer) {
+			let provider,
+				account: any = signer.getAddress();
+			const infuraUrl = 'https://mainnet.infura.io/v3/';
+
+			provider = new ethers.providers.JsonRpcProvider(
```
JChiaramonte7, line 172: We shouldn't redefine the provider here. If you move this logic outside of the connector you should be able to retrieve the provider globally from the connector.

containers/Connector/Connector.tsx
```diff
@@ -160,13 +163,31 @@ const useConnector = () => {
...
+				}
+			});
+		}
+	}, [signer]);
+
 	// load previously saved wallet
 	useEffect(() => {
 		// disables caching if in IFrame allow parent to set wallet address
-		if (onboard && selectedWallet && !walletAddress && !isIFrame()) {
+		if (onboard && selectedWallet && !walletAddress && !ensName && !ensAvatar && !isIFrame()) {
```
JChiaramonte7, line 187: Not sure why this logic was added/needed?

JChiaramonte7 (changes requested):
Can you attach a screenshot of this working? It's not working for me in the deployment preview.

containers/Connector/Connector.tsx
0xwontonsoup: I didn't add that logic in, so im not sure either
Felt i should leave it alone

containers/Connector/Connector.tsx
0xwontonsoup: since the default chainid of the network that is set in the `Connector.tsx` is for Optimism, then i will still need to instantiate a new Ethereum mainnet network

i'll use the same logic that is in `Connector.tsx` to do that in `UserMenu.tsx`:
```
const provider = loadProvider({
				networkId,
				infuraId: process.env.NEXT_PUBLIC_INFURA_PROJECT_ID,
				provider: window.ethereum,
			});
```

containers/Connector/Connector.tsx
0xwontonsoup: i understand the intent to turn it into a reusable hook, but if i do i expect issues since the logic does NOT return anything

it merely sets the state variable

sections/shared/Layout/AppLayout/Header/UserMenu/UserMenu.tsx
```diff
@@ -56,6 +59,25 @@ const UserMenu: FC<UserMenuProps> = ({ isTextButton }) => {
...
+	);
+
+	useEffect(() => {
+		if (signer) {
+			let account: any = signer.getAddress();
+			getENSNameAndAvatarUrl(account, ensProvider).then((ensObj) => {
+				if (ensObj !== null) setEns(ensObj?.name);
+			});
+		}
+	}, [signer]);
```
0xwontonsoup, line 79: if you still don't want to add a useEffect hook inside of `UserMenu.tsx`, i understand

just lmk 😃

JChiaramonte7:
Can you drop in a screenshot of this working?

0xwontonsoup:
> Can you drop in a screenshot of this working?

sure, give me a few min

0xwontonsoup:
> Can you drop in a screenshot of this working?

Here's a screenshot:

Screen Shot 2022-03-01 at 7 42 04 PM

JChiaramonte7:
Thanks. It's still not working for me. Can you let me know if you can see your ENS name here?: https://kwenta-rm2xu25op-kwenta.vercel.app/

This is the deployment URL for your latest commit.

sections/shared/Layout/AppLayout/Header/UserMenu/UserMenu.tsx
JChiaramonte7: I think this is okay for now. I don't think we show a user's address anywhere else.

0xwontonsoup:
> Thanks. It's still not working for me. Can you let me know if you can see your ENS name here?: https://kwenta-rm2xu25op-kwenta.vercel.app/
> 
> This is the deployment URL for your latest commit.

Hey @JChiaramonte7 @platschi, i think the 403 error in the console is the main issue
When i click on the error and view the preview on the network tab, i see that the error message from the infura url (`https://mainnet.infura.io/v3/62f6d669dcdb479d87f5719c78f5f776`) reads:

Screen Shot 2022-03-02 at 10 26 28 AM

sections/shared/Layout/AppLayout/Header/UserMenu/UserMenu.tsx
```diff
@@ -56,6 +59,27 @@ const UserMenu: FC<UserMenuProps> = ({ isTextButton }) => {
...
 
+	/**
+	 * @dev We require this new provider since we need one connected to Ethereum
+	 * mainnet and NOT to the Optimism network.
+	 */
+	const infuraUrl = 'https://mainnet.infura.io/v3/';
+
+	let ensProvider = new ethers.providers.JsonRpcProvider(
+		infuraUrl + process.env.NEXT_PUBLIC_INFURA_PROJECT_ID
+	);
```
JChiaramonte7, line 70: Can you try
```
let ensProvider = new ethers.providers.InfuraProvider(
	null, process.env.NEXT_PUBLIC_INFURA_PROJECT_ID
);
```
and if that doesn't work
```
let ensProvider = new ethers.providers.InfuraProvider()
```
https://docs.ethers.io/v5/api/providers/api-providers/#InfuraProvider

0xwontonsoup: ill try that rn 🙂

0xwontonsoup:
@JChiaramonte7 still seeing the same error message with the new `ensProvider`:
```js
const apiKey = process.env.NEXT_PUBLIC_INFURA_PROJECT_ID;

let ensProvider = new ethers.providers.InfuraProvider(1, apiKey);
```

Screen Shot 2022-03-02 at 1 20 43 PM

JChiaramonte7:
> @JChiaramonte7 still seeing the same error message with the new `ensProvider`:
> 
> ```js
> const apiKey = process.env.NEXT_PUBLIC_INFURA_PROJECT_ID;
> 
> let ensProvider = new ethers.providers.InfuraProvider(1, apiKey);
> ```
> 
> Screen Shot 2022-03-02 at 1 20 43 PM

Okay let's try to use the default InfuraProvider for now then (no parameters), it seems like we need to fix our Infura provider infrastructure first

sections/shared/Layout/AppLayout/Header/UserMenu/UserMenu.tsx
0xwontonsoup: > > @JChiaramonte7 still seeing the same error message with the new `ensProvider`:
> > ```js
> > const apiKey = process.env.NEXT_PUBLIC_INFURA_PROJECT_ID;
> > 
> > let ensProvider = new ethers.providers.InfuraProvider(1, apiKey);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Screen Shot 2022-03-02 at 1 20 43 PM
> 
> Okay let's try to use the default InfuraProvider for now then (no parameters), it seems like we need to fix our Infura provider infrastructure first

updated the `ensProvider` to accept no parameters

I'll wait for an update from you then on the infura provider 🙂

sections/shared/Layout/AppLayout/Header/UserMenu/UserMenu.tsx
```diff
@@ -56,6 +59,23 @@ const UserMenu: FC<UserMenuProps> = ({ isTextButton }) => {
 		[redeemableDeprecatedSynths?.totalUSDBalance]
 	);
 
+	/**
+	 * @dev We require this new provider since we need one connected to Ethereum
+	 * mainnet and NOT to the Optimism network.
+	 */
+	let ensProvider = new ethers.providers.InfuraProvider();
```
JChiaramonte7, line 66: Looks like this works. Can we move this to the `connector.tsx` and expose `staticMainnetProvider` and then use that here like so: `const { connectWallet, signer, staticMainnetProvider } = Connector.useContainer();`?

sections/shared/Layout/AppLayout/Header/UserMenu/getUserENSNameAndAvatarUrl.ts
```diff
@@ -0,0 +1,10 @@
+export default async function getENSNameAndAvatarUrl(_account: string, _provider: any) {
```
JChiaramonte7, line 1: Can we rename this to `getENSName` because it doesn't look like we're doing anything with the avatar

sections/shared/Layout/AppLayout/Header/UserMenu/getUserENSNameAndAvatarUrl.ts
```diff
@@ -0,0 +1,10 @@
...
+	const name = await _provider.lookupAddress(_account);
+
+	if (!name) return null;
+
+	const resolver = await _provider.getResolver(name);
+	const avatar = resolver ? await resolver.getText('avatar') : null;
```
JChiaramonte7, line 7: Can we remove all logic and references to the avatar since we're not using it

sections/shared/Layout/AppLayout/Header/UserMenu/UserMenu.tsx
```diff
@@ -28,6 +29,7 @@ import SettingsModal from 'sections/shared/modals/SettingsModal';
 
 import ConnectionDot from '../ConnectionDot';
 import NetworksSwitcher from '../NetworksSwitcher';
+import getENSNameAndAvatarUrl from './getUserENSNameAndAvatarUrl';
```
JChiaramonte7, line 32: rename this file as well

JChiaramonte7 (changes requested):
Everything looks like it's working now. We'll rely on the default Infura provider for now until we fix our infra

JChiaramonte7 (approved):

PR #378, opened 2022-02-27
Link futures data to v2 UI

koredefashokun:
## Description
This PR begins the process of connecting the components of the new UI to the futures data.

## Related issue
N/A

## Motivation and Context
This PR is part of the Kwenta UI refresh

## How Has This Been Tested?
N/A


sections/futures/Trade/Trade.tsx
```diff
@@ -242,6 +242,8 @@ const Trade: React.FC<TradeProps> = () => {
 			<MarketInfoBox
 				availableMargin={futuresMarketsPosition?.remainingMargin ?? zeroBN}
 				buyingPower={sUSDBalance}
```
Tburm, line 244: I know this wasn't changed in this PR, but this value doesn't seem right to me. Shouldn't "buying power" refer to the amount you could invest given your deposited margin? I would think if "available margin" is 0 this should also be 0, but it shows your sUSD balance.

Tburm (approved):
:+1:

Tburm (approved):
Pretty big one but changes all look good and testing on the branch works as expected 👍🏼

Issue #321, opened 2022-02-02
Settle fee reclamations if >12 trades


platschi:
### Subject of the issue
There is an edge case situation on L1 when a trader trades 12 consecutive trades into a specific synth, the trader needs to settle the rebate in order to be able to trade a 13th attempt. The system will automatically disallow any further trading attempts if the trader does not settle. Currently the Kwenta UI leaves the trader in the dark might this situation kick in.

More information see [SIP-37](https://sips.synthetix.io/sips/sip-37) and [Kwenta Docs](https://docs.kwenta.io/resources/fee-reclamation).

See also for more information on the bug report [this Discord conversation](https://discord.com/channels/852273007370960937/852530414287454278/933826672313127013).

### Proposed solution
 - Instead of displaying our usual trade submission section, the interface should show a message and ask the trader to settle. 
 
The logic here requires a query to `Exchanger.settlementOwing` with the traders address and output synth to retrieve the `numEntries`. If `numEntries >= 12` require the trader to settle.

- Example message: `You have 12 consecutive trades into <SYNTH_NAME_HERE>. Please trade out of <SYNTH_NAME_HERE> or settle: (Settle Button)`.
- Add a "Learn more" link sending the trader to [this](https://docs.kwenta.io/resources/fee-reclamation) Kwenta Docs article

The format should be similar to the `NoSynthsCard`. Design mockup example:

![rebate_design](https://user-images.githubusercontent.com/548702/152077051-fd88304c-b790-442d-acf4-d2ac8522f54c.png)

### Steps to reproduce
- On L1 Kovan, trade into the same synth 12 times in order to test this issue.

### Rewards
75 SNX

platschi:
Assigned to @kethcode

kethcode:
I've made some work on this.  feel free to use some or none of it.  https://github.com/kethcode/kwenta/tree/fix/settle_consecutive_trades

PR #386, opened 2022-03-04
User can settle synths explicitly by UI

LeifuChen:
## Description
On L1, when one trades 12 consecutive trades into a specific synth, it is required to settle this synth in order to make the 13th attempt. The system will automatically block any further attempts before the settlement. We add a "settle" button and a user prompt (also a link to the doc of fee reclamation) to allow the user to settle synths by themselves on the Kwenta UI.

## Related issue
#321 

## Motivation and Context

Before this change, the user can't settle synths by themselves on the Kwenta UI.

## How Has This Been Tested?

1. User swaps from sUSD to sJPY 12 times
2. The UI shows a message "You have 12 consecutive trades into sJPY. Please trade out of sJPY or settle:” and a button "Settle". (There is a waiting period after completing the trade. After that, user can attempt to settle) 
4. Click on the button "Settle"
5. A pop-up showing "Settle sJPY" and a transaction is submitted to the user's wallet.
6. User confirms the transaction.
7. After a few minutes, the user can swap from sUSD into sJPY again.

message
image
metamask


LeifuChen:
I have tested it twice, and the code is working. However,  I have the following comments about the implementation.
1. The current UI is unable to refresh automatically after the settlement. Instead, it requires manual refresh (e.g., click the arrows in the middle). So I need some help to make it better.
2. Do we want to show the "submitting order" after clicking the "settle" button? The button text won't change right now. 

I hope I can have some input from your side. Thanks.

sections/exchange/hooks/useExchange.tsx
```diff
@@ -718,6 +727,46 @@ const useExchange = ({
...
+				const tx = await synthetixjs.contracts.Exchanger.settle(
+					walletAddress,
+					ethers.utils.formatBytes32String(baseCurrencyKey as string),
+					gas
+				);
+
+				if (tx != null) {
+					monitorTransaction({
+						txHash: tx.hash,
+						onTxConfirmed: () => {
```
JChiaramonte7, line 754: > The current UI is unable to refresh automatically after the settlement. Instead, it requires manual refresh (e.g., click the arrows in the middle). So I need some help to make it better.
Do we want to show the "submitting order" after clicking the "settle" button? The button text won't change right now.

In `onTxConfirmed` you could also try to invalidate the `useNumEntriesQuery` query. This should hopefully return 0 now that all transactions have been settled.

https://react-query.tanstack.com/guides/query-invalidation

JChiaramonte7 (commented):
Tested and works great!

sections/exchange/hooks/useExchange.tsx
LeifuChen: @JChiaramonte7 Thanks for your input. It reminds me to use `refetch()` to update the state before and after the settlement. The new change has been pushed.

LeifuChen:
The countdown timer of the settlement waiting period has been added.
image

PR #391, opened 2022-03-08
Futures Position Table

Tburm:
Adding the futures position table to the dashboard page

## Description
* Edit hook to pull current positions from the subgraph
* Create a table on the Dashboard page to display this data
* Create components to represent PnL properly

This component is reusable, and will be added to the "Market" page when changes are merged


queries/futures/utils.ts
```diff
@@ -199,6 +193,13 @@ export const mapTradeHistory = (futuresPositions: any[]): PositionHistory[] => {
 					};
 				}
 			)
+			.filter(({ isOpen }: { isOpen: boolean }) => {
```
koredefashokun, line 196: Just a nit, but I think we can do something like `=> openOnly ? isOpen : true;`

koredefashokun (approved):
Looks good. Just left a few small comments. Also not sure if @asherism intends for there to be an empty state for the tables, or we can just leave the table header. In the case of the latter, maybe we should remove the empty row:
image

asherism:
> Was thinking the same about the empty table. I tried a couple things to remove the table when empty but none of them made sense, I'll ask Yashar.

i think we can just do an empty entry with dashes for data

Issue #392, opened 2022-03-09
Switch from L1 to L2 does not recognize ETH wallet balance


platschi:
### Subject of the issue
When switching on Kwenta Exchange from L1 to L2, the application does not recognize the users ETH wallet balance but instead shows the `Not enough L2 ETH for transaction fees` modal. This modal should not be shown if the wallet has ETH in it. 

See screenshot, connected wallet on L2 with ETH is not recognized and a page refresh is necessary to have it read by the Kwenta dApp.

<img width="1231" alt="Screen Shot 2022-03-09 at 13 11 27" src="https://user-images.githubusercontent.com/548702/157482566-0f6f5232-a56d-4d43-9886-64ca03e9d9c3.png">

### Your environment
* L1 Mainnet switch to L2 Mainnet via MetaMask
* Firefox

### Steps to reproduce
0. Make sure you have a wallet with some ETH on both L1 and L2, and be connected to Ethereum L1.
1. Visit the [Kwenta Exchange page](https://kwenta.io/exchange), connect your L1 wallet. 
2. Click the `Switch to L2` button.
3. Note that the modal shows `Not enough L2 ETH for transaction fees. Get some via Optimism Bridge:` even though the L2 wallet has some ETH.

### Reward
25 SNX

PR #398, opened 2022-03-12
fix L2Gas check ETH balance

burtrico:
[https://community.optimism.io/docs/developers/build/differences/#using-eth-in-contracts](https://community.optimism.io/docs/developers/build/differences/#using-eth-in-contracts)
Per the Optimism Docs above, ETH was previously represented as an ERC20 on Optimism, but now functions the same as ETH on L1. In order to access the balance of ETH on L2, developers now can access the L2 OVM_ETH contract at [0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000](https://optimistic.etherscan.io/address/0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000). Note that the document does mention that the contract should not be called directly or it will throw an error. Using the current setup in this commit though, it appears to be loading the user's ETH balance correctly and not throwing an error.

## Description
Inside L2Gas.tsx under Containers, the L2WETH contract was imported to check the balance of a user's ETH for L2. Since after the OVM 2.0 update, the OVM_ETH contract should now be used to check for a user's ETH balance. After adding the L2OVMETH contract address and ABI to the kwenta repo, inside L2Gas.tsx, I changed the L2wETH contract to the L2ovmETH contract. 

## Related issue
(https://github.com/Kwenta/kwenta/issues/392)

## Motivation and Context
When switching from L1 to L2, a user's eth balance is not recognized until a hard refresh.

## How Has This Been Tested?
Using a brave browser and metamask wallet connected, I switched between L1 & L2 networks. My wallet had ETH on both L1 & L2. The hasNoBalance ETH message was no longer displayed on L2 after using the OVM_contract to check for a user's ETH balance. 

Screen Shot 2022-03-11 at 7 04 15 PM


JChiaramonte7:
Hmm this does look like it did fix the issue, but I'm not sure why we had to introduce a new ABI for L2 OVM_ETH? It looks like the contract is essentially still an ERC20 (and mostly the same as L2 WETH) so the previous ABI should have worked (just needed to set the new address). 

If Optimistic ETH is native now shouldn't we just be able to look up a balance with: https://docs.ethers.io/v5/api/providers/provider/#Provider-getBalance

burtrico:
> Hmm this does look like it did fix the issue, but I'm not sure why we had to introduce a new ABI for L2 OVM_ETH? It looks like the contract is essentially still an ERC20 (and mostly the same as L2 WETH) so the previous ABI should have worked (just needed to set the new address).
> 
> If Optimistic ETH is native now shouldn't we just be able to look up a balance with: https://docs.ethers.io/v5/api/providers/provider/#Provider-getBalance

Very good point. I'll update this to use provider.getBalance(), update the other code in L2Gas, and remove the L2OVM_ETH contract since its not necessary.

burtrico:
-> L2Gas.tsx now uses provider.getBalance(address) to find a user's balance
-> I'm not sure yet how to track a wallet's ETH transfers since there is no longer an ERC20 contract (WETH) associated with OE ETH that shows transfer events. The OVM_ETH currently inside the subscribe function of L2Gas does not track transfer events. We should be able to use the method that we use on L1, but I have not found that code yet.

JChiaramonte7 (approved):
ty @burtrico looks cleaned up

PR #400, opened 2022-03-14
Shorting e2e test

raffiegang:
## Description
This is an end-2-end automated test opening a short. The subgraphs are queried to determine the short has been  opened succesfully. It does so in a way that after submitting the transaction to short the subgraph should eventually update within 1 minute.


tsconfig.json
```diff
@@ -15,6 +15,6 @@
 		"jsx": "preserve",
 		"baseUrl": "."
 	},
-	"exclude": ["node_modules"],
+	"exclude": ["node_modules", "tests/e2e/helper/*.ts"],
```
JChiaramonte7, line 18: Reason for exclusion?

JChiaramonte7:
Heads up @raffiegang @platschi this is still being reviewed, just off of `Kwenta:test/shorting-e2e`

tsconfig.json
raffiegang: Not to break the default build . Without this removal the default build will not run.

PR #404, opened 2022-03-15
Futures fixes

koredefashokun:
## Description
This PR makes general fixes to the futures screen. It includes:
- Linting fixes on some files
- Linking of 24H price change data, replacing dummy data
- Highlighting the fact that there's currently no linked data for fee information on the position card
- Leverage input and slider fixes
- Max layout width for futures page (will be added to other pages in the near future)
- Correct unrealized PnL data (matches version used by @Tburm on the dashboard)
- Accurate average entry information on the position card
- Updating synthetix packages to `2.64.0-alpha`
- Adjust queries to match new function names from contracts-interface

## Related issue
N/A

## Motivation and Context
This is one of the final fixes to prepare the Kwenta v2 soft launch.

## How Has This Been Tested?
N/A


Tburm:
I'm seeing unrealized PnL even though I don't have a position open
image

koredefashokun:
@Tburm Pulled it from the data we have on the dashboard:
image

Tburm:
@koredefashokun I see, the reason is most likely that the subgraph is still pointing at the old contracts then.

sections/futures/UserInfo/UserInfo.tsx
```diff
@@ -105,7 +105,6 @@ const UserInfo: React.FC<UserInfoProps> = ({ marketAsset }) => {
 			<TabPanel name={FuturesTab.POSITION} activeTab={activeTab}>
 				<PositionCard
 					position={futuresMarketsPosition ?? null}
-					positionHistory={positionHistory ?? null}
```
Tburm, line 108: The position history is blank now, I'm assuming it was this change. Is there a reason to remove this?

Tburm (commented):
* 24hr change at the top doesn't update when switching markets, only on refresh
* "24 hr funding" is probably displaying instantaneous funding? We should make blank or change the title

sections/futures/UserInfo/UserInfo.tsx
koredefashokun: The position history data I was pulling was incorrect. It was only used to calculate average entry price, which we already had from the subgraph (I think).

Tburm: Ok, we will get this from the subgraph long-term. Can you disable this tab/button?

Tburm (approved):

Issue #402, opened 2022-03-14
Data fixes for futures


JChiaramonte7:
- [x] Getting current portfolio value
- [x] Unrealized PnL needs to include fees? and funding
- [x] Pull current positions directly from contract

PR #407, opened 2022-03-15
Dashboard portfolio

Tburm:
Adding a hook to pull futures portfolio value, and display on dashboard

## Description
* Portfolio value hook - this value includes all margin values plus any unrealized gains
* Display portfolio value on dashboard

## Related issue
[Issue](https://github.com/Kwenta/kwenta/issues/402)


sections/dashboard/PortfolioChart/PortfolioChart.tsx
```diff
@@ -1,12 +1,33 @@
...
 	const { t } = useTranslation();
 
+	const markets = futuresMarkets.map((market: FuturesMarket) => {return market.asset})
+	const portfolioValueQuery = useGetCurrentPortfolioValue(markets);
+	const portfolioValue = portfolioValueQuery?.data ?? null;
+
 	return (
-		<Chart />
+		<Chart>
+			<PortfolioTitle>Futures Portfolio Value</PortfolioTitle>
```
koredefashokun, line 23: I think this says "Portfolio Value" on the designs, but I guess this is specific for a reason, right?

queries/futures/useGetCurrentPortfolioValue.ts
```diff
@@ -0,0 +1,61 @@
...
+		QUERY_KEYS.Futures.Positions(markets || [], walletAddress || ''),
+		async () => {
+			const {
+				contracts: { FuturesMarketData },
+			} = synthetixjs!;
+			if (!markets) return null;
+
+			try {
+				const positionsForMarkets = await Promise.all(
+					(markets as [string]).map((market: string) =>
```
koredefashokun, line 32: Not sure if this is intentional, but I think this might be interpreted by Typescript as an array with one string. For multiple markets, I think we should do `string[]` instead of `[string]`.

sections/dashboard/PortfolioChart/PortfolioChart.tsx
Tburm: Yes, this is intentional. I think eventually this will say "Portfolio Value" but for now this only includes futures. I think eventually it will cover shorts and spot assets, so want to be more specific for now.

queries/futures/useGetCurrentPortfolioValue.ts
Tburm: fixed

koredefashokun (approved):

Issue #359, opened 2022-02-18
Dynamic shortable asset list


JChiaramonte7:
### Feature Proposal
Generate shortable asset list for each network.  

### Rationale
Currently the shortable synth asset list is hardcoded.
https://github.com/Kwenta/kwenta/blob/e84be821ca0db2ecaaf4a53221eb380b4752ea06/sections/shorting/constants.ts#L3-L4

### Implementation
- Look at the current list of synths and check if each asset is short-able with `CollateralManager.shortableSynthsByKey(CURRENCY_KEY)` which returns non-zero if the asset is shortable. 
-   Completely replace `SYNTHS_TO_SHORT` and `SYNTHS_TO_SHORT_L1`
- Use [multicall](https://github.com/Destiner/ethcall) to batch the calls.

### Rewards
150 SNX

platschi:
Assigned to @kethcode

kethcode:
CollateralManager.shortableSynthsByKey isn't available on the L1 contract. I'm going to leave the L1 list hardcoded, since L1 shorts are deprecated anyway.

kethcode:
Ok, I've got an alternate approach for the L1 that will handle replacing it fine.

kethcode:
Here's you're hook.  I can't seem to navigate the async, so I'm done.  Give it to someone else.

	import { CurrencyKey, Synths } from 'constants/currency';
import Connector from 'containers/Connector';
import { ethers } from 'ethers';

export const getSynthShortList = async () => {
	const { synthetixjs } = Connector.useContainer();
	let SYNTHS_TO_SHORT_L1 = [Synths.sBTC, Synths.sETH, Synths.sLINK];
	let SYNTHS_TO_SHORT: CurrencyKey[] = [];

	for(var synth in Synths) {
		var retval = parseInt(await synthetixjs!.contracts.CollateralManager.shortableSynthsByKey(ethers.utils.formatBytes32String(synth)))
		if(retval > 0)
		{
			SYNTHS_TO_SHORT.push(synth as CurrencyKey)
		}
	}

	console.log("SYNTHS_TO_SHORT:", SYNTHS_TO_SHORT)

	return { SYNTHS_TO_SHORT_L1, SYNTHS_TO_SHORT };

};

PR #409, opened 2022-03-16
Dynamic Shortable Asset List

kethcode:
##### useGetShortableSynths.ts
Added a new query. Updated ethcall to 4.3.6 to enable OP support.  L2 Synths are retrieved from synthetixjs using multicall, memoed into each file that uses it.  L1 list is hard coded into the query since they are deprecated, and the L1 contract gives us no method to retrieve them dynamically. 

##### ShortingStats.tsx
Replaced SYNTHS_TO_SHORT and SYNTHS_TO_SHORT_L1 with a single query that populates SYNTHS_TO_SHORT, and updated hooked memos.

##### useFeeReclaimPeriodsQuery.ts
new version of ethcall errors when passed {} parameter. It also returns unknown[] instead of [any,any,any], like the previous version did, so I cast the result.

##### useRedeemableDeprecatedSynthsQuery.ts
new version of ethcall errors when passed {} parameter

##### useTokensBalancesQuery.ts
new version of ethcall errors when passed {} parameter

##### Resolved Various Lint Errors that somehow made it into dev (came down with my upstream fetch)

```
/home/kethic/archive/kwenta/components/TVChart/TVChart.tsx
  85:5  warning  React Hook React.useEffect has missing dependencies: 'autosize', 'containerId', 'fullscreen', 'interval', 'libraryPath', 'overrides', and 'studiesOverrides'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
```

```
/home/kethic/archive/kwenta/sections/exchange/TradeCard/Cards/CombinedMarketDetailsCard.tsx
   2:25  warning  'useMemo' is defined but never used                                                                                         @typescript-eslint/no-unused-vars
  50:5   warning  React Hook useEffect has a missing dependency: 'rates24hQuery.isSuccess'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

```
/home/kethic/archive/kwenta/sections/exchange/hooks/useExchange.tsx
  541:3  warning  React Hook useMemo has an unnecessary dependency: 'isL2'. Either exclude it or remove the dependency array  react-hooks/exhaustive-deps
```

## Related issue
https://github.com/Kwenta/kwenta/issues/359

## Motivation and Context
Allow for automated deployment of new synthetix synths to kwenta shorting market

##### ShortingHistory.tsx
The right-side drop-down populates with the correct L2 synths. I opened and closed positions on every L2 synth, confirming it was selectable, and showed up correctly under Your Positions.

##### ShortingReward.tsx
This is only used on L1.  Verified the synth reward list is correct for L1 after the change.

##### ShortingStats.tsx
This is used on L1 and L2.  Verified the synth list displayed is correct on both L1 and L2 after the change.

##### useFeeReclaimPeriodsQuery.ts
I don't know how to test this one other than the console.log commands I used during dev.

##### useRedeemableDeprecatedSynthsQuery.ts
DashboardCard uses this.  Synth list appears to still work after the change.

##### useTokensBalancesQuery.ts
This is used in useExchange and SelectTokenModal.  Both of these appear to still work after the change.


kethcode:
I don't know what to do to resolve these checks.  they never pass, and it concerns me.

components/TVChart/TVChart.tsx
```diff
@@ -82,7 +82,17 @@ export function TVChart({
...
+	}, [
+		baseCurrencyKey,
+		quoteCurrencyKey,
+		autosize,
+		containerId,
+		fullscreen,
+		interval,
+		libraryPath,
+		overrides,
+		studiesOverrides,
```
koredefashokun, line 94: Adding these causes some flickering on the `TVChart` component when it is refocused.

queries/synths/useFeeReclaimPeriodsQuery.ts
```diff
@@ -50,10 +50,9 @@ const useFeeReclaimPeriodsQuery = (
 			}
 
 			const waitingPeriodsRaw = (await ethcallProvider.all(
-				waitingPeriodCalls,
-				{}
+				waitingPeriodCalls
 			)) as ethers.BigNumberish[];
-			const feesRaw = await ethcallProvider.all(feeCalls, {});
+			const feesRaw = (await ethcallProvider.all(feeCalls)) as [any, any, any];
```
koredefashokun, line 55: Not sure if this type is correct. `settlementOwing` returns a tuple of three `uint` values. I think this should be:

```suggestion
			const feesRaw = (await ethcallProvider.all(feeCalls)) as [ethers.BigNumberish, ethers.BigNumberish, ethers.BigNumberish][];
```

koredefashokun:
@kethcode Sorry for the delayed review. Left a few comments above.

kethcode:
Thank you for the review. I will make these changes

queries/synths/useFeeReclaimPeriodsQuery.ts
kethcode: > Not sure if this type is correct. `settlementOwing` returns a tuple of three `uint` values. I think this should be:

this does not work.  tried it, and sub requires a number type, and throws an error on ethers.BigNumberish.  I understand the desire to type this, but whatever is coming back needs iterable, so it's not a Number either.  We can create a ticket to fix this later, if we want.  I've added a TODO for now

components/TVChart/TVChart.tsx
kethcode: I've removed these and added // eslint-disable-next-line react-hooks/exhaustive-deps

components/TVChart/TVChart.tsx
```diff
@@ -82,17 +82,8 @@ export function TVChart({
...
-		autosize,
-		containerId,
-		fullscreen,
-		interval,
-		libraryPath,
-		overrides,
-		studiesOverrides,
-	]);
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, []);
```
koredefashokun, line 86: I think we need the `baseCurrencyKey` and `quoteCurrencyKey` in the deps array here (only these two). Sorry for the confusion.

queries/synths/useFeeReclaimPeriodsQuery.ts
```diff
@@ -52,7 +52,7 @@ const useFeeReclaimPeriodsQuery = (
 			const waitingPeriodsRaw = (await ethcallProvider.all(
 				waitingPeriodCalls
 			)) as ethers.BigNumberish[];
-			const feesRaw = (await ethcallProvider.all(feeCalls)) as [any, any, any];
+			const feesRaw = (await ethcallProvider.all(feeCalls)) as [any, any, any]; // TODO: replace any with correct iterable type
```
koredefashokun, line 55: Sorry about the confusion here as well. Tested it locally and it should be:
```suggestion
			const feesRaw = (await ethcallProvider.all(feeCalls)) as [ethers.BigNumber, ethers.BigNumber, ethers.BigNumber][];
```

koredefashokun:
Also, it looks like there are conflicts in the package-lock.json. I think you should be able to fix them by deleting the file and re-running `npm i`. Only caveat is to make sure you're on npm v7.

kethcode:
@koredefashokun I saw that.  However, I need to upgrade ethcall to 4.3.6, so the changes need to occur.  Someone else just pushed a lock file with 3.2.0 in it since I submit the PR.  Should I pull the latest dev v1, merge and recommit?

queries/synths/useFeeReclaimPeriodsQuery.ts
kethcode: will fix

components/TVChart/TVChart.tsx
kethcode: will fix

kethcode:
@koredefashokun @platschi should be done now.

kethcode:
I give up.  Turning off my linter, deleting this PR, and I'll resubmit with a narrower scope.

PR #426, opened 2022-03-17
Fix issues with leverage

koredefashokun:
## Description
This PR fixes rounding issues with the leverage slider.

## Related issue
N/A

## Motivation and Context
This is part of the Kwenta v2 soft launch.

## How Has This Been Tested?
N/A

### DIFF
--- a/components/TVChart/TVChart.tsx
+++ b/components/TVChart/TVChart.tsx
@@ -39,7 +39,7 @@ export function TVChart({
 			symbol: baseCurrencyKey + ':' + quoteCurrencyKey,
 			datafeed: DataFeed,
 			interval: interval,
-			container_id: containerId,
+			container: containerId,
 			library_path: libraryPath,
 
 			locale: 'en',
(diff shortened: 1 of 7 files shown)


Tburm:
I'm sort of confused by these changes. It works perfectly when you do not have a position open, but when you have an open position the leverage values stops making sense.

For example, if you have a long position open and want to go short, the leverage slider will be >10x
image

That said, the "Confirm order" modal displays the correct final position, so I think it's just a decision of how we display it on this page.

koredefashokun:
Interesting observations @Tburm. I think you are right. Didn't notice it since I was testing on a market that didn't have an open position. However, this is using the calculations that we have for max leverage. Does this mean there's something wrong with the way we're making these calculations?

Tburm:
@koredefashokun No, the calculations are correct. It's just an odd way to display it. Instead of saying "you're going 16x short" We should do something like Ralf mentioned:

`2.5x --> 6x` where each leverage amount is colored based on position size (going from 2.5x long to 6x short)

Tburm (approved):

Issue #412, opened 2022-03-16
v2: High severity issues 


platschi:
## Unrealized PnL looks weird (@Tburm)
 
- [x] Done

![pnl1](https://user-images.githubusercontent.com/548702/158634855-3d6a2951-0100-4057-9ea5-2f4373a7150b.png)
![pnl2](https://user-images.githubusercontent.com/548702/158634894-7142b457-a40d-4905-ba3d-42104876fbe3.png)

## Buying Power & Margin Usage data incorrect (@Tburm)

- [x] Done

![margin1](https://user-images.githubusercontent.com/548702/158635261-ddc4f4ca-dbd2-4721-9399-ab42e8cd16ea.png)
![margin2](https://user-images.githubusercontent.com/548702/158635273-e28815ee-68cb-4e95-9a2e-373e185869ae.png)

## No error on orders that are way too big / will fail, just fails silently without sending (@koredefashokun)

- [x] Done

Trying an order that's too big just shows 0 order instead of error / notification, and nothing happens on "confirm order”

![0order](https://user-images.githubusercontent.com/548702/158635665-939425ba-b508-4be1-938d-f1b3922aec61.png)

## Closed position is showing as open (but with 0 values), same for no position (first time using a market)

- [x] Done

![note1](https://user-images.githubusercontent.com/548702/158635890-aecf610d-848f-41a6-ba0b-f25519f65645.png)

## Submits failing tx for just over-leverage (@koredefashokun)

- [x] Done

An order that will fail is being submitted to fail instead of failing after RPC error for estimateGas as usual. E.g. an order for 11x leverage. This is different from the 1000x leverage that will fail silently and won’t submit. Both conditions should notify the user that the order will fail because of (revert reason).

## Volume is counted incorrectly, shows negative for negative orders

- [x] Done

Volume is negative, but should be summed in abs() values for both negative and positive orders

![vol1](https://user-images.githubusercontent.com/548702/158636252-41d9ee1f-2256-4330-be11-b0324749d29d.png)

## Open Interest doesn't show sign/skew (@Tburm)

- [x] Done

negative / positive skew should be visible

## Open Interest in Dashboard is showing wrong value (@Tburm)

- [x] Done

Open interest $ value is showing the same number as volume and is incorrect in this case. 

![vol2](https://user-images.githubusercontent.com/548702/158636576-5e5bf5f2-e9c9-4ac7-a2b5-49da13a3b4e5.png)
[...]

PR #430, opened 2022-03-17
Margin Data Fixes

Tburm:
Correcting the data displayed about margin on the markets page

## Description
* Display "total" and "available" margin separately
* Correct the buying power calculation
* Correct the margin usage calculation
* Calculate gas price on deposit/withdraw modal open
* Avoid setting gas to null before resetting on modal changes

## Related issue
[Issue](https://github.com/Kwenta/kwenta/issues/412)

image

### DIFF
--- a/components/BaseModal/BaseModal.tsx
+++ b/components/BaseModal/BaseModal.tsx
@@ -94,7 +94,7 @@ const DismissButton = styled.button`
 	${resetButtonCSS};
 	position: absolute;
 	right: 20px;
-	color: ${(props) => props.theme.colors.blueberry};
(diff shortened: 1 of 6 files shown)


koredefashokun:
@Tburm Is it intended behaviour that the input on the deposit modal has a '0' in it before the user inputs anything?

Tburm:
@koredefashokun Yes, I did that intentionally. I like Having some placeholder there when they open the modal, and it also guarantees that we calculate a gas price to display immediately instead of waiting for an input.

koredefashokun (approved):

Issue #458, opened 2022-03-18
Long/Short Improvements


platschi:
- [x] Make sure clicking the buttons does not reset the amount input values
- [x] Move Long/Short buttons above Amount input

Issue #456, opened 2022-03-18
Txs failing: Out of gas 


platschi:
Users reporting failed transactions among the board, e.g. for depositing sUSD, opening positions etc.

- Add a gas buffer (buffer it by a 100k gas)

Issue #415, opened 2022-03-16
v2: Minimum margin requirement info


platschi:
### Subject of the issue
People may miss depositing margin, and be confused why nothing works, probably 0 margin should make every other action not available / display a notification on press.

The minimum margin needed to trade is $100. When depositing <100$ and/or having a margin <100$, the interface should show an information message regarding this fact.

### Temporary solution

- [x] disable place order button if margin below $100 and some sort of clear status message near it
- [x] display warning if depositing <100$

"Minimum margin to trade is 100$"

### Ideas:

- [x]  Show Info Box under 'Deposit' / Withdrawal boxes when Margin <100$
- [x] On Deposit Modal, show information message when (current margin + deposit input amount) <100$.

asherism:
modal: 
https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=1925%3A10763
<img width="600" alt="image" src="https://user-images.githubusercontent.com/11524787/159590190-a1b2e55a-2557-4155-8b04-6479b0c390c2.png">

futures ui:
https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=1935%3A12431
<img width="468" alt="image" src="https://user-images.githubusercontent.com/11524787/159590269-83d0e756-793f-460a-b252-3dcae897d12a.png">

PR #471, opened 2022-03-21
Add gas buffer

koredefashokun:
## Description

This PR makes a few changes to the market page. These include:
- Use gas normalization function to prevent gas limit issues.
- Moving the position buttons above the order sizing inputs.
- Adding a disclaimer regarding minimum order amounts to the deposit modal.
- Replacing dollar sign with `sUSD` prefix for margin values.

## Related issue
N/A

## Motivation and Context
This PR fixes issues raised in the soft launch of Kwenta v2.

## How Has This Been Tested?
N/A

image

image


platschi:
Relevant issues:

- #456 
- #458 
- #415

Tburm (approved):
Approving, but for clarity this will raise the gas limit by 5k as that's the default buffer we have set. If we want to change that amount it's set as a constant.

Issue #483, opened 2022-03-22
Deposit/Withdraw Margin Modal Improvements


platschi:
### Subject of the issue
Currently it is possible to enter random amounts and launch a deposit transaction. Depositing or withdrawing an amount higher than the Max wallet balance will result in a failed transaction.

Second, the initial value in the input field is 0. When clicking and entering an amount, the initial 0 will not be overwritten and typing 1 will result in input: 01.

### Deliverables
- [ ] Do not allow input of amounts > Max to avoid failed transactions on Deposit and Withdraw Margin Modals
- [ ] Remove placeholder value upon input  on Deposit and Withdraw Margin Modals

### Rewards
25 SNX

PR #499, opened 2022-03-23
483 deposit withdrawal modal

brossetti1:
## Description
- adding internationalization to deposit modal button (extra)
- adding placeholder for better initial value (replace 0 when user enters input)
- adding disabled ability to deposit button (Do not allow input of amounts > Max to avoid failed transactions)
- added internationalized error message display for values > sUSD value in user wallet
- added `cursor: not-allowed;` to button when `disabled`

## Related issue
https://github.com/Kwenta/kwenta/issues/483

## Motivation and Context

for the requirement - `replace 0 when user enters input`: 

Adding a placeholder value to the `CustomInput` allows a value to be seen in the input while also not conflicting with the input `value` when the user goes to type a value in.

for the requirement - `Do not allow input of amounts > Max to avoid failed transactions`:

I added a useEffect hook to take into consideration the values that should invalidate (disable) the button to trigger a transaction. the user has to input atleast `100` sUSD and cannot exceed the value of sUSD that is in their wallet. If either of these conditions fails, then the deposit button will be disabled and the user cannot submit a transaction (that might fail).

I added a message to replace the error that is seen when sUSD value in a user's wallet is exceeded.

## How Has This Been Tested?
* load the local dev environment and branch off of branch `v2-dev`
* make changes to modal
*  test UI that I cannot submit transactions when disabled conditions are disqualifing
* test the UI that I can trigger a transaction when the `amount` input value is greater than 100 sUSD AND less than the sUSD in the users wallet.


platschi:
Not good yet, see comments in ticket thread

platschi (approved):

sections/futures/Trade/DepositMarginModal.tsx
```diff
@@ -29,6 +29,9 @@ type DepositMarginModalProps = {
 	market: string | null;
 };
 
+const PLACEHOLDER = '$0.00';
+const MIN_DEPOSIT_AMOUNT = wei('100');
```
koredefashokun, line 33: I think with the SCCP that was just merged, this can change to this now:
```suggestion
const MIN_DEPOSIT_AMOUNT = wei('50');
```

koredefashokun: @brossetti1 Just one small change, but everything looks good.

PR #513, opened 2022-03-24
gas price summary component update re status l1/l2

keybig:
the component showing the gas fee text does a Logical OR check on `isMainnet` and `!isCustomGasPrice`  followed by a ternary operation `?:`

```
<SummaryItemLabel>
                {isMainnet || !isCustomGasPrice
                    ? t('exchange.summary-info.max-fee-gwei')
                    : t('exchange.summary-info.gas-price-gwei')}
            </SummaryItemLabel>
```

The reason the text is not updating with L2, is that the !isCustomGasPrice value returns true. This triggers the true ternary ? statement of Max Fee (GWEI).

I have updated the component with a simple ternary using `isL2`, and if true return the specified text, and if false return the specified text.

I have also updated the `en.json` file to change GWEI to Gwei, as requested.

## Related issue
While testing in local, running the ternary with `isMainnet` rather than `isL2` did not trigger the changes in the component necessary. I am not sure why this is, or if it had something to do with my local setup.

## How Has This Been Tested?
Visual testing to confirm the 

image

image

image

### DIFF
--- a/sections/exchange/FooterCard/TradeSummaryCard/GasPriceSummaryItem.tsx
+++ b/sections/exchange/FooterCard/TradeSummaryCard/GasPriceSummaryItem.tsx
@@ -66,7 +66,9 @@ const GasPriceSummaryItem: FC<GasPriceSummaryItemProps> = ({
 	return (
 		<SummaryItem {...rest}>
 			<SummaryItemLabel>
-				{isL2 ? t('exchange.summary-info.gas-price-gwei') : t('exchange.summary-info.max-fee-gwei')}
+				{isMainnet
+					? t('exchange.summary-info.max-fee-gwei')
+					: t('exchange.summary-info.gas-price-gwei')}
 			</SummaryItemLabel>
 			<SummaryItemValue>
 				{gasPrice != null ? (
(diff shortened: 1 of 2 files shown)


platschi (approved):

JChiaramonte7:
I believe only Mainnet has EIP-1559 and "Max Fee" (cc @brossetti1 ). So we should only show "Max Fee" on mainnet and not other testnets like Kovan.

sections/exchange/FooterCard/TradeSummaryCard/GasPriceSummaryItem.tsx
```diff
@@ -66,9 +66,7 @@ const GasPriceSummaryItem: FC<GasPriceSummaryItemProps> = ({
 	return (
 		<SummaryItem {...rest}>
 			<SummaryItemLabel>
-				{isMainnet || !isCustomGasPrice
-					? t('exchange.summary-info.max-fee-gwei')
-					: t('exchange.summary-info.gas-price-gwei')}
+				{isL2 ? t('exchange.summary-info.gas-price-gwei') : t('exchange.summary-info.max-fee-gwei')}
```
JChiaramonte7, line 69: Let's only show "Max Fee" on mainnet

JChiaramonte7 (changes requested):

platschi (approved):

JChiaramonte7 (approved):

Issue #518, opened 2022-03-24
Market refresh on price changes


platschi:
### Subject of the issue
When the price changes, the position card (Position Size, Unrealized P&L, Leverage, Liq.) doesn't change automatically. Users have to refresh the page manually in order to update the position card

PR #519, opened 2022-03-24
Let the position card change when the price moves

LeifuChen:
## Description
When the price moves, the position card (Position Size, Unrealized P&L, Leverage, Liq) doesn't change automatically. The user has to refresh the page to update the position card.

## Related issue
#518

## Motivation and Context
The position card (Position Size, Unrealized P&L, Leverage, Liq) is supposed to change automatically when the price moves. It will increase the user experience.

## How Has This Been Tested?
1. Open a position and take a screenshot of the position card
2. Wait for the price moves (around 5-10 mins)
3. Compare the position info on dapp with the screenshot taken above

### DIFF
--- a/sections/futures/UserInfo/UserInfo.tsx
+++ b/sections/futures/UserInfo/UserInfo.tsx
@@ -37,7 +37,6 @@ const UserInfo: React.FC<UserInfoProps> = ({ marketAsset }) => {
 	const futuresMarketPositionQuery = useGetFuturesPositionForMarket(marketAsset);
 	const futuresPositionHistoryQuery = useGetFuturesPositionHistory(marketAsset);
 
-
 	const exchangeRates = useMemo(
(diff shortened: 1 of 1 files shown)


platschi:
Vercel Preview version does not show `PositionCard` data for my open positions.

Screen Shot 2022-03-24 at 18 17 45

LeifuChen:
> Vercel Preview version does not show `PositionCard` data for my open positions.
Thanks for testing. The open position issue has been fixed.

Tburm (changes requested):
My testing did not result in refreshing of positions on price changes. The variable storing the rate does not update, so unless we also update the `marketAssetRate` value when the price updates, the position card will not update.

JChiaramonte7:
Yeah this issue can be resolved at the moment. We don't have a mechanism to detect oracle updates at the moment

LeifuChen:
#666 fixed the reported issue.

Issue #476, opened 2022-03-21
Disable Input when margin = 0


platschi:
### Improvements
On the Markets page, if a user has a margin of 0$, input fields for amounts and leverage should be disabled. The same goes for the Withdraw button, given there's nothing to withdraw.

### Deliverables

- [x] If nothing is deposited (total margin = 0 sUSD), disable inputs for Amounts and Leverage
- [x] Disable "Withdraw" button (greyed out) if margin = 0 sUSD

### Rewards
25 SNX

PR #531, opened 2022-03-28
Disable Order Input and Withdraw Button when margin = 0

Flocast:
## Description

Adding a disable attribute with a condition of total margin = 0 for the Withdraw button and for the Order sizing inputs.
## Related issue

https://github.com/Kwenta/kwenta/issues/476
## Motivation and Context

On the Markets page, if a user has a margin of 0$, input fields for amounts and leverage should be disabled. The same goes for the Withdraw button, given there's nothing to withdraw.

## How Has This Been Tested?

Verified that inputs are disabled when there is no margin deposited.
Deposited margin and verified that inputs worked properly.

Using Firefox and Op Kovan.


sections/futures/Trade/Trade.tsx
```diff
@@ -243,7 +243,10 @@ const Trade: React.FC<TradeProps> = () => {
 				<MarketActionButton onClick={() => setIsDepositMarginModalOpen(true)}>
 					Deposit
 				</MarketActionButton>
-				<MarketActionButton onClick={() => setIsWithdrawMarginModalOpen(true)}>
+				<MarketActionButton
+					disabled={futuresMarketsPosition?.remainingMargin! <= zeroBN}
```
koredefashokun, line 247: Since remainingMargin is of the `Wei` type, I think this should be:
```suggestion
					disabled={futuresMarketsPosition?.remainingMargin?.lte(zeroBN)}
```

sections/futures/Trade/Trade.tsx
```diff
@@ -272,6 +275,7 @@ const Trade: React.FC<TradeProps> = () => {
 			<PositionButtons selected={leverageSide} onSelect={setLeverageSide} />
 
 			<OrderSizing
+				disabled={futuresMarketsPosition?.remainingMargin! <= zeroBN}
```
koredefashokun, line 278: Same here. I believe this should be:
```suggestion
				disabled={futuresMarketsPosition?.remainingMargin?.lte(zeroBN)}
```

koredefashokun (approved):

koredefashokun (commented):
This looks good, except for the fact that I think we should also disable the inputs when a wallet is not connected. What do you think @platschi @JChiaramonte7?

JChiaramonte7:
> This looks good, except for the fact that I think we should also disable the inputs when a wallet is not connected. What do you think @platschi @JChiaramonte7?

Separate ticket but yes we can do that 👍

platschi:
> This looks good, except for the fact that I think we should also disable the inputs when a wallet is not connected. What do you think @platschi @JChiaramonte7?

Agree with Jeremy, will create separate ticket.

koredefashokun (approved):

PR #535, opened 2022-03-28
Fix 24H Volumes

Tburm:
24 hour volume does not match between the dashboard and markets pages. 

## Description
The graph defaults to returning 100 records. On the dashboard we query all markets, so if there are over 100 trades in 24 hours we are only summarizing a subset of trades.

This change increases the limit to 1000 (the max for the graph) which will buy us time. Eventually when we have >1000 trades across all markets we will need to update our logic to query the graph.

## Related issue
No related issues

## How Has This Been Tested?
Compared volumes across markets and dashboard pages to ensure they match


koredefashokun (approved):

JChiaramonte7 (approved):

JChiaramonte7:
This can be merged for today's release

Issue #448, opened 2022-03-18
Clarify Funding Rate in MarketDetails


platschi:
- We need to come up with a solution to name Funding % correctly, adding e.g. 24H unit would not be entirely correct.

### Deliverables

- [x] Calculate Funding Rate Avg for 24h
- [x] Decide what rate we want to show, e.g. 8h or 24h time frame

PR #536, opened 2022-03-28
24 Hour Funding Rate

Tburm:
Calculate the 24 hour funding rate from the `FundingRateUpdated` events

## Description
We are currently displaying the instantaneous funding rate. This rate is volatile, so we should instead display an average over the past 24 hours.

## Related issue
Issue #448


queries/futures/utils.ts
```diff
@@ -171,6 +173,17 @@ export const calculateDailyTradeStats = (futuresTrades: FuturesOneMinuteStat[])
...
 };
 
+export const calculateFundingRate = (minFunding: FundingRateUpdate, maxFunding: FundingRateUpdate): Wei | null => {
+	if (!minFunding || !maxFunding) return null;
+
+	// clean values
+	const fundingStart = wei(minFunding.funding)
+	const fundingEnd = wei(maxFunding.funding)
+
+	const avgFundingRate = fundingEnd.sub(fundingStart).div(fundingStart)
```
artdgn, line 183: I think this should be divided by time difference instead: 
```js
const fundingDiff = fundingEnd.sub(fundingStart); // funding is already in ratio units
const timeDiff = maxFunding.timestamp - minFunding.timestamp; // I'm guessing some type and unit conversion needs to happen here for this to actually work

if (timeDiff === 0) {
  return null; // use fallback instanteneous value, or different a calculation that uses an earlier event
}

return fundingDiff.mul(SECONDS_PER_DAY).div(timeDiff); // convert to 24h period
```

sections/futures/MarketDetails/MarketDetails.tsx
```diff
@@ -146,12 +150,12 @@ const MarketDetails: React.FC<MarketDetailsProps> = ({ baseCurrencyKey }) => {
 				),
 			},
 			'Funding Rate': {
```
artdgn, line 152: I think this should be `'Funding Rate 24h'`, because 24h is not the usual default value (and there is no agreed time unit for this AFAIK). Maybe even `Funding 24h` would be acceptable if short version needed.

A detailed tooltip is still in order I think, but adding the time unit is the very minimum IMO.

sections/futures/MarketDetails/MarketDetails.tsx
```diff
@@ -146,12 +150,12 @@ const MarketDetails: React.FC<MarketDetailsProps> = ({ baseCurrencyKey }) => {
...
-				value: marketSummary?.currentFundingRate
-					? formatPercent(marketSummary?.currentFundingRate ?? zeroBN, { minDecimals: 6 })
+				value: avgFundingRate
+					? formatPercent(avgFundingRate ?? zeroBN, { minDecimals: 6 })
 					: NO_VALUE,
```
artdgn, line 155: This should still fallback to `marketSummary?.currentFundingRate` IMO in case two events in the period aren't available. In which case it should also add some indication that a different value is being used (e.g. `(inst)` suffix) or something.

artdgn (commented):
Added comments for calculation, using fallback value, and edge case handling if only event is available.

JChiaramonte7 (approved):

sections/futures/MarketDetails/MarketDetails.tsx
Tburm: These issues will be addressed in Issue #643

PR #564, opened 2022-03-30
dynamic list of shortable assets

kethcode:
## Description
Built against `dev` but would not compile unless I ran `npm i` against `v2-dev`.  

##### useGetShortableSynths.ts
Added a new query. Updated `ethcall` to 4.4.0 to enable OP support.  L2 Synths are retrieved from synthetixjs using multicall, memoed into each file that uses it.  L1 list is hard coded into the query since they are deprecated, and the L1 contract gives us no method to retrieve them dynamically. 

##### ShortingStats.tsx
Replaced `SYNTHS_TO_SHORT` and `SYNTHS_TO_SHORT_L1` with a single query that populates `SYNTHS_TO_SHORT` based on L1 or L2 being currently active, and updated hooked memos.

##### useFeeReclaimPeriodsQuery.ts
new version of `ethcall` errors when passed {} parameter. It also returns unknown[] instead of [any,any,any], like the previous version did, so I cast the result.

##### useRedeemableDeprecatedSynthsQuery.ts
new version of `ethcall` errors when passed {} parameter

##### useTokensBalancesQuery.ts
new version of `ethcall` errors when passed {} parameter

## Related issue
https://github.com/Kwenta/kwenta/issues/359

## Motivation and Context
Allow for automated deployment of new synthetix synths to kwenta shorting market

##### ShortingHistory.tsx
The right-side drop-down populates with the correct L2 synths. I opened and closed positions on every L2 synth, confirming it was selectable, and showed up correctly under Your Positions.

##### ShortingReward.tsx
This is only used on L1.  Verified the synth reward list is correct for L1 after the change.

##### ShortingStats.tsx
This is used on L1 and L2.  Verified the synth list displayed is correct on both L1 and L2 after the change.

##### useFeeReclaimPeriodsQuery.ts
I don't know how to test this one other than the console.log commands I used during dev.

##### useRedeemableDeprecatedSynthsQuery.ts
DashboardCard uses this.  Synth list appears to still work after the change.

##### useTokensBalancesQuery.ts
This is used in useExchange and SelectTokenModal.  Both of these appear to still work after the change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
 		"bnc-onboard": "1.37.3",
 		"date-fns": "2.21.3",
 		"date-fns-tz": "1.1.4",
-		"ethcall": "3.2.0",
+		"ethcall": "4.4.0",
 		"ethers": "5.5.0",
 		"framer-motion": "4.1.17",
 		"graphql-request": "3.4.0",
--- a/queries/synths/useFeeReclaimPeriodsQuery.ts
+++ b/queries/synths/useFeeReclaimPeriodsQuery.ts
@@ -50,10 +50,14 @@ const useFeeReclaimPeriodsQuery = (
 			}
 
 			const waitingPeriodsRaw = (await ethcallProvider.all(
-				waitingPeriodCalls,
-				{}
+				waitingPeriodCalls
 			)) as ethers.BigNumberish[];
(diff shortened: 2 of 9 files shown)


koredefashokun (approved):

koredefashokun:
Tested this and it works correctly.

Issue #441, opened 2022-03-18
Add Futures Positions to Market Page


platschi:
### Subject of the issue
Community members feedback made it clear that it is desired to see a table of open positions in other markets on the market page. 

### Desired layout
<img width="915" alt="Screen Shot 2022-03-29 at 19 34 31" src="https://user-images.githubusercontent.com/548702/160717934-74dae85d-bf96-42f0-8413-3dbb9f8554fc.png">

### Deliverables

- [x] Add `FuturesPositionsTable` to Markets Page below Open Positions box.
- [x] Make sure to show other positions you have open in other markets, but not the market you are on

E.g. if you have BTC, LINK and ETH positions open, and are visiting the ETH market, show in this table only BTC and LINK open positions. If you are on the LINK market page, hide the LINK position in this table, and so on.

### Rewards
50 SNX

asherism:
https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=1560%3A9243
<img width="1221" alt="image" src="https://user-images.githubusercontent.com/11524787/159749133-f6c2198d-06a5-47ed-b45d-c074dec5f8be.png">

PR #587, opened 2022-04-04
added FuturesPositionsTable to MarketInfo

burtrico:
Inside of the Markets page, I added a futures positions table for the user's positions in all markets (excluding the current market since that is already shown on the page). 

## Description
From the dashboard overview file, I copied the futuresPositionsTable from the return value and related futures market query code. I then added this to the MarketInfo file, placing the futuresPositionsTable below the UserInfo (current markets positions for the user). I then filtered out the current market from the futuresPositionsTable. This ends up displaying all of the user's futures positions inside of the Markets page.

Note that assuming the user has no other positions open, the blank "other futures positions" table will currently be displayed.

## Related issue
[](https://github.com/Kwenta/kwenta/issues/441)

## Motivation and Context
While a user is trading on the market's page, they do not want to have to switch pages or have multiple windows opens to keep track of their other positions. We want them to have everything they need while waiting to trade futures on the Markets page.

## How Has This Been Tested?
Using brave browser, metamask, on optimism mainnet, and 0.1x leverage, I tested this out using an eth perp. I ended up at the desired result. 

Screen Shot 2022-04-03 at 9 49 34 PM

### DIFF
--- a/sections/futures/MarketInfo/MarketInfo.tsx
+++ b/sections/futures/MarketInfo/MarketInfo.tsx
@@ -5,6 +5,8 @@ import styled from 'styled-components';
 import useSynthetixQueries from '@synthetixio/queries';
 
 import useSelectedPriceCurrency from 'hooks/useSelectedPriceCurrency';
+import FuturesPositionsTable from 'sections/dashboard/FuturesPositionsTable';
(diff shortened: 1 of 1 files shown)


brossetti1 (approved):
this looks good 💯

PR #598, opened 2022-04-04
Avg Entry Price

Tburm:
Implement the average entry price from the subgraph entity

## Description
* Change the title on the position card
* Retrieve the average entry price from the subgraph for `FuturesPositionsTable`

## Related issue
Closes #534 

image

image


platschi:
Situation: I have an open position on ETH. From dashboard, I visit LINK-PERP. In the Futures Positions overview at the bottom, the Avg Entry price is loaded with a dash and only shows up after about a minute of delay. 

Screen Shot 2022-04-04 at 17 31 29

Tburm:
@platschi It's expected that the data will arrive separate from other table data, but the delay is too long. I added a dependency that looks like it fixed this locally.

koredefashokun (approved):

PR #616, opened 2022-04-05
Use Infura endpoint as default when wallets are not connected

### LINKED ISSUE #470: Show markets data when not connected to wallet
### Blocker
Leifu Infura PR (https://github.com/Kwenta/kwenta/commit/2330c9e88370426dbe78e287f6c01c5a9dae9e48) on dev should be merged into v2

### Subject of the issue
Several community members requested to be able to read available Futures market data if not connected to a wallet.

> can we add a ticket to be able to play around with Kwenta futures without having a wallet connected?

> can we show the available futures markets data even if no wallet is connected?

> Please let me view exchange prices and other stuff without connecting my wallet.

### Deliverables
- Make v2 data visible if not connected to wallet

koredefashokun:
## Description
This PR uses Infura as a default/fallback when wallets are not connected, so we can display information on futures markets on the dashboard.

## Related issue
closes #470 

## Motivation and Context
This fixes an important UX issue on the v2 beta.

## How Has This Been Tested?
N/A

image

### DIFF
--- a/queries/futures/useGetFuturesMarkets.ts
+++ b/queries/futures/useGetFuturesMarkets.ts
@@ -67,7 +67,7 @@ const useGetFuturesMarkets = (options?: UseQueryOptions<FuturesMarket[]>) => {
 				);
 		},
 		{
-			enabled: isAppReady && isL2 && !!synthetixjs,
+			enabled: isAppReady && !!synthetixjs,
 			...options,
 		}
 	);
--- a/queries/futures/useGetFuturesMarkets.ts
+++ b/queries/futures/useGetFuturesMarkets.ts
@@ -3,7 +3,7 @@ import { useRecoilValue } from 'recoil';
 import { wei } from '@synthetixio/wei';
 
 import { appReadyState } from 'store/app';
-import { isL2State, networkState, walletAddressState } from 'store/wallet';
+import { isL2State, networkState } from 'store/wallet';
 
 import Connector from 'containers/Connector';
 import QUERY_KEYS from 'constants/queryKeys';
--- a/queries/futures/useGetFuturesMarkets.ts
+++ b/queries/futures/useGetFuturesMarkets.ts
@@ -3,7 +3,7 @@ import { useRecoilValue } from 'recoil';
 import { wei } from '@synthetixio/wei';
 
 import { appReadyState } from 'store/app';
-import { isL2State, networkState } from 'store/wallet';
+import { isL2State, isWalletConnectedState, networkState } from 'store/wallet';
 
 import Connector from 'containers/Connector';
 import QUERY_KEYS from 'constants/queryKeys';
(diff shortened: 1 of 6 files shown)


Tburm:
@koredefashokun Is it possible to get this to default to Optimism? Because this uses `window.ethereum` I see market for whichever network I have set in my wallet. If that's Ethereum Mainnet, it doesn't display any market.

koredefashokun:
@Tburm, do you mean that in the context of the "fallback" state? i.e. when no wallet is connected, we should automatically fetch the data from mainnet-ovm, or we change the default network to mainnet-ovm globally?

brossetti1:
looks like we would have to switch the `DEFAULT_NETWORK_ID` if you want to accomplish optimism as the default fallback

https://github.com/Kwenta/kwenta/blob/78ea76b1ff602e0305550f8eb3eff8a297e60b28/constants/defaults.ts#L19

koredefashokun:
Yeah. Saw that @brossetti1. However, in this case I only set the id to mainnet-ovm when we're using the Infura fallback. Chose to not change the `DEFAULT_NETWORK_ID` yet, because I wasn't sure what else depended on it. I think I'll go ahead and change it now though.

Tburm:
@koredefashokun I did mean the former. Even after changing the default, I'm not seeing it default to Optimism mainnet.

To test:
* Disconnect wallet
* Set wallet to Ethereum Mainnet in metamask
* Refresh page

If you do this, there are no markets displayed.

Tburm (approved):
looks good 👍🏼

queries/futures/useGetFuturesMarkets.ts
```diff
@@ -3,26 +3,37 @@ import { useRecoilValue } from 'recoil';
...
+	const provider = loadProvider({
+		networkId: 10,
+		infuraId: process.env.NEXT_PUBLIC_INFURA_PROJECT_ID,
+	});
+
+	const snxjs = synthetix({ provider, networkId: 10, useOvm: true });
+
+	const isWalletConnected = useRecoilValue(isWalletConnectedState);
+	const isL2 = useRecoilValue(isL2State);
+	const isReady = isAppReady && !!snxjs;
```
brossetti1, line 28: this is the main way to fix these queries - since the Futures markets are only supported on Optimism, the FuturesMarketData contract pull needs to happen from an optimism network - if you're connected to mainnet - this query wiill always throw an error regardless of where it is used because the mainnet synthetix.js has no reference to `FuturesMarketData` contracts. This is the error that gets thrown when using the Connector synthetix provider:

Screen Shot 2022-04-07 at 3 05 10 PM

queries/futures/useGetFuturesMarkets.ts
```diff
@@ -3,22 +3,28 @@ import { useRecoilValue } from 'recoil';
...
+	const isL2 = useRecoilValue(isL2State);
+	const isReady = isAppReady && !!synthetixjs;
+
 	return useQuery<FuturesMarket[]>(
 		QUERY_KEYS.Futures.Markets(network.id),
 		async () => {
+			if (isWalletConnected && !isL2) {
+				return null;
+			}
+
```
brossetti1, line 27: short circuit if connected and not on l2 so query does not fail because contracts do not exist

utils/network.ts
```diff
@@ -15,9 +15,9 @@ export function isSupportedNetworkId(id: number | string): id is NetworkId {
 	return id in NetworkNameById;
 }
 
-export async function getDefaultNetworkId(): Promise<NetworkId> {
+export async function getDefaultNetworkId(walletConnected: boolean = true): Promise<NetworkId> {
 	try {
-		if (window.ethereum) {
+		if (walletConnected && window.ethereum) {
```
brossetti1, line 20: only use specified network if user is connected -  this will [default to optimism](https://github.com/Kwenta/kwenta/pull/616/files#diff-1d57bb0118fc478a61a04d99d5b0d2405b1d01576521d8acb319ae2ff500243eR19) when not connected

queries/futures/useGetFuturesMarkets.ts
brossetti1: discussed this with Troy, Ralf, and korede - we worked out a better solution on the futures queries

platschi (approved):
Tested deployed version and works fine when not connected to wallet. Switched several networks, all good, no errors encountered. Works fine when connecting and disconnecting again, etc.

brossetti1 (approved):

Issue #595, opened 2022-04-04
Verify functionality of new markets


platschi:
New Future markets have been released to `ovm-mainnet` but are currently **paused**: SOL, AVAX, MATIC, AAVE, UNI, EUR, GOLD, SILVER. We need to be ready for the moment they are activated.

### Current issues

- [x] On Op Kovan, deposit and tx succeeds for new markets but no balance is shown.

- [x] **Dashboard**: Futures Markets currently does not list paused markets, make sure all 8 new available markets are listed correctly once going live. Currently on Op Kovan, only 6 out of 11 markets are listed/shown.

- [x] **Markets**: Markets dropdown only shows 5 markets currently on OP Kovan, while there are 11 markets available. Make sure it shows all available markets.

PR #620, opened 2022-04-05
Get correct market key from futures-market config

koredefashokun:
## Description
This PR fixes an issue where the data on the futures market page was not displaying correctly, because the market key for new assets were not being parsed correctly.

## Related issue
- #595 

## Motivation and Context
This fixes issues with trading in the new markets (currently paused on mainnet-ovm)

## How Has This Been Tested?
N/A

image

### DIFF
--- a/utils/futures.ts
+++ b/utils/futures.ts
@@ -1,0 +1,18 @@
+import { NetworkId, NetworkIdByName } from '@synthetixio/contracts-interface';
+
+import futuresMarketsKovan from 'synthetix/publish/deployed/kovan-ovm/futures-markets.json';
+import futuresMarketsMainnet from 'synthetix/publish/deployed/mainnet-ovm/futures-markets.json';
+
+export const getMarketKey = (asset: string | null, networkId: NetworkId) => {
+	if (networkId === NetworkIdByName['mainnet-ovm']) {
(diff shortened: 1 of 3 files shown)


utils/futures.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const getMarketKey = (asset: string | null, networkId: NetworkId) => {
+	if (networkId === NetworkIdByName['mainnet-ovm']) {
+		return futuresMarketsMainnet.find((market) => market.asset === asset)?.marketKey || 'sETH';
+	} else if (networkId === NetworkIdByName['kovan-ovm']) {
+		return futuresMarketsKovan.find((market) => market.asset === asset)?.marketKey || 'sETH';
+	} else {
+		return 'sETH';
+	}
+};
```
koredefashokun, line 14: I chose to default to 'sETH', since thats the new default market. However, we never actually call this function with a falsy input, so it's basically unreachable code put there to satisfy TS.

JChiaramonte7 (commented):
LGTM but letting @brossetti1 review as well so that he's across this

utils/futures.ts
brossetti1: this sounds reasonable to me, I can wait for this to merge and rebase to incorporate this into the open prs I have dealing with asset vs marketKey

brossetti1 (approved):

Issue #617, opened 2022-04-05
Disable Inputs when no wallet connected


platschi:
### Subject of the issue
Ticket #476 introduces to disable the order amount input and withdrawal button when the margin is 0$. On the same line, we should also disable the amount input fields when no wallet is connected. 

### Relevant because:
Once PR #616 will get merged, it will be possible for website visitors to view all Futures Markets data without being connected to a wallet.

### Similar issues
- #476 (#531)

### Deliverables

- [ ] Make sure inputs are disabled when no wallet is connected.

### Reward
25 SNX

platschi:
Unfortunately just got overruled and we want to keep inputs enabled if not connected for users to "play around".

PR #646, opened 2022-04-08
Fix disable input no wallet connection

lohba:
## Description

Disable the order amount input and withdrawal button when wallet is not connected 
## Related issue

Resolves: #617

### DIFF
--- a/sections/futures/Trade/Trade.tsx
+++ b/sections/futures/Trade/Trade.tsx
@@ -251,7 +251,10 @@ const Trade: React.FC<TradeProps> = () => {
 					{t('futures.market.trade.button.deposit')}
 				</MarketActionButton>
 				<MarketActionButton
-					disabled={!futuresMarketsPosition?.remainingMargin || futuresMarketsPosition.remainingMargin.lte(zeroBN)}
+					disabled={
+						!futuresMarketsPosition?.remainingMargin ||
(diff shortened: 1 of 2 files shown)


platschi:
Hey @lohba, almost good to go this ticket, just one last thing, can you remove the `package-lock.json` file from the commit?

platschi (changes requested):
Hey @lohba, almost good to go this ticket, just one last thing, can you remove the package-lock.json file from the commit?

platschi (approved):

sections/futures/Trade/Trade.tsx
```diff
@@ -255,7 +255,10 @@ const Trade: React.FC<TradeProps> = () => {
 					{t('futures.market.trade.button.deposit')}
 				</MarketActionButton>
 				<MarketActionButton
-					disabled={futuresMarketsPosition?.remainingMargin?.lte(zeroBN)}
+					disabled={
+						!futuresMarketsPosition?.remainingMargin ||
```
LeifuChen, line 259: The code will be easier to understand if we use `isWalletConnected` to check the status of the wallet connection. https://github.com/Kwenta/kwenta/blob/v2-dev/store/wallet/index.ts#L35-L37

LeifuChen (changes requested):
The solution is working well. Just request a minor change to make it more understandable for someone else. Thanks.

PR #652, opened 2022-04-08
Funding Fallback

Tburm:
Add a fallback to the instantaneous funding rate when average funding is not available.

## Description
This change will ensure that we always have a value in the "Funding Rate" area of Market Details

## Related issue
None

## Motivation and Context
When we are missing recent funding events, this metric will fail to calculate and will be blank.

### DIFF
--- a/sections/futures/MarketDetails/MarketDetails.tsx
+++ b/sections/futures/MarketDetails/MarketDetails.tsx
@@ -66,8 +66,13 @@ const MarketDetails: React.FC<MarketDetailsProps> = ({ baseCurrencyKey }) => {
 	const pastPrice = dailyPriceChanges.find((price: Price) => price.synth === baseCurrencyKey);
 
 	const data: MarketData = React.useMemo(() => {
-		const fundingTitle = `${avgFundingRate ? '24H' : 'Inst.'} Funding Rate`;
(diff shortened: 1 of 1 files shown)


sections/futures/MarketDetails/MarketDetails.tsx
```diff
@@ -66,6 +66,14 @@ const MarketDetails: React.FC<MarketDetailsProps> = ({ baseCurrencyKey }) => {
 	const pastPrice = dailyPriceChanges.find((price: Price) => price.synth === baseCurrencyKey);
 
 	const data: MarketData = React.useMemo(() => {
+		const fundingTitle = `${
+			fundingRateQuery.failureCount > 0 && !avgFundingRate && !!marketSummary ? 'Inst.' : '24H'
```
koredefashokun, line 70: We might have to come back to this for i18n purposes (seems like a pain here though lol)

koredefashokun (commented):
Just left a small comment.

sections/futures/MarketDetails/MarketDetails.tsx
Tburm: agreed, I think we could just pull translations here instead but we're also talking about changing to 1 hr. Resolving this for now, can revisit when we make those changes.

Issue #452, opened 2022-03-18
Swap UI Implementation


platschi:
Implementing the new v2 Swap UI design into Kwenta. The Swap UI consists of two elements: The (basic) Synth Exchange and the Advanced Trading Mode.

### Purpose of the ticket
This ticket tracks the Swap UI implementation

### Deliverables

- [x] Implement new Synth Exchange design (basic)
- [x] Write unit test for new synth exchange

### Ticket Tracker

- #276

#### Basic Synth Exchange
- #459

asherism:
advanced swap:
https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=1846%3A9749
<img width="1097" alt="image" src="https://user-images.githubusercontent.com/11524787/159050705-bc473b7c-0508-463d-9662-0549b8d2577a.png">

simple swap:
https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=1885%3A16044
<img width="1103" alt="image" src="https://user-images.githubusercontent.com/11524787/159050765-4f6f1506-5609-4adc-b204-651b5caad0cb.png">

PR #654, opened 2022-04-08
Basic Swap V2

LeifuChen:
## Description
Implementing a new v2 Swap UI into Kwenta: (basic) Synth Exchange.

## Related issue
#452

## Motivation and Context
Migrate more functions from V1 to V2.

## How Has This Been Tested?
Swap different tokens via UI.

image


LeifuChen:
- [ ] 1) Navbar: Probably as soon as we merge Exchange, we should rename "Markets" to "Futures" ?
- [ ] 2a) Select Synths modal: Can we add a condition to hide the buttons for "Equities"/"Commodities" if there are 0 available?
- [ ] 2b) Select Synths modal: There are still a bunch of labels in  elderberry color, we probably want to change them to color: `${(props) => props.theme.colors.common.secondaryGray};` to be in line with the rest of the design
<p align="center">
  
</>

- [ ] 3) Just testing it I would have probably have liked it to have the "Select Synths" fields auto-filled on first visit with a default pair, e.g. sUSD-sETH, or maybe at least for "sUSD" to be set as first default maybe. But not sure that is desired, that's a question for product. Similar to Uniswap that shows default ETH and then second pair "Select token".
- [ ] 4) Same with color here, elderberry should be `secondaryGray`. Not sure Price Adjustment applies to L2, so maybe we can hide that info if network != Ethereum ?
<p align="center">
  
</p>

- [ ] 5) More general question: Is this the default name for the synth by Synthetix? I think this should be `Synthetic Crude Oil (WTI)"` or something like this, or am I technically buying an Oil PERP here?
- [ ] 6) The arrow in the middle of the Synth Exchange modal, I noticed I wanted to click it to swap the two selected synths I had. Is this functionality desired by the design? 
For the rest, exchanging synths works like a charm, it looks super clean and just feels totally smooth to use. Great work!
- [ ] 7) When I select sUSD, and type an input, e.g. 100 ; and then second step "Select Synth", all the inputs reset to 0. I would want that the 100 sUSD in the first input won't reset, this kind of forces me to go up - down - up in my exchange flow
- [ ] 8) PR https://github.com/Kwenta/kwenta/pull/655 removes this check on dev (v1) because dHedge integration does not seem to work using this. We need to check this for v2-dev to make sure dHedge integration works without issues 

<p align="center">
  
</p>

LeifuChen:
@platschi I have made the changes based on your feedback (see the above tracklist). Please verify the functionalities. Thanks.

platschi:
Works absolutely great now, exchange works as expected. A few nitpicks design things:

- [ ] Make sure long Synth names have some spacing between lines, see:

Screen Shot 2022-04-12 at 10 09 22

- [ ] Select Synths: Clicking a button, e.g. in below example "Commodities", works great, but as a user I have no way to know that I need to click one more time on the Commodities button to get back to the "All Synths" overview. I would suggest to implement a similar, subtle effect equal as for the "Long/Short" buttons on Futures Markets, e.g. using `transition: all 0.1s ease-in-out 0s` and `transform: scale(0.98)` or so. @asherism what do you think?

Screen Shot 2022-04-12 at 10 10 27

LeifuChen:
- [X] Adapt the long synth name in UI

<p align="center">

</p>

- [X] Add a new effect to show if the button is selected
<p align="center">

</p>

LeifuChen:
@platschi The following changes have been made.
1. Updated the selected effect of the `categoryButton`
2. Removed "Synth Fees" in tooltip
3. Showed "Max Fee" on L1 and removed the "EDIT" button
4. Enabled the input of "Into".

sections/shared/components/GasPriceSelect/GasPriceSelect.tsx
```diff
@@ -36,22 +31,24 @@ const GasPriceSelect: FC<GasPriceSelectProps> = ({ gasPrices, transactionFee, ..
 	const [gasSpeed, setGasSpeed] = useRecoilState<keyof GasPrices>(gasSpeedState);
```
Tburm, line 31: Can delete?

sections/shared/components/GasPriceSelect/GasPriceSelect.tsx
```diff
@@ -36,22 +31,24 @@ const GasPriceSelect: FC<GasPriceSelectProps> = ({ gasPrices, transactionFee, ..
...
 	const [customGasPrice, setCustomGasPrice] = useRecoilState(customGasPriceState);
```
Tburm, line 32: Can delete?
```suggestion
```

sections/shared/components/GasPriceSelect/GasPriceSelect.tsx
Tburm, line 31: Can delete?
```suggestion
```

Tburm (changes requested):
Balance is a little crowded with a large balance 🐳  :
image

When cancelling a transaction, we probably want more padding on this card/button:
image

I had a test transaction fail with "out of gas" on Optimism Kovan:
* [tx](https://kovan-optimistic.etherscan.io/tx/0x430a581e87ae0aae4376ddea2a8975ec5de44f372c670fa6e344d3aae2dc2164)
* [tenderly](https://dashboard.tenderly.co/tx/optimistic-kovan/0x430a581e87ae0aae4376ddea2a8975ec5de44f372c670fa6e344d3aae2dc2164)

After a transaction succeeds, it's really nice that balances are updated. It would also be good to clear the inputs to make it clear the transaction went through.

For assets with high value (sBTC) we should add more decimals:
image

sections/exchange/TradeCard/CurrencyCard/CurrencyCard.tsx
```diff
@@ -246,27 +322,44 @@ const Slippage = styled.div`
...
+	padding-left: 12px;
+`;
+
+const WalletBalanceContainer = styled(FlexDivRow)<{ disableInput?: boolean }>`
 	${(props) =>
 		props.disableInput &&
 		css`
 			pointer-events: none;
 		`}
+	width: 136px;
```
Tburm, line 342:
```suggestion
	width: 100%;
```

LeifuChen:
@Tburm  @platschi  The following issues have been fixed.
1. Update the precision of select currency modal
2. Increase the width of TxConfirmationModal
3. Fix the wallet balance display for whales
4. Remove the unnecessary set method
5. Reset the input fields after the confirmation of the trades
6. "Into" input fields only update when a synth selected

sections/exchange/TradeCard/CurrencyCard/CurrencyCard.tsx
LeifuChen: Increased the width of  `StyledBaseModal` to 300px.

sections/shared/components/GasPriceSelect/GasPriceSelect.tsx
LeifuChen: I removed the set method, which is not required.

sections/shared/components/GasPriceSelect/GasPriceSelect.tsx
LeifuChen: I removed the set method, which is not required.

sections/shared/components/GasPriceSelect/GasPriceSelect.tsx
LeifuChen: This is a duplicated comment :)

Tburm (dismissed):

Tburm (approved):

Tburm (approved):

Issue #405, opened 2022-03-15
Upgrade to web3-onboard


platschi:
### Subject of the issue
Kwenta currently uses `bnc-onboard`, an outdated BlockNative library to onboard users to web3. The use of an outdated package results in several issues, such as #343, #344 or #392. We should upgrade Kwenta to `web3-onboard`, a newer, modernized version of `bnc-onboard`. 

### Links
-  [bnc-onboard docs](https://docs.blocknative.com/onboard-older-version)
- [web3-onboard docs](https://docs.blocknative.com/onboard)
- [web3-onboard github repo](https://github.com/blocknative/web3-onboard)

### Deliverables
- Upgrade to `web3-onboard`
- Remove `bnc-onboard`
- Refactor code to support `web3-onboard` accordingly
- Support Networks: Ethereum, Optimism (make sure testnets also work: Kovan, Optimistic Kovan)
- Test functionality 
- Make sure all available wallet modules are integrated to enhance support/functionality of Kwenta

### Rewards
75 SNX

platschi:
@LeifuChen

platschi:
Obsolete, upgraded to rainbowkit eventually.

PR #659, opened 2022-04-10
Upgrade to web3-onboard

leovct:
Migrate from the old blocknative library `bnc-onboard` to `web3-onboard`.

## Description
Update the `Connector` component (`containers/Connector\config.ts` and `containers/Connector\Connector.tsx`). The migration procedure is not defined in the blocknative library so I tried to do my best to integrate with the new library while keeping all the features already present.

I also updated the way the option modal is displayed (`sections/shared/modals/WalletOptionsModal/WalletOptionsModal.tsx`) since we don't need to check if a wallet is of hardware type. The new web3-onboard library already supports this feature natively.

Note: at the moment, the website compiles successfully but when building a page, it returns an error. I thought it was related to the new web3-onboard library ([see this issue on their repository](https://github.com/blocknative/web3-onboard/issues/889)) but I can't reproduce the error on a fresh new next.js repository...

```
error - Error [ERR_REQUIRE_ESM]: require() of ES Module .../kwenta/node_modules/@web3-onboard/react/dist/index.js from .../kwenta/.next/server/pages/_app.js not supported.
Instead change the require of index.js in .../kwenta/.next/server/pages/_app.js to a dynamic import() which is available in all CommonJS modules.
```

## Related issue
#405  

## Motivation and Context
Migrate from an old library to a new one with more features, releases and bug fixes. The library is widely used in the web3 space and users are becoming familiar with the new web3-onboard layout.

## How Has This Been Tested?
Not yet.


leovct:
This PR is based on the old `dev` branch.\
I opened a new PR based on the new `v2-dev` branch here #678.

PR #660, opened 2022-04-11
ci: add e2e tests to pipelines

drptbl:
## Description
Added e2e tests to the pipelines. However they seem to be unstable and fail, they have to be verified and fixed.

Also Cypress dashboard integration is missing in PRs (https://dashboard.cypress.io/). It's already set up in CI but new account and project has to be created on dashboard and then the bot has to be added to organization.


platschi:
This was branched against the old v1 dev branch, we need to update this PR to work on the new v2 `dev` branch.

PR #661, opened 2022-04-11
fixing tooltip issue - no divs as children of p tags

brossetti1:
validateDOMNesting(...): <div> cannot appear as a descendant of <p>.

## Description
Noticed a small issue with the tooltips on the markets page - cannot render `div` under `p` tags.

solution is to render a span instead of a `p` tag

## Related issue
N/A

## Motivation and Context
tooltips were recently changed (Friday), I believe this issue could be a symptom of the changes

## How Has This Been Tested?
in development environment - no error after changes; no styling changes on market row

Screen Shot 2022-04-10 at 6 21 38 PM

### DIFF
--- a/components/Tooltip/TooltipStyles.tsx
+++ b/components/Tooltip/TooltipStyles.tsx
@@ -11,64 +11,62 @@ interface ToolTipStyleProps {
 }
 
 export const Tooltip = styled.div<ToolTipStyleProps>`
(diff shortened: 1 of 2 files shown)


platschi (approved):

Tburm (approved):
was wondering where this came from, good catch

Issue #664, opened 2022-04-11
Implement incremental refresh


platschi:
- [x] Implement incremental refresh

PR #666, opened 2022-04-11
Add incremental data refresh

Tburm:
Add incremental data refresh to react queries

## Description
Adding data refresh to:
* Oracle price data on market details
* External price on market details (oracle price is a dependency)
* Position card (unrealized pnl)

## Related issue
#664 

## Motivation and Context
Oracle prices do not update on the frontend when the contract receives updates, which can cause a disconnect between actual positions and what is shown on the page if a trader leaves it open for a long time (>15 minutes)

## How Has This Been Tested?
* Open this version of the app
* Open live version at `v2.beta.kwenta.io`
* Leave them open for 15+ minutes
* Check that price updates on new version

### DIFF
--- a/sections/futures/MarketDetails/MarketDetails.tsx
+++ b/sections/futures/MarketDetails/MarketDetails.tsx
@@ -30,7 +30,7 @@ type MarketData = Record<string, { value: string | JSX.Element; color?: string }
 const MarketDetails: React.FC<MarketDetailsProps> = ({ baseCurrencyKey }) => {
 	const { network } = Connector.useContainer();
 	const { useExchangeRatesQuery } = useSynthetixQueries();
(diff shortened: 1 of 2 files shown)


Tburm:
@platschi There is a more complicated issue here with the TV chart. We should have a separate issue to upgrade the TV chart to use `useQuery` so we can refetch that as well.

koredefashokun (approved):
Looks good to me. I think this might mean that we can get rid of some of the manual refetches we do in some places (like after orders are placed).

Tburm:
@koredefashokun I think we should leave the manual refetches. It would be a little clunky if that didn't update right away, and instead updated a minute later.

koredefashokun:
Fair enough, @Tburm . My bad. My brain gets twisted when computing milliseconds sometimes lol

Issue #628, opened 2022-04-06
Store last visited market in localstorage


platschi:
### Subject of the issue
The Markets navigation tab currently defaults to ETH-PERP (#614). Ideally, we would want to store the last visited market in _localstorage_, and when a user clicks on the "Markets" navigation tab, the website routes to the last visited market instead of the default market (ETH-PERP). If no last visited market is found in _localstorage_, the Markets tab should return to the default market.

### Similar ticket
- #614 

### Deliverables

- [x] Store last visited market in localstorage
- [x] Clicking Markets tab should fetch last visited market from localstorage and route to this market
- [ ] If no last visited market stored, it should default to ETH-PERP.

### Reward
40 SNX

PR #669, opened 2022-04-11
Store last visited market in localstorage

0xwontonsoup:
### Related issue
- #628


Tburm (approved):
Tested and it works as expected 👍🏼

Issue #557, opened 2022-03-29
Improve unsupported network wallet state


platschi:
### Subject of the issue
When a user switches to an unsupported network, Kwenta resolves in a white "Unexpected Error" page. This happens in the following situations:

-  When connected to Kwenta on Optimism L2 or Ethereum L1 (or testnets Kovan resp. Optimistic Kovan) and a user switches to another network (e.g. Goerli), the white screen appears
- When the wallet by default is set to Goerli network, and a user visits Kwenta, the pages does not even load but only shows the white error screen.

The error has been reported to appear also with other networks, e.g. Polygon, etc.

### Expected design
When a user is connected to an unsupported network*, the Kwenta Futures page should load correctly and show the following wallet state:

<img width="1627" alt="unsupported" src="https://user-images.githubusercontent.com/548702/160687763-0187069a-eafd-4187-9354-3d09f4acefcb.png">

### Deliverables

- [x] Make sure that when a user is connected to an unsupported network, the unsupported network wallet state is loaded
- [x] Show "Switch to L2" CTA button when on unsupported network to the left of the wallet state (not in design mockup, but available in code, currently visible when connected to L1)

Note: Currently supported networks are Ethereum Mainnet, Optimism Mainnet, Kovan and Optimistic Kovan. Everything else should be treated as "unsupported".

### Relevant tickets
Design ticket #427

### Reward
75 SNX

PR #680, opened 2022-04-12
Improve unsupported network state

leovct:
## Description
When a user switches to an unsupported network, Kwenta resolves in a white "Unexpected Error" page.\
Note: unsupported networks are any other networks than Ethereum, Optimism, Ethereum Kovan and Optimism Kovan.

This PR aims to solve this issue by displaying the unsupported wallet state when a user is connected to an unsupported network. This new state consists of the "Switch to L2" CTA button and an "Unsupported Network" button as shown below.

unsupported-wallet-state

I could not rely on the `isSupportedNetworkId(id)` function defined in `utils/network.ts` since it always returned that the network was supported, even for Goerly. That's why I decided to define a `SUPPORTED_NETWORKS` constant array in `constants/networks.ts`, containing all the chain ids of the supported networks. I then updated the `isSupportedNetworkId(id)` function to simply check if the id of the network is in the `SUPPORTED_NETWORKS` array or not.

Then I updated how the user menu behaves by checking if the network is supported in `sections/shared/Layout/AppLayout/Header/UserMenu/UserMenu.tsx`. If the answer is yes, the menu stays the same but if the answer is no, it displays the "Switch to L2" CTA button as well as the "Unsupported wallet" button. At the same time, I defined a new translation for "unsupported network".

I also updated `pages/_app.tsx` because there were no tests to verify if the network was supported. It was responsible for the page not loading when the user came, logged on to an unsupported network. So I added this necessary check and the website now loads no matter what chain the user is logged on to.

Finally, I updated `containers/Connector/Connector.tsx` because the network state was not always up to date. When the user was connected on an unsupported network, sometimes, the network state was not updated and still displayed the supported wallet state.

## Related issue
#557 

## Motivation and Context
This PR solves a bug that was degrading the user experience. The website now behaves properly no matter what chain the user is logged on to. It also makes it clear to the user (in terms of UI/UX) that they are or are not on the right network.

### Test 1
- Connect to the website while already being connected to a supported network with Metamak (i.e. Ethereum or Optimism) and check that the website displays the "Switch to L2" CTA button, the address (or ens name) and the dots button.
- Switch to another supported network, the color or the button updates but the rest of the user menu stays the same.
- Switch to L2, the website displays the "Optimism" button, the address (or ens name) and the dots button.
- Switch to an unsupported network (i.e. Goerli) and check that the website displays the "Switch to L2" CTA button and the "Unsupported Network" button.

### Test 2
- Connect to the website while already being connected to an unsupported network with Metamak (i.e. Goerli) and check that the website displays the "Switch to L2" CTA button and the "Unsupported Network" button.
- Switch to another unsupported network (i.e. Avalanche) and check that the website displays the "Switch to L2" CTA button and the "Unsupported Network" button.
- Switch to L2, the website displays the "Optimism" button, the address (or ens name) and the dots button.
- Switch back to another unsupported network and check again that the website displays the "Switch to L2" CTA button and the "Unsupported Network" button.
- Switch to a supported network (i.e. Ethereum) and check that the website displays the "Switch to L2" CTA button, the address (or ens name) and the dots button.

### DIFF
--- a/constants/network.ts
+++ b/constants/network.ts
@@ -5,3 +5,10 @@ export const GWEI_DECIMALS = 9;
 export const ETH_UNIT = 1000000000000000000;
 
 export type GasLimitEstimate = BigNumber | null;
+
+export const SUPPORTED_NETWORKS = [
+	1, // Ethereum (mainnet)
+	10, // Optimism (mainnet)
+	42, // Ethereum Kovan (testnet)
+	69, // Optimism Kovan (tesnet)
+];
--- a/containers/Connector/Connector.tsx
+++ b/containers/Connector/Connector.tsx
@@ -6,9 +6,9 @@ import {
 } from '@synthetixio/transaction-notifier';
 import { loadProvider } from '@synthetixio/providers';
 
-import { getDefaultNetworkId, getIsOVM } from 'utils/network';
+import { getDefaultNetworkId, getIsOVM, isSupportedNetworkId } from 'utils/network';
 import { useSetRecoilState, useRecoilState, useRecoilValue } from 'recoil';
-import { NetworkId, SynthetixJS, synthetix } from '@synthetixio/contracts-interface';
+import { NetworkId, SynthetixJS, synthetix, NetworkName } from '@synthetixio/contracts-interface';
 import { ethers } from 'ethers';
 
 import { ordersState } from 'store/orders';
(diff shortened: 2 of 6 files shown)


leovct:
Yep, it's a typo sorry! I fixed it :)

Tburm (changes requested):
I found myself in two states that were unexpected...

** Clicking Unsupported Network Button **

1. Connect to a supported network
2. Switch to unsupported network
3. Click "Unsupported Network" button

In this case, it brings up the wallet connection modal. I'm already connected, so I could expect the "switch" and "disconnect' modal, or for the button to do nothing.

** Disconnecting wallet **

1. Connect to a supported network
2. Click wallet button -> disconnect wallet

At this point the button turns into the "Unsupported Network" button. I would expect this to offer the "Connect Wallet" button as if you are disconnected.

leovct:
Thanks for the review! I fixed both issues. Good catch on the disconnect wallet issue, I missed it. Now, the "Unsupported Network" button does nothing. I tried to make it display the "Switch and Disconnect" modal but the styling was not the same and I didn't want to mess up the WalletActions component.

Tburm (approved):
Approving the changes, but @platschi maybe we should think about what the actions of the "Unsupported Network" button should be?

PR #687, opened 2022-04-13
Profit Calculator

0xwontonsoup:
* Mostly clean
* Built everything mostly from scratch (refactoring rn to reuse the [`CustomInput`](https://github.com/0xwontonsoup/kwenta/blob/feat/profit-calculator/components/Input/CustomInput.tsx) component :) 
* Removing unused stuff

## Update
* Fixed underflow/overflow errors (bug on my part, mistake when using `.toFixed()` 😀 
* Reused `CustomInput` to reduce diff
* Fixed PnL calculations for when the `leverageSide` is set to `long` or `short`


0xwontonsoup:
Since im trying to work solely with `BigNumber` types (converting user input strings to `BigNumber` types) handling overflow/underflow errors is a PITA

0xwontonsoup:
For whoever ends up reviewing this, ive been able narrow down the underflow error i sometimes get on certain inputs:

1. for the `Position Size` input on the left, if you enter either `1.1` or `20.1`, you'll see this error
2. for the `Entry Price`, if you enter `20,000.1`, you'll see the error
3. for `Exit Price`, same as `Entry Price`, you'll see the same error

components/Button/TabButton.tsx
```diff
@@ -1,25 +1,32 @@
...
+const TabButton: React.FC<TabButtonProps> = ({ title, detail, badge, active, icon, ...props }) => {
 	return (
 		<StyledButton {...props}>
 			<div>
 				<p className="title">{title}</p>
 				{detail && <p className="detail">{detail}</p>}
 			</div>
 			{!!badge && <div className="badge">{badge}</div>}
+			{icon ? (
+				<img className="icon" src={`${calculatorIcon}`} height={'15px'} width={'auto'} />
```
Tburm, line 26: Hard-coding the calculator icon into our general component isn't a good solution. If we want an `icon` property we should be able to specify the icon, or you can just add this SVG icon in your specific implementation of the button.

components/Button/TabButton.tsx
```diff
@@ -62,6 +69,14 @@ const StyledButton = styled(Button)`
 		border-radius: 4px;
 	}
 
+	.icon {
+		margin-left: 6px;
+	}
+
+	.left-position {
+		
+	}
```
Tburm, line 78: Do we need this?

components/Input/CustomInput.tsx
```diff
@@ -29,10 +33,12 @@ const CustomInput: React.FC<CustomInputProps> = ({
 	return (
 		<CustomInputContainer style={style} className={className}>
 			<input
+				id={id}
+				defaultValue={defaultValue}
 				disabled={disabled}
 				placeholder={placeholder}
 				value={value}
-				type="number"
+				inputMode={id ? 'decimal' : undefined}
```
Tburm, line 41: This change breaks key functionality on the futures page. We are no longer preventing input of non-numeric characters in position sizes:
image

This also throws an error when you input some non-numeric character.

components/Input/CustomInput.tsx
```diff
@@ -58,8 +64,11 @@ const CustomInputContainer = styled.div`
 	padding: 12px 14px 12px 10px;
 
 	input {
-		display: flex;
-		flex: 1;
+		display: ${(props) => (props.className === 'profit-calc' ? '' : 'flex')};
+		flex: ${(props) => (props.className === 'profit-calc' ? '' : '1')};
+
+		width: ${(props) => (props.className === 'profit-calc' ? '100%' : '')};
```
Tburm, line 70: Avoid hard-coding implementation specific logic into components

sections/futures/ProfitCalculator/PnLs.tsx
```diff
@@ -0,0 +1,59 @@
...
+	let rateOfReturn: any = 0,
+		profit: any = ethers.BigNumber.from(0),
+		loss: any = ethers.BigNumber.from(0);
+
+	const labels = ['Exit PnL', 'Stop PnL', 'R:R'];
+
+	if (
+		parseFloat(props.entryPrice.toString()) !== 0 &&
+		parseFloat(props.exitPrice.toString()) !== 0 &&
+		parseFloat(props.stopLoss.toString()) !== 0
```
Tburm, line 24: If these are BigNumbers, use BigNumber functions. Something like `!props.entryPrice.eq(0)`

sections/futures/ProfitCalculator/PnLs.tsx
```diff
@@ -0,0 +1,59 @@
...
+	) {
+		console.log('eveerageSide: ', props.leverageSide);
```
Tburm, line 26: remove

sections/futures/ProfitCalculator/PnLs.tsx
```diff
@@ -0,0 +1,59 @@
+import { ethers } from 'ethers';
+import { BigNumber } from '@ethersproject/bignumber';
+
+import StatWithContainer from './StatWithContainer';
+import { PositionSide } from '../types';
+
+export const PnLs = (props: {
+	scalar: number;
```
Tburm, line 8: Can avoid passing this as a property

sections/futures/ProfitCalculator/PnLs.tsx
```diff
@@ -0,0 +1,59 @@
...
+	stopLoss: BigNumber;
+	exitPrice: BigNumber;
+	entryPrice: BigNumber;
+	amountInAsset: BigNumber;
+	leverageSide: PositionSide;
```
Tburm, line 13: You can pass properties directly, so we can avoid using `props.*`. Set the types separately and do something like:

```
PnLs = ({
    entryPrice,
    exitPrice,
    ...
}: YourType)
```

sections/futures/ProfitCalculator/PnLs.tsx
```diff
@@ -0,0 +1,59 @@
...
+			loss = props.entryPrice.sub(props.exitPrice).mul(props.amountInAsset).toNumber();
+		}
+
+		rateOfReturn = (profit / Math.abs(loss)).toFixed(2);
+	}
+
+	const returnStateVar = (index: number) => {
+		if (index === 0) return (profit / props.scalar).toString();
+		if (index === 1) return (loss / props.scalar).toString();
+		if (index === 2) return rateOfReturn;
```
Tburm, line 42: I don't understand why we are using an if statement here. Can't we store in an object with key/value pairs?

sections/futures/ProfitCalculator/ProfitCalculator.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+	const handleSetInput = (_e: any, _stateVar: any, _stateVarName: string) => {
+		let isNum, isFloat, isUglyFloat1, isUglyFloat2, clampDecimals;
+
+		_stateVar = _e.currentTarget.value;
+		isNum = /^\d+$/.test(_stateVar);
+		isFloat = /^[0-9]+\.[0-9]+$/.test(_stateVar);
+		isUglyFloat1 = /^\.[0-9]+$/.test(_stateVar);
+		isUglyFloat2 = /^[0-9]+\.$/.test(_stateVar);
+		clampDecimals = /^\d+\.\d{0,3}$/.test(_stateVar);
```
Tburm, line 37: All of this can be avoided if we just use a numeric input from the start. An input with `type="number"` ensures that users can't input things that aren't numeric, and removes the need for any of this checking.

sections/futures/ProfitCalculator/ProfitCalculator.tsx
```diff
@@ -0,0 +1,251 @@
...
+import styled from 'styled-components';
+
+import PnLs from './PnLs';
+import ProfitDetails from './ProfitDetails';
+import BaseModal from 'components/BaseModal';
+import LabelWithInput from './LabelWithInput';
+import PositionButtons from '../../../sections/futures/PositionButtons';
+import { PositionSide } from '../types';
+
+const scalar = 100;
```
Tburm, line 14: I don't believe this constant is necessary if we handle the math functions below in a more elegant way.

sections/futures/ProfitCalculator/ProfitCalculator.tsx
```diff
@@ -0,0 +1,251 @@
...
+		if (isNum || isFloat || isUglyFloat1 || isUglyFloat2 || clampDecimals) {
+			if (_stateVar.indexOf(' ') >= 0) {
+				// if includes whitespace
+				_stateVar.trim();
+			}
+
+			if (!isNaN(_stateVar) && _stateVar !== '') {
+				if (_stateVarName === 'entryPrice') {
+					const clampedInput = parseFloat(parseFloat(_stateVar).toFixed(2));
+					setEntryPrice(ethers.BigNumber.from(clampedInput * scalar));
```
Tburm, line 48: Sticking to `BigNumber` or `Wei` types throughout will avoid all of these `parseFloat` hacks

sections/futures/ProfitCalculator/ProfitCalculator.tsx
```diff
@@ -0,0 +1,251 @@
...
+						<LabelWithInput
+							defaultValue={marketAssetRate}
+							labelText={'Entry Price: '}
+							placeholder={'$43,938.11'}
+							onChange={(e: any) => handleSetInput(e, entryPrice, 'entryPrice')}
+						/>
+						<ProfitCalcGrid>
+							{/* LEFT column */}
+							<LeftColumn>
+								<LabelWithInput
```
Tburm, line 139: Probably want to use a numeric type like `NumericInput` where we already have a component. Maybe look at how we do this for order sizing.

sections/futures/ProfitCalculator/ProfitCalculator.tsx
```diff
@@ -0,0 +1,251 @@
...
+									id={'exit-price'}
+									labelText={'Exit Price: '}
+									placeholder={'$46,939.11'}
```
Tburm, line 142: I'd prefer placeholders that are derived from the input. So if the user has `ETH-PERP` selected, we take the `sETH` price and add 5% for exit price, decrease by 5% for stop loss.

sections/futures/ProfitCalculator/LabelWithInput.tsx
```diff
@@ -0,0 +1,44 @@
+import styled from 'styled-components';
+import CustomInput from '../../../components/Input/CustomInput';
```
Tburm, line 2: can just do `components/Input/CustomInput`

sections/futures/ProfitCalculator/ProfitDetails.tsx
```diff
@@ -0,0 +1,98 @@
...
+				{/* TAKE PROFIT */}
+				<RowName>
+					<RowText className="row-name">{'Take Profit:'}</RowText>
+				</RowName>
+				<Details style={{ justifySelf: 'right' }}>
+					<RowText>{`Sell`}</RowText>
+					<RowText style={{ marginRight: '10px', marginLeft: '10px' }} className="gray-font-color">
+						{`at`}
+					</RowText>
+					<RowText>{(props.exitPrice.toNumber() / props.scalar).toFixed(2)}</RowText>
```
Tburm, line 37: Scalar isn't a good solution here, you should try using our other standards for displaying number types where we can pass a BigNumber or Wei value.

sections/futures/ProfitCalculator/StatWithContainer.tsx
```diff
@@ -0,0 +1,38 @@
+import styled from 'styled-components';
+
+export const StatWithContainer = (props: { label: string; stateVar: any; type: number }) => {
+	return (
+		<>
+			<StatContainer>
+				<StatLabel>{props.label}</StatLabel>
+				{props.type === 0 ? <Stat style={{ color: '#7FD482' }}>{`$ ${props.stateVar}`}</Stat> : ''}
+				{props.type === 1 ? <Stat style={{ color: '#EF6868' }}>{`$ ${props.stateVar}`}</Stat> : ''}
+				{props.type === 2 ? <Stat style={{ color: '#ECE8E3' }}>{`${props.stateVar}x`}</Stat> : ''}
```
Tburm, line 10: We should avoid hard-coding these if statements.

sections/futures/ProfitCalculator/ProfitDetails.tsx
```diff
@@ -0,0 +1,98 @@
...
+				</Details>
+				{/* SIZE */}
+				<RowName>
+					<RowText className="row-name">{'Size:'}</RowText>
+				</RowName>
+				<Details style={{ justifySelf: 'right' }}>
+					<RowText style={{ marginRight: '10px' }}>
+						{(props.marketAssetPositionSize.toNumber() / props.scalar).toFixed(2)}
+					</RowText>
+					<RowText className="gray-font-color">{`${props.marketAsset}-PERP`}</RowText>
```
Tburm, line 58: This displays `sETH-PERP` where other parts of the app display `ETH-PERP`. You can find the code to format this elsewhere.

sections/futures/ProfitCalculator/ProfitCalculator.tsx
```diff
@@ -0,0 +1,251 @@
...
+import { ethers } from 'ethers';
+import { BigNumber } from '@ethersproject/bignumber';
+import { useTranslation } from 'react-i18next';
+import styled from 'styled-components';
+
+import PnLs from './PnLs';
+import ProfitDetails from './ProfitDetails';
+import BaseModal from 'components/BaseModal';
+import LabelWithInput from './LabelWithInput';
+import PositionButtons from '../../../sections/futures/PositionButtons';
```
Tburm, line 11: can just import from `sections/futures/...`

Tburm (changes requested):
I'm requesting some pretty significant changes. There is a lot of hard-coding in if statements that is unnecessary, and changes to components that breaks other parts of the app.

On BigNumbers, our standard is to use a Synthetix library called `Wei`. Instead of directly using `ethers.BigNumber.from(val)` you can use this for convenience:
```
// you:
stopLoss: BigNumber;

// fixed:
stoploss: Wei;
```

```
// you:
setStopLoss(ethers.BigNumber.from(clampedInput * scalar))

// fixed:
setStopLoss(wei(clampedInput * scalar))
```

0xwontonsoup:
commit [3ea6026](https://github.com/Kwenta/kwenta/pull/687/commits/3ea60267b08834c440ffe6083094110929394b65) fixes:
* fixed `entryPrice` not getting initialized on initial page render
* corrected gain and loss calculations
* removed custom function and additional css in `CustomInput` component. Instead, added styles to Instance of `CustomInput` in `LabelWithInput` component's definition

0xwontonsoup:
one thing i ignored, @Tburm, but want to discuss is using `marketAsset` to display the current synth-pair: 
Screen Shot 2022-04-21 at 9 50 18 PM

In the figma design, it shows `sBTC-PERP` so i assumed this meant that all synths would be displayed as `s***-PERP`

Screen Shot 2022-04-21 at 9 54 31 PM

Is displaying `ETH-PERP` the correct display name?

0xwontonsoup:
@Tburm i also wanted to discuss this comment

 
 I noticed in a figma design that Marko shared with me that there would be a share button added soon so i figured adding the `icon` prop would be helpful there too. The benefit is that since the `icon` prop is optional we don't need to create a new component for either of the calculator share buttons

Tburm (approved):

Tburm (approved):

sections/futures/ProfitCalculator/ProfitCalculator.tsx
```diff
@@ -0,0 +1,225 @@
...
+	return (
+		<>
+			<BaseModal
+				onDismiss={() => setOpenProfitCalcModal(false)}
+				isOpen={true}
+				/**
+				 * @todo IDK how to make this comply with this project's style, e.g.
+				 *       `t('modals.confirm-transaction.title')`
+				 */
+				title={t('Profit Calculator')}
```
JChiaramonte7, line 101: See the `en.json` file. cc @Tburm

Tburm: @0xwontonsoup The file is located at `/translations/en.json`, you can see how we're pulling individual translations there. You can add new entries for the calculator in `futures`

sections/futures/ProfitCalculator/ProfitDetails.tsx
```diff
@@ -0,0 +1,112 @@
...
+	marketAssetPositionSize,
+}) => {
+	const entryOrderDetails = leverageSide === PositionSide.LONG ? 'Long' : 'Short';
+
+	return (
+		<>
+			<StyledProfitDetails>
+				{/* ENTRY ORDER */}
+				<RowName>
+					<RowText className="row-name">{'Entry Order:'}</RowText>
```
Tburm, line 36: Can you handle these translations similar to the comment Jeremy left?

sections/futures/ProfitCalculator/PnLs.tsx
```diff
@@ -0,0 +1,56 @@
...
+	entryPrice,
+	amountInAsset,
+	leverageSide,
+}) => {
+	let rateOfReturn: any = 0,
+		profit: any = 0,
+		loss: any = 0;
+
+	// Calculate values for each stat
+	if (entryPrice !== '' && exitPrice !== '' && stopLoss !== '' && amountInAsset !== '') {
```
Tburm, line 27: I'd prefer if this didn't check _every_ input for _every_ calculation and instead checks the necessary input for that calculation.

For example, if I have entry price, exit price, and size it should calculate "Exit PnL", but leave the "Stop PnL" blank.

image

components/Button/TabButton.tsx
```diff
@@ -1,25 +1,32 @@
 import React from 'react';
 import styled from 'styled-components';
 import Button from './Button';
+import calculatorIcon from '../../assets/svg/futures/calculator-icon.svg';
```
Tburm, line 4: can import from `assets/svg/...`

components/Button/TabButton.tsx
Tburm, line 26: I see what you're saying about wanting another icon in the future. The issue with this implementation is that we're hard-coding the calculator icon as a boolean instead of specifying the icon we would like to display.

Tburm (changes requested):

0xwontonsoup (commented):
lmk if there's any new changes needed @Tburm @JChiaramonte7

Tburm (approved):

platschi (changes requested):
When I select "Short", and then click into "Entry Price", Profit Calculator crashes witha white _unexpected error_ page

Tburm (approved):
Looks good, and works as expected. Thanks @platschi for catching a couple things after my first review 👍🏼 

@brossetti1 FYI these buttons may conflict with the buttons you made for the deposit/withdrawal history

PR #690, opened 2022-04-13
Next price

koredefashokun:
## Description
This PR adds support for creating next-price orders on futures markets.

Note:
This PR also includes some changes related to hoisting data fetching to a parent component, for a bunch of components on the `futures` page that share similar data requirements. While the changes were made to make sure that refetching can be done correctly, they also have the added benefit of removing multiple fetches for the same data, and making sure that data is consistent across multiple components that depend on it. However, there will be other changes in the near future to clean up data flow and state management, especially on the futures page.

## Related issue
N/A

## Motivation and Context
This PR is part of the Kwenta v2 beta.

## How Has This Been Tested?
N/A


Tburm (changes requested):
Requested some changes about how we're displaying fees

sections/futures/Trade/NextPrice.tsx
```diff
@@ -0,0 +1,93 @@
...
+	// 	if (volatility.includes('%')) {
+	// 		inputRef?.current?.setSelectionRange(volatility.length - 1, volatility.length - 1);
+	// 	}
+	// }, [volatility]);
+
+	return (
+		<NextPriceContainer>
+			<p className="next-price-description">
+				Next-Price orders are subject to volatility and execute at the very next on-chain price.{' '}
+				<a href="https://example.com" rel="noreferrer" target="_blank">
```
platschi, line 21: Instead of hardcoding the URL, can you add it to `constants/links.ts` and point to it from here so we can update it on-the-fly once done?

koredefashokun: Done @platschi

sections/futures/Trade/NextPrice.tsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import styled from 'styled-components';
+import { EXTERNAL_LINKS } from 'constants/links';
+
+// import Button from 'components/Button';
```
Tburm, line 5: Can remove a lot of comments here?

Tburm (commented):
looks good other than the requested order self-execution changes

sections/futures/Trade/NextPrice.tsx
koredefashokun: @Tburm Yeah. Left those in because we will probably still need them when we add volatility control when placing next-price orders.

sections/futures/Trade/NextPriceConfirmationModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+	const transactionFee = useMemo(
+		() => newGetTransactionPrice(gasPrice, gasLimit, ethPriceRate, l1Fee),
+		[gasPrice, gasLimit, ethPriceRate, l1Fee]
+	);
+
+	const orderDetails = useMemo(() => {
+		const newSize = side === PositionSide.LONG ? tradeSize : -tradeSize;
+
+		return { size: (positionSize ?? zeroBN).add(newSize) };
```
Tburm, line 83: This logic isn't correct. If I have a short position open and I'm increasing the short, it subtracts `newSize` from `positionSize`, but `positionSize` has already applied `.abs()`

Example: I'm short 1 ETH. I submit an order to increase my short by .5 ETH. The modal will display `Size: .5 ETH`

sections/futures/Trade/NextPriceConfirmationModal.tsx
```diff
@@ -80,7 +80,7 @@ const NextPriceConfirmationModal: FC<NextPriceConfirmationModalProps> = ({
 	const orderDetails = useMemo(() => {
 		const newSize = side === PositionSide.LONG ? tradeSize : -tradeSize;
 
-		return { size: (positionSize ?? zeroBN).add(newSize) };
+		return { size: (positionSize ?? zeroBN).add(newSize).abs() };
```
Tburm, line 83: Won't this still be incorrect if someone is short and they are increasing their short? I think we need to adjust `positionSize` based on their position side as well.

koredefashokun: @Tburm I think position size is already either positive/negative depending on what side the position is on. If I recall correctly, that's actually how we get the side of the position on the frontend.

sections/futures/Trade/NextPrice.tsx
```diff
@@ -0,0 +1,29 @@
...
+import styled from 'styled-components';
+import { EXTERNAL_LINKS } from 'constants/links';
+
+const NextPrice: React.FC = () => (
+	<NextPriceContainer>
+		<p className="next-price-description">
+			Next-Price orders are subject to volatility and execute at the very next on-chain price.{' '}
+			<a href={EXTERNAL_LINKS.Trade.NextPriceBlogPost} rel="noreferrer" target="_blank">
+				Learn more ↗
+			</a>
```
Tburm, line 11: Should add this to translations

sections/futures/Trade/Trade.tsx
```diff
@@ -255,10 +255,10 @@ const Trade: React.FC = () => {
 				txHash: orderTxn.hash,
 				onTxConfirmed: () => {
 					onLeverageChange('');
-					setTimeout(() => {
-						futuresMarketPositionQuery.refetch();
+					setTimeout(async () => {
 						futuresPositionHistoryQuery.refetch();
 						marketQuery.refetch();
+						refetch();
```
Tburm, line 261: Would it make sense to refetch "contract" things immediately, and to refetch the "subgraph" things after some delay?

Tburm (approved):
A couple minor comments, but nothing blocking the release. Great work @koredefashokun !

platschi (approved):

Issue #376, opened 2022-02-25
Implement Leaderboard Designs


Tburm:
Implement leaderboard designs as found [here](https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=862%3A3578)
Update April 1st, 2022:

### Tickets need to be solved for MVP implementation
- #565 
- #566

### Related issues for leaderboard
- #567 
- #569 
- #570 

### Deliverables
Deliverables:
- [x] Change leaderboard/statistics tabs to buttons
- [x] Fix search bar styling
- [x] Edit formatting of Total PnL column
- [x] Add trader selection

These when above related issues are done:

- [x] Implement leaderboard pagination design

Issue #566, opened 2022-03-30
Move Statistics


platschi:
- [x] Move Statistics out of Leaderboard and into its own page / Navigation element

PR #691, opened 2022-04-13
Leaderboard Trader Selection

Tburm:
Add a trader selection to the leaderboard tab.

NOTE: This is a draft until the subgraph syncs

## Description
* Add trader selection
* Add position data to the selection

## Related issue
#376 

## How Has This Been Tested?
Tested positions on the subgraph and checked data between subgraph and UI

image


platschi:
- We can remove the Leaderboard and Statistics button, as Stats will later go into nav (- #566)

JChiaramonte7:
Two things:

1. Can we vertically align the address?
image

2. We should also push the address to the route when you click on a trader to view their history. Otherwise the back button goes to the dashboard.

LeifuChen:
@Tburm The `StyledSearchInput` is a little bit narrow with the placeholder (so it will be resized when we type the wallet address). Should we add more padding space there?
image

LeifuChen (approved):
Looks good to me.

PR #697, opened 2022-04-14
Data Bugs

Tburm:
Fixing a couple bugs on fetching data from subgraphs

## Description
* Fetch 24h volume before connected
* Fetch 24h price change before connected
* Fetch >100 trades


platschi (approved):
Works as expected

Issue #676, opened 2022-04-12
Fix Futures Position table


platschi:
### Subject of the issue
For higher values, numbers in the Position Table columns on the dashboard merge into each other, making the table unreadable. Additionally, parenthesis for the uPnL look weird (see screenshot).

![Screen Shot 2022-04-13 at 18 32 12](https://user-images.githubusercontent.com/548702/163274018-6d9e62ce-da07-4006-9f3d-6877b14cb2b6.png)

### Deliverables
- [ ] Use css stacking for the uPNL column (similar to how OI is solved in the Markets table)

### Reward
25 SNX

PR #702, opened 2022-04-15
update css on pnl container to flex column allowing stacked display

keybig:
## Description
This issue was simple to fix. The existing css for the element had `flex-direction: flex-row` which renders the items in one row. I changed this to `flex-direction: column` to get the stacking display. 

Next I removed the css class associated with the change in percent, as it was used to set a margin to give space between the items when displayed in one line. 

Next I removed `align-items: center` as all items in the table render the cells with content positioned to the left.

## Related issue
https://github.com/Kwenta/kwenta/issues/676

## Motivation and Context
For higher values, numbers in the Position Table columns on the dashboard merge into each other, making the table unreadable. Additionally, parenthesis for the uPnL look weird (see screenshot).

## How Has This Been Tested?
just visual tests to confirm everything looks good, and `npm run build` to make sure no errors came up. would appreciate any guidance for running specific tests before sending a commit. just need a point in the right direction :)

In the below screenshot, I manually put an extremely large number in the percent gain, as I could not simulate a trade showing huge gains. The point is to show large percent gains display properly
image

normal display with real trade numbers
image

### DIFF
--- a/sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
+++ b/sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
@@ -169,7 +169,7 @@ const FuturesPositionsTable: FC<FuturesPositionTableProps> = ({
 								<DefaultCell>{NO_VALUE}</DefaultCell>
 							) : (
 								<PnlContainer>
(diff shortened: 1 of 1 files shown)


platschi (approved):

LeifuChen (approved):
@keybig Great Job. The code looks good to me and thanks for your patience.

Issue #704, opened 2022-04-15
Leaderboard: Show ENS


platschi:
### Subject of the issue
The leaderboard currently only fetches wallet addresses of traders. To enhance the leaderboard, we could read ENS names of traders instead of their wallet addresses, if available.

### Deliverables

- [x] Show ENS name instead of wallet address in the leaderbord, if available
- [x] Show PFP of ENS, if available (screenshot 1)
- [x] Make sure clicking on an ENS name leads to the wallets trades history correctly (screenshot 2)

### Design Mockup
Note, the PFP can be the same size as the OP icon in the NetworkSwitcher modal

![leaderboard1](https://user-images.githubusercontent.com/548702/163568728-7822c75d-fba1-4f03-894a-6142effdd14a.png)

![Screen Shot 2022-04-15 at 09 06 55](https://user-images.githubusercontent.com/548702/163568895-55722909-b3f4-451a-a1d8-aa55e5c9b6fe.png)

### Reward
50 SNX

PR #710, opened 2022-04-17
feat: added ENS support on leaderboard page.

zhyd1997:
## Related issue

Resolves: #704 

## Motivation and Context

Reference: [Provider ENS methods](https://docs.ethers.io/v5/api/providers/provider/#Provider--ens-methods).

leaderboard

traderHistory

### DIFF
--- a/hooks/useENS.ts
+++ b/hooks/useENS.ts
@@ -1,0 +1,31 @@
+import { ethers } from 'ethers';
+import { useEffect, useState } from 'react';
+
+const useENS = (address?: string): { ensName: string | null; ensAvatar: string | null } => {
+	const [ensName, setENSName] = useState<string | null>(null);
+	const [ensAvatar, setENSAvatar] = useState<string | null>(null);
+
+	useEffect(() => {
+		let mounted = true;
+
+		(async () => {
+			if (address && ethers.utils.isAddress(address)) {
+				const provider = ethers.providers.getDefaultProvider();
+				const name = await provider.lookupAddress(address);
+				if (name && mounted) {
+					const avatar = await provider.getAvatar(name);
+					setENSAvatar(avatar);
+					setENSName(name);
+				}
+			}
+		})();
+
+		return () => {
+			mounted = false;
+		};
+	}, [address]);
+
+	return { ensName, ensAvatar };
+};
+
+export default useENS;
--- a/hooks/useENS.ts
+++ b/hooks/useENS.ts
@@ -1,19 +1,19 @@
-import { ethers } from 'ethers';
+import Connector from 'containers/Connector';
 import { useEffect, useState } from 'react';
 
 const useENS = (address?: string): { ensName: string | null; ensAvatar: string | null } => {
 	const [ensName, setENSName] = useState<string | null>(null);
 	const [ensAvatar, setENSAvatar] = useState<string | null>(null);
+	const { staticMainnetProvider } = Connector.useContainer();
 
 	useEffect(() => {
 		let mounted = true;
 
 		(async () => {
-			if (address && ethers.utils.isAddress(address)) {
-				const provider = ethers.providers.getDefaultProvider();
-				const name = await provider.lookupAddress(address);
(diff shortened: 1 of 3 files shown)


zhyd1997:
**Note**: 

the `#3` trader `kai.eth` has no ENS avatar:

traderNoENSAvatar

so I used another eth domain (`guoyu.eth`) to demonstrate that the PR changes can retrieve the ENS avatar correctly.

sections/leaderboard/Leaderboard/Leaderboard.tsx
```diff
@@ -166,13 +170,35 @@ const Leaderboard: FC<LeaderboardProps> = ({ compact }: LeaderboardProps) => {
...
+													{compact && cellProps.row.original.rank + '. '}
+													<StyledTrader>
+														{ensName ? (
+															<>
+																{ensAvatar && (
+																	<img
+																		src={ensAvatar}
+																		alt={ensName}
+																		width={16}
+																		height={16}
```
zhyd1997, line 188: The design token is a multiple of `8`.

and the size of the OP icon in the NetworkSwitcher modal is `17` px.

So I set the ENS avatar size to `16` px (8 x 2).

zhyd1997, line 188: The design token is a multiple of `8`.

and the size of the OP icon in the NetworkSwitcher modal is `17` px.

So I set the ENS avatar size to `16` px (8 x 2).

zhyd1997 (commented):
The design token is a multiple of `8`.

and the size of the OP icon in the NetworkSwitcher modal is `17` px.

So I set the ENS avatar size to `16` px (8 x 2).

platschi (approved):
Works as expected, showing ENS names incl. avatar (first available avatar loading for an ENS name on page 6), clicking trades shows ens name instead of address.

👍

hooks/useENS.ts
```diff
@@ -0,0 +1,31 @@
...
+const useENS = (address?: string): { ensName: string | null; ensAvatar: string | null } => {
+	const [ensName, setENSName] = useState<string | null>(null);
+	const [ensAvatar, setENSAvatar] = useState<string | null>(null);
+
+	useEffect(() => {
+		let mounted = true;
+
+		(async () => {
+			if (address && ethers.utils.isAddress(address)) {
+				const provider = ethers.providers.getDefaultProvider();
```
Tburm, line 13: This will create a new provider for every single call to the function, and does not inherit the provider that already exists. I don't think this solution is viable at the current size of our leaderboard, which will continue to grow.

You should get the signer using the method, only once: [ENS for wallet](https://github.com/Kwenta/kwenta/blob/a2f09c957d38c7fa9fec0d9643c963f1b79563e2/sections/shared/Layout/AppLayout/Header/WalletActions.tsx#L37)

hooks/useENS.ts
```diff
@@ -0,0 +1,31 @@
+import { ethers } from 'ethers';
+import { useEffect, useState } from 'react';
+
+const useENS = (address?: string): { ensName: string | null; ensAvatar: string | null } => {
```
Tburm, line 4: I'd love to see this take a list of addresses instead of a single address. This way, we could call the function any time our list of addresses in the leaderboard changes, and return an ENS object with the name and avatar for each address. It would reduce the number of calls to the hook by hundreds with the same result.

Tburm (changes requested):

hooks/useENS.ts
Tburm: I'm open to that if we can be sure that we are only making the ENS call, and not doing additional stuff like creating a new provider on every call.

zhyd1997: I will try to use a signer you mentioned.

zhyd1997: I just found that there is a `staticMainnetProvider` that can be used.

Tburm (approved):

Issue #707, opened 2022-04-16
Avg Entry Price not displayed


platschi:
### Subject of the issue
Community members reported that after opening or modifying a new position, the `Avg. Entry Price` and `Fees` in the `PositionCard` do not update and stay at $0.00.

The `FuturesMarketsTable` on the dashboard does not display the Avg Entry Price as well. See screenshots.

![perp](https://user-images.githubusercontent.com/548702/163676006-04297bc8-838c-4eb5-a3f5-fd816b711f95.png)

<img width="979" alt="Screen Shot 2022-04-16 at 10 06 03" src="https://user-images.githubusercontent.com/548702/163676121-bce8d338-7762-4e96-975e-c5441b22984f.png">

PR #713, opened 2022-04-18
Add fields to queries

Tburm:
Hotfix to correct missing funding and average pricing data on the position card.

Our utility function to clean data expects more fields that we are pulling in the query, so it's failing and not returning data to the Position Card.

Fixed:
image

#707


LeifuChen (approved):
Nice job! The change works well.

PR #721, opened 2022-04-19
Default wallet behavior

Tburm:
Fixing the default wallet behavior on connection. Right now we default to "loading..." while the ENS hook is running. This change will default to the wallet address, and only update if an ENS name exists for that address.

It should be better for addresses without an ENS name, as they never see "loading...".

It should be better for addresses with an ENS name, as they see their wallet address first before their ENS name updates which is common on other frontends.

### DIFF
--- a/sections/shared/Layout/AppLayout/Header/WalletActions.tsx
+++ b/sections/shared/Layout/AppLayout/Header/WalletActions.tsx
@@ -100,7 +100,7 @@ export const WalletActions: FC = () => {
 
 	useEffect(() => {
 		if (signer) {
-			setWalletLabel('loading...');
+			setWalletLabel(truncatedWalletAddress!);
 			signer.getAddress().then((account: string) => {
 				const _account = account;
 				getENSName(_account, staticMainnetProvider).then((_ensName: string) => {


platschi (approved):
Works good

sections/shared/Layout/AppLayout/Header/WalletActions.tsx
```diff
@@ -100,7 +100,7 @@ export const WalletActions: FC = () => {
 
 	useEffect(() => {
 		if (signer) {
-			setWalletLabel('loading...');
+			setWalletLabel(truncatedWalletAddress!);
 			signer.getAddress().then((account: string) => {
 				const _account = account;
 				getENSName(_account, staticMainnetProvider).then((_ensName: string) => {
```
brossetti1, line 106: how do we feel about something like this?

```
useEffect(() => {
	if (signer) {
		setWalletLabel('');
		signer
			.getAddress()
			.then((account: string) => {
				const _account = account;
				getENSName(_account, staticMainnetProvider).then((_ensName: string) => {
					setEns(_ensName);
					setWalletLabel(_ensName || truncatedWalletAddress!);
				});
			})
			.catch((_) => setWalletLabel(truncatedWalletAddress!));
	}
}, [signer, truncatedWalletAddress]);
```

brossetti1: yeah that makes sense - ok sounds good!

brossetti1 (approved):

sections/shared/Layout/AppLayout/Header/WalletActions.tsx
Tburm: AAVE is a good example of this ^^

Issue #639, opened 2022-04-07
Implement trade history panel


platschi:
Implement trade history & skew panel outlined in #629 

- Add websocket stream

### Blocking issues
- #629

PR #729, opened 2022-04-20
Add skew and trading history

LeifuChen:
## Description
Implement a trade history & skew panel

## Related issue
#639 

## Motivation and Context
Show the trade history and skew panel.

## How Has This Been Tested?
1. Switch to the different markets and check the skew and trade history
3. Check the layout when there is no trade

image
image
image


queries/futures/types.ts
```diff
@@ -163,9 +163,11 @@ export type FuturesTradeWithPrice = {
 };
 
 export type FuturesTrade = {
-	size: string;
+	size: Wei;
 	asset: string;
-	price?: string;
+	price?: Wei;
+	trxId: string;
```
Tburm, line 169: Is `trxId` correct? Other places we use `txnHash`

queries/futures/useGetFuturesTrades.ts
```diff
@@ -0,0 +1,58 @@
...
+						asset: true,
+					}
+				);
+				return response ? mapTrades(response) : null;
+			} catch (e) {
+				console.log(e);
+				return null;
+			}
+		},
+		{ enabled: isAppReady && isL2 && !!walletAddress, ...options }
```
Tburm, line 54: `!!walletAddress` wll prevent this running before they are connected

queries/futures/useGetFuturesTrades.ts
```diff
@@ -0,0 +1,58 @@
...
+import { getFuturesTrades } from './subgraph';
+import { DEFAULT_NUMBER_OF_TRADES } from 'constants/defaults';
+
+const useGetFuturesTrades = (
+	currencyKey: string | undefined,
+	options?: UseQueryOptions<FuturesTrade[] | null>
+) => {
+	const isAppReady = useRecoilValue(appReadyState);
+	const isL2 = useRecoilValue(isL2State);
+	const walletAddress = useRecoilValue(walletAddressState);
```
Tburm, line 20: can remove this

Tburm (commented):
Minor changes requested, but we should definitely populate this if a wallet isn't connected.

One other comment that I think Yashar had as well, the color of the scroll bar is kind of obnoxious. Can we make the background of the scrollbar transparent or something? :
image

queries/futures/useGetFuturesTrades.ts
LeifuChen: Yes. Thanks for your catch. I fixed this. There is no need to connect the wallet to show the trades history.

queries/futures/types.ts
LeifuChen: well, it's `id` in query so I just use `trxId`. Will change it to `trxHash`.

queries/futures/useGetFuturesTrades.ts
LeifuChen: Yes. Thanks.

LeifuChen:
@Tburm @platschi  The following changes have been made.
- [X] show data even the wallet is disconnected 
- [X] remove the placeholder row when no trades 
- [X] remove the extra padding space of the table row 
- [X] fix the shifted layout issue
- [X] rename the query type

Tburm (approved):

PR #736, opened 2022-04-21
Responsive layout

asherism:
leftSideContent, rightSideContent, mainContent are now just regular divs being positioned within a container with grid @ 20% 60% 20% percentages respectively


Tburm (commented):
It looks like this did a number on the layout for the exchange and leaderboard:
image

image

Table things also look pretty sketchy at smaller sizes:
image

Same with the market details, which I believe we could just stack into 2 rows for smaller screen sizes:
image

platschi (approved):

Issue #712, opened 2022-04-18
Paused Markets issues


platschi:
- Countdown for open TradFi markets should only show during 24h before closure, not it shows from Monday morning on counting down to Friday.
- Euro-PERP did not show countdown nor were marked as closed
- Closing times on Friday were set off by one hour. 

### Needs to be decided
- Skip Countdown and just show a badge message on Fridays that Markets are about to be closed

PR #742, opened 2022-04-22
Fix paused markets

brossetti1:
issue with paused markets - the wrong query is being used from the synthetix contracts

## Description
implemented a new query that specifically fetches the market and closure reason then built that into a hook similar to the `useMarketIsClosed` hook.

The images below only show that the new hook works
## Related issue
https://github.com/Kwenta/kwenta/issues/712

## Motivation and Context
present the paused badges when markets are closed (XAU, XAG, EUR).

## How Has This Been Tested?
- [X] - local development forcing hook and query to closed
 

Screen Shot 2022-04-25 at 1 21 18 PM
Screen Shot 2022-04-25 at 1 21 29 PM

### DIFF
--- a/constants/currency.ts
+++ b/constants/currency.ts
@@ -43,8 +43,8 @@ export const CRYPTO_CURRENCY = [
 ];
 
 export const CRYPTO_CURRENCY_MAP = keyBy(CRYPTO_CURRENCY);
-
-export const FIAT_SYNTHS: Set<CurrencyKey | 'EUR'> = new Set([
+export type FUTURES_FIAT = 'EUR' | 'JPY' | 'USD' | 'AUD' | 'GBP' | 'CHF';
+export const FIAT_SYNTHS: Set<CurrencyKey | FUTURES_FIAT> = new Set([
 	Synths.sEUR,
 	Synths.sJPY,
 	Synths.sUSD,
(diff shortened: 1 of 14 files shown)


constants/currency.ts
```diff
@@ -44,13 +44,14 @@ export const CRYPTO_CURRENCY = [
...
 
-export const FIAT_SYNTHS: Set<CurrencyKey> = new Set([
+export const FIAT_SYNTHS: Set<CurrencyKey | 'EUR'> = new Set([
 	Synths.sEUR,
 	Synths.sJPY,
 	Synths.sUSD,
 	Synths.sAUD,
 	Synths.sGBP,
 	Synths.sCHF,
+	'EUR',
```
brossetti1, line 54: not sure if we want to do something like this here to avoid additional fiats not registering as the release:

```
export type FUTURES_FIAT = 'EUR' | 'JPY' | 'USD' | 'AUD' | 'GBP' | 'CHF';
export const FIAT_SYNTHS: Set<CurrencyKey | FUTURES_FIAT> = new Set([
	Synths.sEUR,
	Synths.sJPY,
	Synths.sUSD,
	Synths.sAUD,
	Synths.sGBP,
	Synths.sCHF,
	'EUR',
	'JPY',
	'USD',
	'AUD',
	'GBP',
	'CHF',
]);
```

koredefashokun: I think this is a good idea. Also something for us to note when we start writing our own queries and types from scratch.

koredefashokun (approved):
Code changes look good.

koredefashokun (approved):

Issue #740, opened 2022-04-22
Futures Portfolio Value broken


platschi:
Futures Portfolio Value does not calculate total portfolio value, but shows $0.00 no matter how many open positions or deposited margin a user has.

PR #743, opened 2022-04-22
Portfolio Value

Tburm:
Fixing RPC error from portfolio value hook

## Description
Portfolio value was using incorrect market keys for the query. I cleaned up the RPC errors and the value populates on testnet.

## Related issue
- #740


koredefashokun (approved):
Changes look good

PR #776, opened 2022-04-29
Fix closed markets

brossetti1:
fixing paused markets, the issue was that the destructuring of the query for closed futures markets was still not returning legit values, but was returning false due to the way the response was manipulated.

## Description
there were still some issues when we tested the paused markets, KALUB from Synthetix was able to keep the market closed by giving `reasonCode = 3` (circuit-breaker) instead of `reasonCode = 2` (market-closed) - market-closed will reset after 5 minutes so reasonCode 3 allows us to test these markets.

there is a load delay on the markets page because we use this query in so many places, not sure the fix there yet but this is worth merging and creating an issue to fix it if its annoying.

## Related issue
N/A

## Motivation and Context
fix displaying paused markets

## How Has This Been Tested?
locally on WTI closed market kovan-optimism
locally on optimism mainnet

Screen Shot 2022-04-29 at 11 31 47 AM
Screen Shot 2022-04-29 at 11 31 55 AM

### DIFF
--- a/components/Badge/MarketBadge.tsx
+++ b/components/Badge/MarketBadge.tsx
@@ -43,15 +43,19 @@ export const MarketBadge: FC<MarketBadgeProps> = ({
 	const timerSetting = isOpen === null ? null : isOpen ? nextTransition : nextOpen;
 	const isMarketTransitioning = useIsMarketTransitioning(timerSetting ?? null);
 
+	if (typeof isFuturesMarketClosed !== 'boolean') {
+		return null;
+	}
+
+	if (isFuturesMarketClosed) {
+		return <Badge>{t(`futures.market.state.${futuresClosureReason}`)}</Badge>;
+	}
+
 	if (isMarketTransitioning && isOpen !== null) {
 		return <TransitionBadge isOpen={isOpen} />;
 	}
 
-	if (!isFuturesMarketClosed) {
-		return null;
-	}
-
-	return <Badge>{t(`futures.market.state.${futuresClosureReason}`)}</Badge>;
+	return null;
 };
 
 export default MarketBadge;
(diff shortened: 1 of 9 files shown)


components/Badge/MarketBadge.tsx
```diff
@@ -43,15 +43,19 @@ export const MarketBadge: FC<MarketBadgeProps> = ({
 	const timerSetting = isOpen === null ? null : isOpen ? nextTransition : nextOpen;
 	const isMarketTransitioning = useIsMarketTransitioning(timerSetting ?? null);
 
-	if (isMarketTransitioning && isOpen !== null) {
-		return <TransitionBadge isOpen={isOpen} />;
+	if (typeof isFuturesMarketClosed !== 'boolean') {
+		return null;
```
brossetti1, line 47: dont even show a badge unless we have a [legit value for if the market is closed or not.]()

hooks/useFuturesMarketClosed.ts
```diff
@@ -12,11 +11,11 @@ const useFuturesMarketClosed = (currencyKey: CurrencyKey | null) => {
 	const isFutureMarketSuspended =
 		futuresMarketSuspendedQuery.isSuccess && futuresMarketSuspendedQuery.data
 			? futuresMarketSuspendedQuery.data.isFuturesMarketClosed
-			: false;
+			: null;
```
brossetti1, line 14: returning `null` here so we know when its a legitamate closed or open market (true/false) or the underlying query is not done yet (null).

queries/futures/useFuturesSuspensionQuery.ts
```diff
@@ -37,14 +37,16 @@ const useFuturesSuspensionQuery = (
...
+				);
 
+				const reason = (isSuspended ? getReasonFromCode(reasonCode) : null) as FuturesClosureReason;
 				return {
-					isSuspended,
-					reasonCode: reason,
-					reason: (isSuspended ? getReasonFromCode(reason) : null) as FuturesClosureReason,
+					isFuturesMarketClosed: isSuspended,
+					reasonCode,
+					futuresClosureReason: reason,
```
brossetti1, line 49: destructuring of this query was incorrect for the `useGetFuturesMarketClosed` hook - which was resulting in `false` values all the time, instead of the correct vvalue when the market IS closed.

confirmed this on kovan testnet

Tburm (approved):

PR #786, opened 2022-05-01
Fix: profit calc

### LINKED ISSUE #785: Bug: profit calc entry price and stop loss are not calculated correctly on a short position
### Subject of the issue
When you click short and then enter a gain or loss %, the exit price and stop loss are calculated using the same formula as a long position.

### Expected behaviour
Entry price and stop loss should be calculated differently depending on whether the position is short or long

### Actual behaviour
In the calculator, if you have a short position and you enter a gain or loss %, such a 5, the exit price (if entering a gain %) or stop loss (if entering a loss %) will be calculated using the same formula as a long position

### Steps to reproduce
1. Go to https://v2.beta.kwenta.io/market/sETH 
2. Click on the `Calculator` button
3. Click the Short button
4. Enter 5 for a gain %
5. See that the exit price is calculated as `(1 + gainPercent) * entryPrice` instead of `(1 - gainPercent) * entryPrice`

0xwontonsoup:
## Description
Problem: 
* When you click short and then enter a gain or loss %, the exit price and stop loss are calculated using the same formula as a long position

## Related issue
Fixes #785 

## Motivation and Context
Entry price and stop loss should be calculated differently depending on whether the position is short or long

## How Has This Been Tested?
Math :D

### DIFF
--- a/components/Badge/MarketBadge.tsx
+++ b/components/Badge/MarketBadge.tsx
@@ -43,15 +43,19 @@ export const MarketBadge: FC<MarketBadgeProps> = ({
 	const timerSetting = isOpen === null ? null : isOpen ? nextTransition : nextOpen;
 	const isMarketTransitioning = useIsMarketTransitioning(timerSetting ?? null);
 
+	if (typeof isFuturesMarketClosed !== 'boolean') {
+		return null;
+	}
+
+	if (isFuturesMarketClosed) {
+		return <Badge>{t(`futures.market.state.${futuresClosureReason}`)}</Badge>;
+	}
+
 	if (isMarketTransitioning && isOpen !== null) {
 		return <TransitionBadge isOpen={isOpen} />;
 	}
 
-	if (!isFuturesMarketClosed) {
-		return null;
-	}
-
-	return <Badge>{t(`futures.market.state.${futuresClosureReason}`)}</Badge>;
+	return null;
 };
 
 export default MarketBadge;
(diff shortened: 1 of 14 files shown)


Tburm:
@0xwontonsoup It looks like this fixes the issue if you start by setting the position side. It doesn't work if I input everything, then change the leverage side.

To test:
* Input exit price, stop loss, position size
* Switch position side from long to short (nothing changes)
* Make a change to exit price (Gain % will update)
* Make a change to stop loss (Loss % will update)

I would expect gain/loss % to change when you change position side.

0xwontonsoup:
> @0xwontonsoup It looks like this fixes the issue if you start by setting the position side. It doesn't work if I input everything, then change the leverage side.
> 
> To test:
> 
> * Input exit price, stop loss, position size
> * Switch position side from long to short (nothing changes)
> * Make a change to exit price (Gain % will update)
> * Make a change to stop loss (Loss % will update)
> 
> I would expect gain/loss % to change when you change position side.

I think what you're pointing out is a separate issue and unrelated #785.
I'm aware of this issue but not sure how to use the latest state of `leverageSide` without creating a new state variable that won't be passed in as a prop to the `PositionsButton` component

Tburm:
@0xwontonsoup This change works as long as the expected behavior is that we hold the gain/loss % constant and calculate the exit/stop prices.

@platschi Is that our expected behavior?

platschi (approved):

Issue #739, opened 2022-04-22
Wallet: Show ENS PFP in wallet modal


platschi:
### Subject of the issue
Currently, ENS names are enabled in the wallet modal. When connected, we show a green dot. 

If the user has a ENS PFP, we should show the PFP instead of the green dot, similar to what we show on the leaderboard for ENS names with PFPs.

<img width="369" alt="Screen Shot 2022-04-25 at 20 57 18" src="https://user-images.githubusercontent.com/548702/165193121-61d89729-9d65-491c-b1db-52921cf8dd01.png">

### Deliverables
- Add ENS PFP support to wallet modal

### Rewards
25 SNX

PR #788, opened 2022-05-02
feat: add ENS PFP to wallet modal

lohba:
## Related issue

Resolves: https://github.com/Kwenta/kwenta/issues/739
Screen Shot 2022-05-02 at 1 17 35 PM


LeifuChen (changes requested):
@lohba Thanks for your efforts. How did you test this feature? I didn't see the avatar by connecting the wallet.

LeifuChen (approved):
@lohba Thanks for the great work. The solution seems not to work for all the avatars but this is the best attempt we could have right now. 
image

PR #790, opened 2022-05-03
Build: upgraded webpack from v4 to v5.

### LINKED ISSUE #741: Next.config still using webpack v4
devDAO members reported issues that we're still using webpack v4 in` next.config.js`. This should be fixed to use webpack 5 as indicated in `package.json`.

Might be also related to issues reported with regard to PR #678 

Screen_Shot_2022-04-21_at_19 09 44

## Reward
25 SNX

zhyd1997:
## Description

Replaced `npm` with `yarn` solved webpack issue, but also need some additional works to do the upgrading task (from v4 to v5).

## Related issue

Closes #741 

## Motivation and Context

[import Uniswap widgets css](https://docs.uniswap.org/sdk/widgets/swap-widget#customizing-width)
[`os` not found](https://github.com/contentful/contentful.js/issues/243#issuecomment-809060496)
[`http` not found](https://github.com/ethers-io/ethers.js/issues/998#issuecomment-673263070)

webpack version fixed


zhyd1997:
Please **do not** trigger PR preview, because I replaced `npm` with `yarn`, which is not compatible with the current CI.

PR #799, opened 2022-05-05
shifting icons to the left and adding active state to svgs

platschi:
## Description
Layout fixes from the `next-price` branch Yashar introduced that were commited after the `next-price` branch was squashed. Readding them here.


asherism (approved):
looks good 🚀

PR #812, opened 2022-05-09
Chore: Update node to v16 and npm to v8

platschi:
Vercel now deploys with node v16. When switching, we received the following warning:

> Warning: Due to "engines": { "node": ">=14" } in your package.json file, the Node.js Version defined in your Project Settings ("16.x") will not apply. Learn More: http://vercel.link/node-version

Therefore, we should update our "engines" to node v16 (and npm v8 respectively) which are the latest LTS versions of each.

This update is only for `v2-dev`, as it potentially have issues with the outdated v1 (main/dev) branches. This update should not affect them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,7 +160,7 @@
 		"testEnvironment": "jest-environment-jsdom"
 	},
 	"engines": {
-		"node": ">=14",
-		"npm": ">=7"
+		"node": ">=16",
+		"npm": ">=8"
 	}
 }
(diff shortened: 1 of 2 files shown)


brossetti1 (approved):
@platschi I do not see any issues with upgrading node, since kovan is down, I cannot test any interactions as I do not have access to the test wallet. anyway you can test deposit, trade, close position, withdrawal, synth swap before merging this?

if kovan comes back online i will make sure to swing back around and test these interactions.

Issue #688, opened 2022-04-13
Navigation component reloading on each page refresh


platschi:
### Subject of the issue
The navigation component is reloading on each page refresh, making a lot of calls to the network. With the sUSD balance implemented in the navigation component (- #603), it causes the wallet balance amount to flicker and continously reloads, shortly showing the "Get sUSD" text every time you navigate within Kwenta. Similar situation with the wallet address, that shows "loading..." on each page change.

### Solution
The navigation component should be permanent, and only elements below the navigation should reload on each page change.

PR #821, opened 2022-05-10
Layout changes

brossetti1:
the header continuously loads on every page change, causing alot of unnecessary calls to be made.

## Description
 This pr consolidates the layout one layer up into `/pages/_app.ts` - nextjs recommends this for a single layout that is shared across pages.

https://nextjs.org/docs/basic-features/layouts#single-shared-layout-with-custom-app - if we need a different layout there are examples of the best way to go about this in their docs, for now, we only use one layout.

## Related issue
https://github.com/Kwenta/kwenta/issues/688

## Motivation and Context
need to fix reloads

## How Has This Been Tested?
locally, on build

current state of switching:
https://user-images.githubusercontent.com/5998100/167944440-cf04160a-5d8d-4593-bd1e-7c6b6ac0934c.mov

see deployed branch to tell the difference

### DIFF
--- a/pages/404.tsx
+++ b/pages/404.tsx
@@ -3,8 +3,6 @@ import { useTranslation } from 'react-i18next';
 import styled from 'styled-components';
 import { Svg } from 'react-optimized-image';
 
-import AppLayout from 'sections/shared/Layout/AppLayout';
-
 import media from 'styles/media';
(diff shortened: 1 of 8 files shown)


platschi (approved):
Works great on deploy, no more flickering.

LeifuChen (approved):
@brossetti1  Good job!

Issue #810, opened 2022-05-09
Op Kovan ETH-PERP bug


platschi:
After opening an ETH position on Optimism Kovan, it is shown three times. On the same line, XAU and XAG markets are enabled but not working (Unexpected error) when visiting them on Optimism Kovan.

<img width="1016" alt="Screen Shot 2022-05-09 at 16 40 07" src="https://user-images.githubusercontent.com/548702/167485142-c2f7ba46-65c5-4f76-a1c9-08ef27dde552.png">

PR #828, opened 2022-05-11
Fix the price of dydx and ape

LeifuChen:
## Description
Add dydx and ape to coingecko price list

## Related issue
#810 

image
image


brossetti1 (approved):
code looks good! tested deploy and all is working correctly

PR #838, opened 2022-05-12
Fix the build error

LeifuChen:
## Description
A fix to #828.
Fix the build error of `useRateUpdateQuery.ts` and `MarketDetails` (big.js invalid number error could be observed in the console and logs when building and opening the market page the first time).

## Related issue
#828 

## Motivation and Context
Remove the error in logs and browser console. Keep the code base clean.

## How Has This Been Tested?
1. run `npm run build` and no error founded
2. open the market page on the browser, no big.js invalid number error reported in the develop tools

### DIFF
--- a/constants/queryKeys.ts
+++ b/constants/queryKeys.ts
@@ -228,6 +228,12 @@ export const QUERY_KEYS = {
 			networkId,
 			walletAddress,
 		],
+		LatestUpdate: (networkId: NetworkId, market: string | null) => [
+			'futures',
+			'latestUpdate',
+			networkId,
+			market,
+		],
 		NextPriceDetails: (
 			networkId: NetworkId,
 			walletAddress: string | null,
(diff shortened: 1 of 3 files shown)


koredefashokun (approved):

koredefashokun:
Approved @LeifuChen. Wanted to check if there was a reason that we're getting rid of the `lastOracleUpdateTime` stuff though. Also not sure why we're refetching the `useGetFuturesMarkets` query. Is there something that changes on it that frequently?

LeifuChen:
> Approved @LeifuChen. Wanted to check if there was a reason that we're getting rid of the `lastOracleUpdateTime` stuff though. Also not sure why we're using a `refetchInterval` on the `useGetFuturesMarkets` query. Is there something that changes on it that frequently?

Hello @koredefashokun, thanks for your review.
1. The current implementation of `useRateUpdateQuery` throws the exception when `synth` is null (you can find this error in the log of `npm run build`). 
2. Just replace `useExchangeRatesQuery` with `useGetFuturesMarkets` since no synthetic assets for APE and DYDX. 
image

PR #839, opened 2022-05-12
Css fixies

asherism:
mostly trades table styling like time format, external link icon, etc


platschi:
There's a conflict with the TimeDisplay component we introduced this release, so the time styling should probably happen in there.

components/Currency/CurrencyPrice/CurrencyPrice.tsx
```diff
@@ -10,6 +10,7 @@ import { formatCurrency, FormatCurrencyOptions } from 'utils/formatters/number';
 
 import { ContainerRowMixin } from '../common';
 import { ethers } from 'ethers';
+import { composeInitialProps } from 'react-i18next';
```
koredefashokun, line 13: I think we can remove this line since we don't use it.

koredefashokun (approved):
Code changes look good.

koredefashokun:
Just noticed @platschi's comment as well. Also a bit worried about the number of commits compared to the actual number of changes. Did we carry over some stuff from another branch?

platschi:
> Just noticed @platschi's comment as well. Also a bit worried about the number of commits compared to the actual number of changes. Did we carry over some stuff from another branch?

Yeah lots of overlap with #806.  I guess we should make sure to set up new branches once a PR is submitted

platschi:
> > Just noticed @platschi's comment as well. Also a bit worried about the number of commits compared to the actual number of changes. Did we carry over some stuff from another branch?
> 
> Yeah lots of overlap with #806. I guess we should make sure to set up new branches once a PR is submitted

@asherism Can you look at this?

platschi (approved):

PR #841, opened 2022-05-12
Fix 24h volume on L2

LeifuChen:
## Description
The original code in #783 couldn't process the trades without a currency symbol. Here is the fix to handle this issue.
image

## Motivation and Context
To try our best to show the 24h volume on L2 with the broken subgraph.

## How Has This Been Tested?
1. Open the page and check the 24 volume column

image


platschi:
Added Spot Markets to "Markets" tab, it was still disabled there

Screen Shot 2022-05-12 at 16 25 15

platschi (approved):
Looks good for a temp solution, just need somebody else to approve leifus and my commits here in this PR

brossetti1 (approved):
agreed - works for me, code looks good here

Issue #503, opened 2022-03-23
Create Tooltip Component


platschi:
### Suggestion
Create a reusable tooltip component for v2.
### Deliverables

- [x] Create reusable tooltip component for Futures v2
- [x] Match the reusable tooltip component to use the v2 tooltip design (see example and css below)
- [x] Replace existing Open Interest tooltip in `MarketDetails` component with new reusable tooltip component

### Resources
v2 Tooltip Design
![Screen Shot 2022-03-25 at 16 16 14](https://user-images.githubusercontent.com/548702/160187018-5f4659b3-6cd9-4eda-94bc-7958b981ae18.png)

v2 CSS from Figma design (use as appropriate):
```css
/* Rectangle 215 */
width: 189px;
height: 50px;


/* D. Cell Gradient */
background: linear-gradient(180deg, #1E1D1D 0%, #161515 100%);
border: 1px solid rgba(255, 255, 255, 0.1);
box-sizing: border-box;
/* D. Hilite-Glo-Shadow */
box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.25), inset 0px 1px 0px rgba(255, 255, 255, 0.1), inset 0px 0px 20px rgba(255, 255, 255, 0.03);
border-radius: 10px;
```

```css
/* Liquidity Mining Program will end on March 23rd, 2021. */
width: 160px;
height: 24px;

font-family: 'Akkurat LL';
font-style: normal;
font-weight: 400;
font-size: 12px;
line-height: 12px;

/* Primary White */
color: #ECE8E3;
```

### Reward
50 SNX

PR #844, opened 2022-05-13
feat: add tooltips to markets

### LINKED ISSUE #461: Markets: Add Tooltips
### Subject of the issue
Ticket #448 introduced a tooltip component, which should make it easier to show tooltips on the UI. We want to introduce more tooltips to aid traders throughout the UI.

All text should be added to `en.json` in order to allow for easy translation in the future.

### `PositionCard`

- Position Size: `Notional value of the open position in USD and asset denomination`
- Leverage: `Denotes individual position leverage, which will vary based on the balance of margin deposited into the market` 

*Leverage tooltip text might need to be changed to be easier understandable*

- Net Funding: `Total funding due at the close of the position`
- Avg. Entry Price: `The average price at which the position was opened`

- Position Side: `Direction of current trade`
- Unrealized P&L: `The current unrealized profit or loss on the open position`
- Liq. Price: `The estimated oracle price that will trigger a liquidation`
- Fees: `Net fees paid to open and/or modify the position
`

### `MarketDetails`

- External Price: `Aggregated market data provided by CoinGecko`
- 24H Change: `Total change in asset value within the past 24 hours`
- 24H Volume: `Total trade volume within the past 24 hours`
- 24H Trades: `Total amount of trades on selected pair within the past 24 hours`
- 1H Funding Rate: `Funding applies to all open positions. A positive rate means longs pay shorts, a negative rate means shorts pay longs. The rate is calculated as the average funding paid by open positions for the last hour. Funding accrues on a position every time it is opened, modified, or closed.`

## Reward
40 SNX

## Deadline
May 18th, 2022

0xwontonsoup:
## Description
Fixes #461 

## Related issues
* #503 
* #448

## Motivation and Context
This PR uses the new custom tooltip introduced in #503 to add tooltips to the `PositionCard` and `MarketDetails` components

## How Has This Been Tested?
Performed simple QA testing on the `MarketDetails` component:
1. Is the new tooltip text being displayed?
2. Is the text in the tooltip correct?
3. Are there any noticeable bugs/issues?
4. Is each tooltip text added to `en.json`?

I did _not_ test the `PositionDetails` component because I did not have any positions open to test with

Screen recording of Tooltips working on `MarketDetails` component:

https://user-images.githubusercontent.com/99768011/168385406-e5d10f0a-ac96-4766-8594-4b410657901f.mov


platschi:
Hey, on a first glance it looks good!

The tooltips in the PositionCard however seem to have a transparent background compared to those in the MarketDetails card, making them difficult to read. We should make sure they're the same (taking MarketDetails tooltips as a base)

Screen Shot 2022-05-17 at 10 09 49

platschi:
@0xwontonsoup thanks for the fixes!  One last lint warning I noticed, kindly have a look at https://github.com/Kwenta/kwenta/pull/844/files#diff-318ad35f575da4bd3219252cf8285104f303d984b319470acf46ec44d444f4d6L217

translations/en.json
```diff
@@ -649,7 +656,17 @@
...
 				"net-funding": "Net Funding",
 				"fees": "Fees",
-				"avg-entry-price": "Avg. Entry"
+				"avg-entry-price": "Avg. Entry",
+				"tooltips": {
+					"position-side": "Direction of current trade",
+					"position-size": "Notional value of the open position in USD and asset denomination",
+					"u-pnl": "The current unrealized profit or loss on the open position",
+					"leverage": "Denotes individual position leverage, which will vary based on the balance of margin deposited into the market",
+					"liquidation-price": "THe estimated oracle price that will trigger a liquidation",
```
platschi, line 665: Typo (T**H**e)

platschi:
Seems like tooltips for "Net Funding" and "Unrealized PnL" on the PositionCard are missing.

components/Tooltip/TooltipStyles.tsx
```diff
@@ -55,6 +57,41 @@ export const Tooltip = styled.div<ToolTipStyleProps>`
...
+			z-index: 2;
+		`}
+		
+		${(props) =>
+			props.preset === 'bottom-z-index-2' &&
+			`
+				bottom: 0;
+				transform: translate(-25%, 125%);
+				z-index: 2;
+			`}
```
koredefashokun, line 94: Instead of creating multiple presets, maybe we can handle the individual cases when using the components i.e. we can add the custom styles when we do `styled(Tooltip)`. It might clean things up.

Hardcoding it here makes the Tooltip component a little bit more complex.

koredefashokun (commented):
The code looks good. Added a comment related to tooltip presets vs custom styles. It's a bit of a nitpick, but it makes the Tooltip component a little less complicated. Also think that `React.ReactNode` might be a bit more consistent with existing code than `JSX.Element`.

0xwontonsoup:
> The code looks good. Added a comment related to tooltip presets vs custom styles. It's a bit of a nitpick, but it makes the Tooltip component a little less complicated. Also think that `React.ReactNode` might be a bit more consistent with existing code than `JSX.Element`.

@koredefashokun When i use `React.ReactNode` i get an error saying that data can be `null`

koredefashokun:
@0xwontonsoup My bad. Looks good.

koredefashokun (approved):

Issue #848, opened 2022-05-17
Menkent atomic swap minAmount issue


platschi:
Atomic swaps not fully functional for all synths on v2 resulting in minAmount errors.

![issue](https://user-images.githubusercontent.com/548702/168831918-8c4e3248-fd48-4d38-b4a1-b61b0c895980.png)

For relevant info, see conversation on Discord.

PR #851, opened 2022-05-17
Menkent fix add a hardcoded slippage and susd

LeifuChen:
## Description
1. Add a hardcoded slippage (1%) into atomic exchange
2. Add `sUSD` into atomic exchange

## Related issue
#848 

## Motivation and Context
A quick fix to the current implementation.

## How Has This Been Tested?
1. Make a swap between sUSD => sEUR on L1
2. Check the transaction function and confirm it is using `exchangeAtomically`

https://kovan.etherscan.io/tx/0x8449e7673e458b02d5c2eeea601e3fcae9d9b8eefb83a80670700d7cf23410a2
image

### DIFF
--- a/sections/exchange/hooks/useExchange.tsx
+++ b/sections/exchange/hooks/useExchange.tsx
@@ -141,6 +141,7 @@ const useExchange = ({
 	const [selectQuoteTokenModalOpen, setSelectQuoteTokenModalOpen] = useState<boolean>(false);
 	const [selectBaseTokenModalOpen, setSelectBaseTokenModalOpen] = useState<boolean>(false);
 	const [txApproveModalOpen, setTxApproveModalOpen] = useState<boolean>(false);
+	const [atomicExchangeSlippage] = useState<string>('0.01');
 	const setOrders = useSetRecoilState(ordersState);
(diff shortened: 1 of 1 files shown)


sections/exchange/hooks/useExchange.tsx
```diff
@@ -141,6 +141,7 @@ const useExchange = ({
 	const [selectQuoteTokenModalOpen, setSelectQuoteTokenModalOpen] = useState<boolean>(false);
 	const [selectBaseTokenModalOpen, setSelectBaseTokenModalOpen] = useState<boolean>(false);
 	const [txApproveModalOpen, setTxApproveModalOpen] = useState<boolean>(false);
+	const [atomicExchangeSlippage] = useState<string>('0.01');
```
koredefashokun, line 144: Can this be a constant? Not sure if we need to make this state, since it's not changing at any point.

LeifuChen: Yes. This is a quick fix so we used the hardcoded value but we may change the slippage rate as a user-defined value later.
image

koredefashokun: Okay makes sense then.

koredefashokun (approved):

Issue #842, opened 2022-05-13
Synth Exchange Input broken


platschi:
### Subject of the issue
When using the Synth Exchange Input "To" field, typing "500" results in 5.00000 with no way for the user to use the backspace button to remove the values and enter new values. See for an illustration attached screen recording.

https://user-images.githubusercontent.com/548702/168290773-224239b5-e428-4365-a9d6-331b660fc8eb.mov

PR #852, opened 2022-05-17
Fix the broken input issue for into field

LeifuChen:
## Description
The user can't use the backspace button to remove the values and enter new values in the `into` fields.

## Related issue
#842

## Motivation and Context
It affects the UX, especially for the user who wants to receive the exact amount of the `into` currency.

## How Has This Been Tested?
1. Select `sLink` as the `baseCurrency` and type `500` in the `into` field.
2. Select `sLink` as the `quoteCurrency` and type `500` in the `from` field.

image
image

### DIFF
--- a/queries/synths/useBaseFeeRateQuery.ts
+++ b/queries/synths/useBaseFeeRateQuery.ts
@@ -6,7 +6,6 @@ import { ethers } from 'ethers';
 import { CurrencyKey } from 'constants/currency';
 import { appReadyState } from 'store/app';
 import QUERY_KEYS from 'constants/queryKeys';
-import Wei from '@synthetixio/wei';
 
 const useBaseFeeRateQuery = (
 	sourceCurrencyKey: CurrencyKey | null,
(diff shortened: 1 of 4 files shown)


queries/synths/useBaseFeeRateQuery.ts
```diff
@@ -20,18 +19,14 @@ const useBaseFeeRateQuery = (
 		async () => {
 			const { SystemSettings } = synthetixjs!.contracts;
 
-			const [sourceCurrencyFeeRate, destinationCurrencyFeeRate] = (await Promise.all([
-				new Wei(
-					SystemSettings.exchangeFeeRate(
-						ethers.utils.formatBytes32String(sourceCurrencyKey as string)
-					)
+			const [sourceCurrencyFeeRate, destinationCurrencyFeeRate] = await Promise.all([
+				SystemSettings.exchangeFeeRate(
```
LeifuChen, line 23: `new Wei` function throws the `[big.js] invalid number` error because `SystemSettings.exchangeFeeRate` may return the value that it couldn't process.

platschi (approved):
Deployment looks good and works fine now 👍

Issue #496, opened 2022-03-22
Show Dynamic Fee


platschi:
### Subject of the issue
Display and break down the dynamic fee / skew fee to give users a better understanding of the fees involved

PR #858, opened 2022-05-19
Add dynamic fee

brossetti1:
adding the dynamic fee to the fee cost display

## Description

- [x] Show the currently applying exchange fee of the market in question in the Fee box (bottom right of Markets page, see screenshot).
- [x] Print the dynamic fee % if it applies, in the same yellow color & clock logo as seen on the Synth Exchange. Example:
- [x] Add tooltip on nextprice order when dynamic fee is active
- [x] Adding tooltip functionality to InfoBox

## Related issue
https://github.com/Kwenta/kwenta/issues/496

## Motivation and Context
we want to display the basis points being charged and fetch the dynamic fee in times of volatility.

## How Has This Been Tested?
locally on kovan and mainnet

Screen Shot 2022-05-19 at 9 34 38 AM
Screen Shot 2022-05-19 at 3 11 33 PM


components/InfoBox/InfoBox.tsx
```diff
@@ -22,7 +22,9 @@ const InfoBox: React.FC<InfoBoxProps> = ({ details, style, className, disabled }
 			{Object.entries(details).map(([key, value]) => (
 				<React.Fragment key={key}>
 					<div>
-						<p className="key">{key}:</p>
+						<div className="key">
+							{key}:{value.tooltip}
```
brossetti1, line 26: maybe this should be this instead:

```suggestion
{key}{value.tooltip}:
```

platschi: 1) Yeah, let's add some spacing so it looks like this:

Screen_Shot_2022-05-20_at_09 57 01

2) Can we also make sure the tooltip also shows when hovering the (yellow text) dynamic fee value, and not only when hoving the icon?

brossetti1: @platschi done - https://github.com/Kwenta/kwenta/pull/858/commits/a1b507aa53256d677e3d46d08a08763f75dbbbf4

platschi (approved):
Thanks Brian! Works and looks great now, very useful addition!

Screen_Shot_2022-05-23_at_11 11 07

LeifuChen (commented):
@brossetti1 Great job. Just leave a minor comment .

LeifuChen (approved):

PR #859, opened 2022-05-19
Small tooltip improvements

platschi:
Tiny text label improvements for tooltips we noticed while discussing with Marko.


Marko-roy (approved):
LGTM! Thanks for these last-minute changes.

PR #860, opened 2022-05-19
Fix the invalid balance and max button click on quote currency card

LeifuChen:
## Description
When the user clicked on the "Max" and "balance" buttons, the input field was not filled automatically. 
The root cause is that the `balanceWithBuffer` and `quoteCurrencyBalance` are strings so the `toFixed` throw the error.

## Related issue
#842 

## Motivation and Context
It affects the UX.

## How Has This Been Tested?
1. Clicked on "Max" and the input was filled automatically
2. Clicked on "balance" and the input was filled automatically


platschi (approved):
Thanks for the fix, works now as expected!

Issue #648, opened 2022-04-08
PositionCard & MarginBox: Preview Modified Positions


platschi:
### Subject of the issue
When adding data to the input fields in the Trade section, the PositionCard and MarginBox should show a preview of how values change (see designs), otherwise just show the current Position/Margin data.

### Related tickets
- #563
- #716

<img src="https://user-images.githubusercontent.com/11524787/163269679-5a40c0c6-06a2-493e-8257-8081692f280f.png"/>

<img width="238" alt="marginbox" src="https://user-images.githubusercontent.com/548702/169853931-543b252e-653e-4aba-9162-e9b694a44fbc.png">

### Deliverables
Preview values (in gold color) should only show if the values change. Otherwise, omit preview. 

Format is  `Current Open Position Value -> Preview Value`

#### PositionCard

- [ ] Side:  Show a preview if "Side" changes with input values. Do not show if side stays the same (e.g. Long -> Long)
- [ ] Size: Show change in position asset size.
- [ ] Liquidation Price: Show estimated liquidation price with given input values
- [ ] Leverage: Show estimated Leverage change
- [ ] Avg. Entry: Calculate estimated Avg. Entry Price from input values 

#### Available Margin Box

- [ ] Available Margin: Calculate change in available margin from input values
- [ ] Buying Power: Calculate change in buying power with current input values
- [ ] Margin Usage: Calculate change in Margin Usage with current input values

### Reward
200 SNX

PR #885, opened 2022-05-24
Add preview trade

brossetti1:
add preview trade to `PositionCard` and `MarketInfo`

#### PositionCard

- [x] Side:  Show a preview if "Side" changes with input values. Do not show if side stays the same (e.g. Long -> Long)
- [x] Size: Show change in position asset size.
- [x] Liquidation Price: Show estimated liquidation price with given input values
- [x] Leverage: Show estimated Leverage change
- [x] Avg. Entry: Calculate estimated Avg. Entry Price from input values 

#### Available Margin Box

- [x] Available Margin: Calculate change in available margin from input values
- [x] Buying Power: Calculate change in buying power with current input values
* Buying power is still off, but its the closest I can get it. In regards to the`MarketInfo` preview, the values are not necessarily easily accessible from the contract, but I have gotten them close
- [x] Margin Usage: Calculate change in Margin Usage with current input values

#### Details

- [X] incorporating error status and [status messaging into preview logic](https://github.com/Kwenta/kwenta/pull/885/files#diff-5a0774336d950c6de6c2b7f214fd8e4f57bdf96a0495d65e70a8da273d36f42fR79-R107)

I integrated these [previews with the error messaging](https://github.com/Synthetixio/synthetix/blob/4d520d726bc013f2642dceb1dad4073fc78f4859/contracts/FuturesMarketBase.sol#L470-L563) that exists on `_postTradeDetails`. Long story short here, when this function fails because of several reasons it will return the `CURRENT` position details with the error message. In order to prevent this from interfering with previewing a trade, I incorporated the error messaging in to the `Trade` component and hide the preview trade when the `_postTradeDetails` returns an error.

## Related issue
https://github.com/Kwenta/kwenta/issues/648

## How Has This Been Tested?
locally on kovan

Screen Shot 2022-05-31 at 5 50 05 PM
Screen Shot 2022-05-31 at 10 01 17 AM


platschi:
Small detail for Size Preview, kindly add space between the yellow Size and USD volume in brackets:

Screen Shot 2022-05-31 at 09 34 53

sections/futures/MarketInfoBox/MarketInfoBox.tsx
```diff
@@ -19,20 +25,82 @@ const MarketInfoBox: React.FC<MarketInfoBoxProps> = ({
...
+		currencyKey as CurrencyKey,
+		potentialTrade
+	);
+
+	const previewTrade = potentialTradeDetails.data ?? null;
+
+	const previewTradeData = React.useMemo(() => {
+		const size = wei(potentialTrade?.size || zeroBN);
+		const potentialAvailableMargin =
+			previewTrade?.margin?.sub(previewTrade?.minInitialMargin) ?? zeroBN;
```
Tburm, line 42: `minInitialMargin` doesn't change based on the position side (long/short) so this isn't correct. It looks like the value is always positive, but some trades will actually increase available margin (long while having an open short position)

From our discussion: `total margin - (notional value * max leverage)`

So the formula would look more like (with better validation):
`previewTrade.margin.sub(potentialTrade.size.mul(maxLeverageValue))`

sections/futures/MarketInfoBox/MarketInfoBox.tsx
```diff
@@ -19,20 +25,82 @@ const MarketInfoBox: React.FC<MarketInfoBoxProps> = ({
 	buyingPower,
 	marginUsage,
 	isMarketClosed,
+	potentialTrade,
+	maxLeverageValue,
```
Tburm, line 29: One issue I'm seeing is that this is not the actual "max leverage" for the market. This is the max leverage available to the user.

When I say "max leverage" I mean 10x, however when you have a position open this value is <10. This is causing issues, and you'll need to figure out how to get the actual static maximum in this component.

sections/futures/MarketInfoBox/MarketInfoBox.tsx
Tburm: See above comment on "max leverage" and how this is incorrect.

sections/futures/MarketInfoBox/MarketInfoBox.tsx
```diff
@@ -19,20 +25,82 @@ const MarketInfoBox: React.FC<MarketInfoBoxProps> = ({
...
+
+	const previewTrade = potentialTradeDetails.data ?? null;
+
+	const previewTradeData = React.useMemo(() => {
+		const size = wei(potentialTrade?.size || zeroBN);
+		const potentialAvailableMargin =
+			previewTrade?.margin?.sub(previewTrade?.minInitialMargin) ?? zeroBN;
+		const potentialBuyingPower = previewTrade?.margin?.mul(maxLeverageValue) ?? zeroBN;
+		const enumeratorMarginUsage = previewTrade?.margin?.sub(potentialAvailableMargin);
+		const potentialMarginUsage = enumeratorMarginUsage?.div(previewTrade?.margin)?.abs() ?? zeroBN;
```
Tburm, line 45: From our discussion: `margin usage = notional value / buying power`

I don't think "available margin" is relevant in this context at all. It should be something like:
`previewTrade.size.mul(previewTrade.price).div(potentialBuyingPower)`

sections/futures/MarketInfoBox/MarketInfoBox.tsx
```diff
@@ -19,20 +25,82 @@ const MarketInfoBox: React.FC<MarketInfoBoxProps> = ({
...
+		potentialTrade
+	);
+
+	const previewTrade = potentialTradeDetails.data ?? null;
+
+	const previewTradeData = React.useMemo(() => {
+		const size = wei(potentialTrade?.size || zeroBN);
+		const potentialAvailableMargin =
+			previewTrade?.margin?.sub(previewTrade?.minInitialMargin) ?? zeroBN;
+		const potentialBuyingPower = previewTrade?.margin?.mul(maxLeverageValue) ?? zeroBN;
```
Tburm, line 43: This is incorrect because the `maxLeverageValue` is dependent on the position, not the actual max leverage of 10x. See comment above.

Tburm (commented):
Left some comments where I see issues with the calculations. I think the key issue is the lack of actual "max leverage" meaning for the entire market (always 10x for now). Instead the "max leverage" refers to the max leverage they can take given their current position, which isn't useful for calculating buying power.

Another issue I noticed is that values don't change when selecting "next price" vs "market" orders. Users pay deposits on top of the fees, so that should be taken into account for their potential margin after the trade.

sections/futures/MarketInfoBox/MarketInfoBox.tsx
```diff
@@ -1,38 +1,125 @@
...
+	potentialTrade,
+	marketMaxLeverage,
+	currencyKey,
 }) => {
+	const totalMargin = position?.remainingMargin ?? zeroBN;
+	const availableMargin = position?.accessibleMargin ?? zeroBN;
+	const buyingPower =
+		position && position?.remainingMargin.gt(zeroBN)
+			? position?.remainingMargin?.mul(marketMaxLeverage ?? zeroBN)
+			: zeroBN;
```
Tburm, line 32: Need to merge in `v2-dev` and update this to match the buying power calculation changes

sections/futures/MarketInfoBox/NextPriceInfoBox.tsx
```diff
@@ -0,0 +1,150 @@
...
+	}, [commitDeposit, nextPriceDetails?.keeperDeposit]);
+
+	// const nextPriceDiscount = useMemo(() => {
+	// 	return (nextPriceFee ?? zeroBN).sub(commitDeposit ?? zeroBN);
+	// }, [commitDeposit, nextPriceFee]);
+
+	const previewTradeData = React.useMemo(() => {
+		const potentialAvailableMargin = availableMargin?.sub(totalDeposit) ?? zeroBN;
+		const potentialBuyingPower =
+			totalMargin.sub(totalDeposit)?.mul(marketMaxLeverage)?.abs() ?? zeroBN;
```
Tburm, line 76: This should include the estimated trade as well as the deposit. There is no guarantee this will match exactly with the actual executed trade, however it's more confusing to ignore the trade itself in this calculation.

In this example I'm placing a trade that will take me almost to max leverage, however buying power stays almost the same:
image

brossetti1:
ok updated the logic - and refactored it back down

sections/futures/MarketInfoBox/NextPriceInfoBox.tsx
brossetti1: @Tburm if you could try this one again - wondering what the result will be on the latest deployment

Tburm (approved):
@brossetti1 take a look at the commit I just pushed. There were some cases at max leverage where things were wrong, and the leverage value was usually off by a bit.

Changes are:
* Handle the case where "inaccessible margin" is higher than margin in the preview. If this was larger, it would be negative and you were taking the absolute value, treating it as a positive number
* Your leverage calculation was `current leverage + trade leverage` but we can just calculate it directly: `notional size / margin`

PR #908, opened 2022-05-27
Tradingview Rerenders

Tburm:
Fixing issues with rerendering and flickering of the TradingView chart.

## Description
- Apply style overrides on re-render
- Avoid re-rendering the entire component when the currency changes
- Store interval selection when switching currency
- Get the current asset from the route
- Fix height changing on first render
- Fix layout heights on first render (issue was `BalanceAction` in the nav)

## Issues
- #1080 

## Motivation and Context
Chart flickering is annoying. We are also not storing objects in localstorage, which means the chart always displays a "hint" on refresh.

## How Has This Been Tested?
In current v2...
* Select any market
* Change to 5m candles
* Change the market
* Chart will "flicker", and reset things like the selected interval


Tburm:
@koredefashokun I can't exactly figure out the excessive flickering that happens at the beginning of the render. Did you mention you knew what was going on here?

koredefashokun:
@Tburm I think so (sorry for the late reply). I just refreshed a bunch of times and noticed that the container that holds the iframe doesn't have a fixed height, so when it first loads, it's empty. Basically, the flicker is the layout shifting.

koredefashokun (approved):

PR #909, opened 2022-05-27
Futures mobile

koredefashokun:
## Description
This pull request contains the initial implementation of mobile futures design for Kwenta. It also includes changes to state management for the futures page, mainly around making sure that desktop and mobile can share state. This is also the first step to building a robust state layer for the futures page that should reduce requests, improve performance, clean up code and make adding and testing new changes much easier.

## Related issue
N/A

## Motivation and Context
This PR is the second phase of the broader mobile responsiveness updates to the Kwenta v2 UI.

## How Has This Been Tested?
N/A


hooks/useFuturesData.ts
```diff
@@ -0,0 +1,225 @@
...
+		marketAsset,
+		position,
+		leverageSide,
+		walletAddress,
+		isLeverageValueCommitted,
+		sizeDelta,
+		setFeeCost,
+	]);
+
+	return {
```
Tburm, line 206: Should we return `orderType`?

sections/futures/MarketInfo/MarketInfo.tsx
```diff
@@ -57,14 +42,9 @@ const MarketInfo: FC<MarketInfoProps> = ({
...
-			<PositionChart marketAsset={baseCurrencyKey} potentialTrade={potentialTrade} />
-			<UserInfo
-				marketAsset={baseCurrencyKey}
-				position={position}
-				openOrders={openOrders}
-				refetch={refetch}
-			/>
+			<MarketDetails />
+			<PositionChart />
+			<UserInfo />
```
Tburm, line 47: 💯

sections/futures/MobileTrade/MobileDepositModal.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const MobileDepositModal = () => {
+	return <div />;
+};
```
Tburm, line 5: Is this a placeholder or something?

sections/futures/MobileTrade/UserTabs/UserTabs.tsx
```diff
@@ -0,0 +1,61 @@
...
+import TabButton from 'components/Button/TabButton';
+
+import OpenPositionTab from './OpenPositionTab';
+import OrdersTab from './OrdersTab';
+import TradesTab from './TradesTab';
+import TransfersTab from './TransfersTab';
+
+const TABS = [
+	{
+		title: 'Open Position',
```
Tburm, line 13: I think this should be called "Position" since they could have one open already

sections/futures/MobileTrade/drawers/TradeConfirmationDrawer.tsx
```diff
@@ -0,0 +1,149 @@
...
+			closeDrawer={closeDrawer}
+			items={dataRows}
+			buttons={
+				<ConfirmTradeButton
+					variant="primary"
+					isRounded
+					onClick={() => {
+						onConfirmOrder();
+						closeDrawer();
+					}}
```
Tburm, line 131: Would be cool if this reset the inputs as well. There is no indication that there is a pending transaction, until it goes through and the position updates.

Tburm: Another option would be to leave the drawer open until the transaction is complete. While it's pending, place a loading spinner in the button. When it succeeds, close the drawer.

sections/futures/MobileTrade/UserTabs/TransfersTab.tsx
```diff
@@ -0,0 +1,117 @@
...
+					sortable: true,
+					width: 50,
+				},
+				{
+					Header: (
+						<StyledTableHeader>{t('futures.market.user.transfers.table.date')}</StyledTableHeader>
+					),
+					accessor: 'timestamp',
+					Cell: (cellProps: any) => (
+						<DefaultCell>{timePresentation(cellProps.value, t)}</DefaultCell>
```
Tburm, line 58: This is only successful for my first few transfers:
image

Tburm: Everything on following pages shows a date

sections/futures/MobileTrade/UserTabs/UserTabs.tsx
Tburm, line 13: Can we call this `Position`? Traders might already have a position open.

Tburm (commented):
Checking in with some comments...

* I'm seeing a lot of errors in the console for `formatNumber` but not sure exactly where they're coming from
* "Close Position" button doesn't work, and doesn't throw any specific error

constants/queryKeys.ts
```diff
@@ -222,9 +226,9 @@ export const QUERY_KEYS = {
 		PotentialTrade: (
 			networkId: NetworkId,
 			market: string | null,
-			trade: PotentialTrade | null,
+			tradeSize: string,
```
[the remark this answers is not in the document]
koredefashokun: Yeah. We store it as a string in state (to make it easier to use with inputs).

sections/futures/MobileTrade/MobileDepositModal.tsx
koredefashokun: Yeah. Basically forgot to implement the deposit/withdraw modals for mobile (since there aren't any designs for it).

sections/futures/MobileTrade/UserTabs/UserTabs.tsx
koredefashokun: @asherism What do you think?

Tburm: so "Position" and "Open Orders" instead of "Open Position" and "Orders"

hooks/useFuturesData.ts
```diff
@@ -0,0 +1,268 @@
...
+const DEFAULT_MAX_LEVERAGE = wei(10);
+
+const useFuturesData = () => {
+	const walletAddress = useRecoilValue(walletAddressState);
+	const exchangeRatesQuery = useExchangeRatesQuery();
+	const router = useRouter();
+	const { synthetixjs, network } = Connector.useContainer();
+	const { useSynthetixTxn, useEthGasPriceQuery } = useSynthetixQueries();
+	const { monitorTransaction } = TransactionNotifier.useContainer();
+	const { handleRefetch } = React.useContext(RefetchContext);
```
avclarke, line 45: A nice little helper I've found in the past is to export a hook in the context like `useRefetchContext` just to wrap in it one place instead of calling through useContext on each use.

`const useRefetchContext = () => React.useContext(RefetchContext)`

koredefashokun: True. Thanks @avclarke. Will add the hook for this.

sections/futures/MobileTrade/drawers/TradeConfirmationDrawer.tsx
koredefashokun: I think that makes sense. @asherism maybe we should add loading states for the buttons?

avclarke (commented):
I like the approach on the state management changes, I'm not that familiar with recoil but it seems this is now similar to a redux setup but without all the boilerplate. Although there's a lot of overhead with redux one nice thing you do get is a consistent structure that can be followed throughout the app. Wondering if there's some sort of PageStateManager component we could create, which abstracts the recoil calls and sets some re-usable framework we can use across the app, more just to outline the pattern for other developers. Probably something for a different PR but maybe something we can think about.

hooks/useFuturesData.ts
```diff
@@ -0,0 +1,272 @@
...
+		}
+	);
+
+	useEffect(() => {
+		if (orderTxn.hash) {
+			monitorTransaction({
+				txHash: orderTxn.hash,
+				onTxConfirmed: () => {
+					onLeverageChange('');
+					handleRefetch('modify-position');
```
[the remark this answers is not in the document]
koredefashokun: Yeah. I think that's out of the scope of this PR though. I can add it to the list of things that we need to add in a separate PR.

store/futures/index.ts
```diff
@@ -0,0 +1,125 @@
...
+	key: getFuturesKey('feeCost'),
+	default: null,
+});
+
+export const dynamicFeeState = atom({
+	key: getFuturesKey('dynamicFee'),
+	default: null,
+});
+
+export const leverageValueCommitedState = atom({
```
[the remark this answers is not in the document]
koredefashokun: Thanks lol

hooks/useFuturesData.ts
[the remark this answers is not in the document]
koredefashokun: It's a bit trickier to do here because I believe some of the queries we refetch here depend on subgraph data, as opposed to raw contract data.

sections/futures/MobileTrade/UserTabs/OrdersTab.tsx
```diff
@@ -0,0 +1,212 @@
...
+	const gasSpeed = useRecoilValue(gasSpeedState);
+	const walletAddress = useRecoilValue(walletAddressState);
+	const position = useRecoilValue(positionState);
+	const currencyKey = useRecoilValue(currentMarketState);
+	const openOrders = useRecoilValue(openOrdersState);
+
+	const { handleRefetch } = useRefetchContext();
+
+	const [selectedOrder, setSelectedOrder] = React.useState<any>();
+	const [action, setAction] = React.useState<'' | 'cancel' | 'execute'>('');
```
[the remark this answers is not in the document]
koredefashokun: Will change.

contexts/RefetchContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+	const positionHistoryQuery = useGetFuturesMarketPositionHistory();
+	const marketQuery = useGetFuturesMarket();
+	useGetFuturesPotentialTradeDetails();
+
+	const handleRefetch = (refetchType: RefetchType, timeout?: number) => {
+		setTimeout(() => {
+			switch (refetchType) {
+				case 'modify-position':
+					marketQuery.refetch();
+					positionHistoryQuery.refetch();
```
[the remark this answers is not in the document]
koredefashokun: Yeah. We might be able to look into more fine-tuned polling. What do you think @Tburm. Part of the reason we moved to this new system for refetching is to avoid too much polling though.

However, I'm not sure this new state management/refetch pattern will be very useful in the unlikely event that subraph/contract data does not update before we call `handleRefetch`.

hooks/useFuturesData.ts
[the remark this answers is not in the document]
avclarke: I'm not sure it would make a difference if contract data or subgraph data, it would just be the query listening to a contract event to trigger a refresh, rather than the refresh being called after each transaction.

queries/futures/useGetFuturesMarketPositionHistory.ts
```diff
@@ -11,18 +11,18 @@ import { PositionHistory } from './types';
 import { getFuturesEndpoint, mapTradeHistory } from './utils';
 import { getDisplayAsset } from 'utils/futures';
 import { FUTURES_POSITION_FRAGMENT } from './constants';
+import { currentMarketState } from 'store/futures';
 
-const useGetFuturesMarketPositionHistory = (
-	currencyKey: string | null,
-	options?: UseQueryOptions<any | null>
-) => {
+const useGetFuturesMarketPositionHistory = (options?: UseQueryOptions<any | null>) => {
```
avclarke, line 16: This query didn't follow the same pattern as you included for the others with setting recoil state and seems to only be used in RefetchContext where the value isn't being used. So not sure it's a redundant query, unless I'm missing something.

contexts/RefetchContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+});
+
+export const RefetchProvider: React.FC = ({ children }) => {
+	const { useSynthsBalancesQuery } = useSynthetixQueries();
+	const walletAddress = useRecoilValue(walletAddressState);
+
+	const synthsBalancesQuery = useSynthsBalancesQuery(walletAddress);
+	const openOrdersQuery = useGetFuturesOpenOrders();
+	const positionQuery = useGetFuturesPositionForMarket();
+	const positionHistoryQuery = useGetFuturesMarketPositionHistory();
```
avclarke, line 29: Left a comment below on the query file but it seems this doesn't set any recoil value inside it and is not being used anywhere else, so maybe a redundant query? Or maybe I'm missing something

contexts/RefetchContext.tsx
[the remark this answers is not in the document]
koredefashokun: Yeah. I think we have to look deeper into fixing the underlying issue here. Indefinite polling can't work as a long-term solution, so we probably need a more robust way of handling updates. It would be much easier if we could just have GraphQL subscriptions on the subgraph directly, but I don't think The Graph currently supports that officially (I think they had a couple of issues with the feature and pulled it). At the same time, I'm slightly worried that it might be a little out of scope here. What do you think about us creating a new issue that addresses this and plan a sync with @Tburm and other frontend devs?

contexts/RefetchContext.tsx
koredefashokun: You're right. It's redundant. I noticed we were querying it for some reason on v2-dev, and wanted to maintain parity, but it seems like it's not being used anywhere else.

queries/futures/useGetFuturesMarketPositionHistory.ts
koredefashokun: Nice catch. Pushed a commit that removes the reference to this file. Not sure why we aren't using it anymore, but there's no need to keep calling it.

Tburm (approved):

Issue #845, opened 2022-05-14
Dashboard uPnL % not correct


platschi:
### Subject of the issue
The uPnL expressed in % is different on the dashboard vs what is shown in the PositionCard. We need to make sure values shown are equal. 

Screenshots example trade shows a 2.55% earning on dashboard, but 13.11% earning on PositionCard.

(1) Trade on dashboard
<img width="332" alt="Screen Shot 2022-05-14 at 10 50 28" src="https://user-images.githubusercontent.com/548702/168428662-f43a0390-0416-427b-9929-e571f71c9a25.png">

(2) Same trade in PositionCard
<img width="332" alt="Screen Shot 2022-05-14 at 10 50 20" src="https://user-images.githubusercontent.com/548702/168428669-072b6cac-56f0-4d0c-a0f6-5d60feadc3c0.png">

PR #934, opened 2022-06-02
Change position card order

Tburm:
Move the uPnL value to reduce confusion

## Description
The uPnL value was directly above the "avg entry price" value, however the two are not related. It could cause trader confusion for the values to be seated next to each other.

This change moves the uPnL above the rPnL and the liquidation price above the avg entry price.

## Related issue
- #845 

image


asherism (approved):
looks good 👌

Issue #234, opened 2022-01-03
Update E2E test for futures


JChiaramonte7:
### Subject of the issue
Extend the testing suite for e2e tests of Kwenta. Tests should be done on Optimistic Kovan. More details see [this page in our docs](https://docs.kwenta.io/dao/contribute/devdao-contribute/testing).

### Deliverables
To make sure Futures work, we would want the following e2e tests wrapped up:

- [ ] Test depositing sUSD in ETH-PERP market
- [ ] Test opening an ETH-PERP long
- [ ] Test closing the ETH-PERP long
- [ ] Test withdrawing sUSD from ETH-PERP market

### Rewards
100 SNX

PR #955, opened 2022-06-06
Test/end 2 end test

joeyzhaozhao:
## Description
end to end testing for futures page

## Related issue
https://github.com/Kwenta/kwenta/issues/234


JChiaramonte7:
@avclarke can you review this one and make sure it meets all expected test cases

we'll also need to write a CI workflow to automate this test in the future (new ticket cc @platschi)

PR #957, opened 2022-06-07
widescreen & tablet responsiveness css

asherism:
-adding fixed side panels at 1725px viewport width
-hiding left side panel at 1200px vw


asherism:
> @asherism Can we rebase these changes against `v2-dev` ? The light theme `themes-switch` branch PR has already been merged into `v2-dev`

image

seems like there is a file missing- @koredefashokun, do you know which branch this might be from?

koredefashokun:
@asherism I'm not sure, but if I remember correctly, I probably deleted it, since we don't use it anymore.

platschi (approved):

Issue #872, opened 2022-05-23
Implement v2 website mobile version


platschi:
### Deliverables
- Implement mobile version of v2 website

### Related tickets
- #780

PR #958, opened 2022-06-07
Mobile new landing page

LeifuChen:
## Description
Implement the v2 website mobile version 

## Related issue
#872

## Motivation and Context
Better UX

## How Has This Been Tested?
Open the mobile to see the new landing page

image
image
image
image
image
image


LeifuChen:
Here is the collection of the feedbacks from @asherism @platschi @JChiaramonte7
- [ ] lots of empty space between the CTA box
image

- [ ] misaligned the footer on the screen over 400px
image

- [ ] overflow on the screen below 375px

- [ ] cut-off effects on blob

- [ ]  icons should be the same height as the logo roughly
image

- [ ] same size of the icons (mirror icon is too big)
image

- [ ] remove the  tags in these [files](https://github.com/Kwenta/kwenta/pull/991/files) . Here is the [reference](https://nextjs.org/docs/messages/no-img-element)

- [ ] use the default provider for landing page

- [ ] border-radius on the dapp screenshot should be 8px and the dark theme specific border style
image

- [ ] fix the border issue on desktop
image

- [ ] fix the border, grid and slideshow effect on mobile
image

constants/links.ts
```diff
@@ -43,6 +43,9 @@ export const EXTERNAL_LINKS = {
 		Optimism: 'https://optimistic.etherscan.io/tx',
 		OptimismKovan: 'https://kovan-optimistic.etherscan.io/tx',
 	},
+	Optimism: {
+		Home: 'https://www.optimism.io/',
```
koredefashokun, line 47: Just a nitpick, but I think we can remove the www:
```suggestion
		Home: 'https://optimism.io/',
```

pages/_app.tsx
```diff
@@ -67,7 +67,7 @@ const InnerApp: FC<AppProps> = ({ Component, pageProps }: AppPropsWithLayout) =>
 					<Layout>
 						<SystemStatus>{getLayout(<Component {...pageProps} />)}</SystemStatus>
 					</Layout>
-					<ReactQueryDevtools />
+					<ReactQueryDevtools position="top-left" />
```
koredefashokun, line 70: Do we want to leave this here permanently? I think that this is better for mobile, but is it okay on desktop?

queries/futures/useGetFuturesCumulativeStats.ts
```diff
@@ -9,12 +9,16 @@ import QUERY_KEYS from 'constants/queryKeys';
 import { FuturesCumulativeStats } from './types';
 import { getFuturesEndpoint } from './utils';
 import { wei } from '@synthetixio/wei';
+import { FUTURES_ENDPOINT_MAINNET } from './constants';
 
-const useGetFuturesCumulativeStats = (options?: UseQueryOptions<FuturesCumulativeStats | null>) => {
+const useGetFuturesCumulativeStats = (
+	homepage?: boolean,
```
koredefashokun, line 15: Just a nit, but maybe we can make a hook that matches the URL to check if we're on the homepage, so we can reuse it. Nothing wrong with this approach though, just think it might be useful for other things in the future as well. I see that you used a similar approach in the `useGetFuturesDailyTradeStats` hook and other hooks.

koredefashokun: On second thought, instead of manually checking if the user is on the homepage, we could just use the Infura endpoint when a user is on an unsupported network. We should be able to do that from within the `getFuturesEndpoint`, I think.

sections/homepage/Features/Features.tsx
```diff
@@ -181,58 +188,175 @@ const StyledFlexDivRow = styled(FlexDivRow)`
...
+		width: 305px;
+	`}
+`;
+
+const FeatureCard = styled(FlexDivRow)`
+	/* background: linear-gradient(180deg, rgba(40, 39, 39, 0.5) 0%, rgba(25, 24, 24, 0.5) 100%); */
+	/* box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.25), inset 0px 1px 0px rgba(255, 255, 255, 0.1),
+	inset 0px 0px 20px rgba(255, 255, 255, 0.03); */
+	/* width: 403px; */
+	/* height: 135px; */
```
koredefashokun, line 203: Maybe we could delete this if we're not using them anymore.

sections/homepage/Features/Features.tsx
```diff
@@ -181,58 +188,175 @@ const StyledFlexDivRow = styled(FlexDivRow)`
...
 	}
 `;
 
+const Container = styled.div`
+	margin: 140px 0px;
+`;
+
 const FeatureIconContainer = styled.div`
-	padding-bottom: 40px;
+	/* padding-bottom: 40px; */
```
koredefashokun, line 338: Seems like we might not need this anymore.

sections/shared/Layout/AppLayout/Header/MobileUserMenu/MobileMenuModal.tsx
```diff
@@ -11,20 +11,21 @@ import FullScreenModal from 'components/FullScreenModal';
...
+	homepage?: boolean;
 	onDismiss(): void;
 };
 
-export const MobileMenuModal: FC<MobileMenuModalProps> = ({ onDismiss }) => {
+export const MobileMenuModal: FC<MobileMenuModalProps> = ({ homepage, onDismiss }) => {
 	const { t } = useTranslation();
 	const { asPath } = useRouter();
-	const menuLinks = useRecoilValue(menuLinksState);
+	const menuLinks = useRecoilValue(homepage ? homepageMenuLinksState : menuLinksState);
```
koredefashokun, line 28: Not an issue (especially since it didn't start in this PR), but was wondering if there was a specific reason we store links in state, if they're constant.

sections/shared/Layout/HomeLayout/Header.tsx
```diff
@@ -120,28 +120,43 @@ const Header: FC = () => {
 					</Links>
 					<MenuContainer>
 						<Link href={ROUTES.Markets.Home}>
-							<Button variant="primary" isRounded={false} size="sm">
+							<Button noOutline={true} size="sm">
```
koredefashokun, line 123: We can save a few keystrokes like this. Looks like there are other instances in this file where we could also do this.
```suggestion
							<Button noOutline size="sm">
```

koredefashokun (commented):
Left a couple of comments on the code. Nothing major.

sections/homepage/Features/Features.tsx
LeifuChen: Good catch. Yashar re-wrote the css here so I didn't notice these comments. I will remove them.

sections/homepage/Features/Features.tsx
LeifuChen: Yes. Removed.

sections/shared/Layout/HomeLayout/Header.tsx
LeifuChen: Thanks for the suggestions. I will clean all of them.

constants/links.ts
LeifuChen: Updated.

sections/shared/Layout/AppLayout/Header/MobileUserMenu/MobileMenuModal.tsx
LeifuChen: I don't think so. Let's remove it.

koredefashokun (dismissed):

Issue #975, opened 2022-06-09
Exchange: Max Button Rounding Issue


platschi:
### Subject of the issue
On the Exchange, clicking `Max` should show 18 decimals otherwise it rounds up values and results in "insufficient balance" error messages.

See screenshot where max results to 35.7530, while the wallet has only 35.75296 sUSD, resulting in "insufficient balance".

![Screen Shot 2022-06-09 at 20 44 13](https://user-images.githubusercontent.com/548702/172962876-1050fb64-a96b-40bc-a186-491a261c449e.png)

PR #977, opened 2022-06-10
To truncate the number without rounding

LeifuChen:
## Description
On the Exchange, clicking `Max` and `Balance` button will round the number automatically, which is not expected.

## Related issue
#975 

## Motivation and Context
To improve the UX.

## How Has This Been Tested?
1. Click on `Max` and `Balance` to check the number is not rounding.

In this case, the actual `max` number is `10.728085525011272`. There is no rounding on display.
image

### DIFF
--- a/utils/formatters/number.ts
+++ b/utils/formatters/number.ts
@@ -33,6 +33,14 @@ export const getDecimalPlaces = (value: WeiSource) => (value.toString().split('.
 
 export const zeroBN = wei(0);
 
+export const truncateNumbers = (value: WeiSource, maxDecimalDigits: number) => {
+	if (value.toString().includes('.')) {
+		const parts = value.toString().split('.');
+		return parts[0] + '.' + parts[1].slice(0, maxDecimalDigits);
+	}
+	return value.toString();
+};
+
 /**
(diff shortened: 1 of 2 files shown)


platschi (approved):
Thanks, tested and this solves the issue!

utils/formatters/number.ts
```diff
@@ -33,6 +33,14 @@ export const getDecimalPlaces = (value: WeiSource) => (value.toString().split('.
 
 export const zeroBN = wei(0);
 
+export const truncateNumbers = (value: WeiSource, maxDecimalDigits: number) => {
+	if (value.toString().includes('.')) {
+		const parts = value.toString().split('.');
+		return parts[0] + '.' + parts[1].slice(0, maxDecimalDigits);
+	}
+	return value.toString();
+};
```
Tburm, line 42: Couldn't we do this with math instead of relying on string matching for the decimal? I'm wondering if this will break if someone uses `,` for the decimal point?

Something like this:
```
const truncValue = Math.floor(value.mul(10**maxDecimalDigits).toNumber()) / (10**maxDecimalDigits)
return truncValue.toString();
```

Tburm (approved):

Issue #954, opened 2022-06-06
Trades table resets page


Tburm:
When the "Trades" table on the "Futures" page receives new data it resets to the first page.

Reproduce:
* Go to "Futures" and select a market where you have >5 trades
* Go to page 2
* Wait 15 seconds, page will reset to first page

Tasks:
- [ ] Trades table does not reset to page 1 when data refreshes

PR #1017, opened 2022-06-15
Trades Table Reset

Tburm:
Fixing a bug where the trades table resets when the data refreshes.

## Description
Changing a default setting on the table component to not reset page on data refresh. This will apply to all tables.

## Related issue
- #954


platschi (approved):
Great fix, works now as expected

Issue #912, opened 2022-05-29
sUSD widget balance not updating


platschi:
### Subject of the issue
When traders deposit or withdraw sUSD on a given PERP market, the sUSD widget in the header does not update once a tx goes through and the values changes. Additionally, the PERP market in question does not show up in the balance dropdown after depositing margin (see example screenshot where 100 sUSD were deposited into ETH-PERP, but the widget does not reflect this:)

<img width="691" alt="Screen Shot 2022-05-29 at 20 11 48" src="https://user-images.githubusercontent.com/548702/170894996-5e41b233-99bd-494a-a07e-df2cc09e8947.png">

### Deliverables
- When traders deposit/withdraw sUSD into a market, the sUSD wallet balance in the header should refresh / refetch data once the tx is confirmed / balance changed.

### Rewards
50 SNX

platschi:
Closed by #1092

PR #1030, opened 2022-06-17
Update wallet balances after withdraw/deposit event

kkpsiren:
## Description

This PR Allows the sUSD and market balances to be automatically updated after executing transaction. Essentially. This PR adds a refetch query for the futuresPosition if  'margin-change' handleRefetch gets triggered.

## Related issue

https://github.com/Kwenta/kwenta/issues/912

## Motivation and Context

See ticket above. 

## How Has This Been Tested?

Did multiple deposits and withdraws on several markets using OE KOVAN.

### DIFF
--- a/contexts/RefetchContext.tsx
+++ b/contexts/RefetchContext.tsx
@@ -8,6 +8,11 @@ import { walletAddressState } from 'store/wallet';
 import useGetFuturesMarket from 'queries/futures/useGetFuturesMarket';
 import useGetFuturesPotentialTradeDetails from 'queries/futures/useGetFuturesPotentialTradeDetails';
 
+import useGetFuturesPositionForMarkets from 'queries/futures/useGetFuturesPositionForMarkets';
+import { getMarketKey } from 'utils/futures';
+import Connector from 'containers/Connector';
+import useGetFuturesMarkets from 'queries/futures/useGetFuturesMarkets';
+
 type RefetchType = 'modify-position' | 'new-order' | 'close-position' | 'margin-change';
 
 type RefetchContextType = {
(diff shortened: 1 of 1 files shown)


platschi:
Works as expected, deposited sUSD into a (OP Kovan) market and the sUSD widget value updated as soon as tx went through. Same for withdrawal.

Before approving, @Tburm we should check if the console errors are related to this PR or just vercel stuff:

Screen Shot 2022-06-21 at 16 20 48

contexts/RefetchContext.tsx
```diff
@@ -28,6 +33,13 @@ export const RefetchProvider: React.FC = ({ children }) => {
 	const marketQuery = useGetFuturesMarket();
 	useGetFuturesPotentialTradeDetails();
 
+	const { network } = Connector.useContainer();
+	const futuresMarketsQuery = useGetFuturesMarkets();
```
Tburm, line 37: We are updating these hooks to set data in recoil. Here is an example of how they differ ([setting value](https://github.com/Kwenta/kwenta/blob/fd444c2d24a628ccf50dd04ed04cd4133147108f/queries/futures/useGetFuturesMarket.ts#L81) and [recoil value](https://github.com/Kwenta/kwenta/blob/fd444c2d24a628ccf50dd04ed04cd4133147108f/queries/futures/useGetFuturesMarket.ts#L22))

For the new hooks here (`useGetFuturesMarkets` and `useGetFuturesPositionForMarkets`) we would want these set up in a similar way.

Tburm (commented):
This works as intended, but if we are adding more hooks to the `RefetchContext` we want them to follow the pattern set by other hooks. This means adding a value in recoil for the result of these queries, and setting it on running the hook.

Reach out to me to start a thread if you have questions.

Issue #951, opened 2022-06-06
Integrate 1inch to Exchange


platschi:
- Add 1inch integration / API to Synth Exchange
- 1inch API should be additive

- "Get more sUSD" widget currently uses Uniswap widget; remove this and forward the button to our exchange once 1inch is integrated

Issue #491, opened 2022-03-22
Error Improvements


platschi:
### Subject of the issue
Currently, errors are displayed with a default message below the Open Position button. While technical, they could be mapped to something more readable for the end user, including a suggestion on how to solve the issue.

Example:

The text could be changed to something as simple as `Max Leverage Exceeded. Please lower your leverage`.

### Deliverables
- [ ] Design how to display error messages
- [ ] Map default error messages to something readable
- [ ] Consider adding suggestions for solutions to mapped error messages

asherism:
https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/?node-id=2042%3A13938

<img width="591" alt="image" src="https://user-images.githubusercontent.com/11524787/161815773-f8405139-5582-49a5-aa4a-ab8f583c1231.png">


we need to come up with an exhaustive list of errors and map them to something human-readable. additionally, we should also display some helpful copy on how the errors can be fixed.

asherism:
| error        | ui message                             | resolution                        |
|--------------|----------------------------------------|-----------------------------------|
| max leverage | your position exceeds maximum leverage | decrease leverage                 |
| no margin    | insufficient margin                    | deposit more susd into the market |
| catch-all    | invalid position                       |                                   |

asherism:
```
_errorMessages[uint8(Status.InvalidPrice)] = "Invalid price";
_errorMessages[uint8(Status.PriceOutOfBounds)] = "Price out of acceptable range";
_errorMessages[uint8(Status.CanLiquidate)] = "Position can be liquidated";
_errorMessages[uint8(Status.CannotLiquidate)] = "Position cannot be liquidated";
_errorMessages[uint8(Status.MaxMarketSizeExceeded)] = "Max market size exceeded";
_errorMessages[uint8(Status.MaxLeverageExceeded)] = "Max leverage exceeded";
_errorMessages[uint8(Status.InsufficientMargin)] = "Insufficient margin";
_errorMessages[uint8(Status.NotPermitted)] = "Not permitted by this address";
_errorMessages[uint8(Status.NilOrder)] = "Cannot submit empty order";
_errorMessages[uint8(Status.NoPositionOpen)] = "No position open";
_errorMessages[uint8(Status.PriceTooVolatile)] = "Price too volatile";
```

asherism:
`"previous order exists"
"no previous order"
"cannot be cancelled by keeper yet"
"target roundId not reached"
[...]

PR #1038, opened 2022-06-21
Integrate Synthswap and 1inch swaps

avclarke:
## Description
Enable users to swap synth to erc20 via synthswap and other tokens via 1inch 

## Related issue
https://github.com/Kwenta/kwenta/issues/951


platschi:
On a first glance looks great! Will add more comments while testing.

1) Search bar only searches synths, not tokens list. We should probably rename it to "Search" and make sure it searches all lists.

Screen Shot 2022-06-21 at 22 11 39

platschi:
From #951, missing (still opens Uniswap widget):

> "Get more sUSD" widget currently uses Uniswap widget; remove this and forward the button to our exchange once 1inch is integrated

platschi:
3) Synth icon is not pulled when trying to swap on vercel deploy. They are however visible for tokens, but not for other synths. See screenshot.

Screen Shot 2022-06-21 at 23 01 19

4) "Exchange Fee" and "Fee Cost" are not visible when swapping from/to tokens. I guess if no (protocol) fees apply, we should just leave the label text and show a placeholder `-` for the values, to keep the UI consistent?

avclarke:
Thanks for the review.

1. Good catch will fix the search
2. ok
3.1. I'm not seeing the issue with synth icons on vercel deploy, seem to be showing up ok for me
3.2. Will look into gas fee issue
4. There will be some fees here but it needs some refactoring on the fee calculation in useExchange to support synthswap. I was thinking we could do that in a separate task. For now I'll make it consistent.

avclarke:
@platschi I do see the issue with icon sometimes actually, will also look into that

avclarke:
@platschi I've addressed all comments apart from the fees. I've left it as is, to hide fees on non synth to synth swap as this was the existing logic. As mentioned previously we'll need to refactor a bit to populate fees section for synthswap, I'm happy to add to this PR but as I will be off from tomorrow to Tuesday I don't think it will be a blocker for merging this feature if we wanted to break fees into a separate task.

platschi:
Sometimes I see this yellow `-100%` indicator but not sure what's that. Seems like the swap values are fine

Screen Shot 2022-06-22 at 17 02 30

avclarke:
> Sometimes I see this yellow `-100%` indicator but not sure what's that. Seems like the swap values are fine
> 
> Screen Shot 2022-06-22 at 17 02 30

@platschi this is the slippage percent, and it was showing -100% because coingecko price fetching was hardcoded to ethereum and failing to fetch prices of optimism addresses. Should be fixed now as updated the query to include the correct network.

platschi:
- When selecting e.g. sUSD -> AELIN, I cannot type in values in the "Into" input field.  However, this works (as expected) if I switch to sUSD -> sLINK, for example. Would expect to be able to input values in both fields.

- sUSD -> ELK does not calculate a value for ELK
- sUSD -> WBTC, sUSD -> USDT, sUSD -> USDC values are extremely low  and/or with a 100% slippage, probably no liquidity here, do we still want to offer these tokens or hide them from the list given it's basically useless to swap them?

avclarke:
>When selecting e.g. sUSD -> AELIN, I cannot type in values in the "Into" input field. However, this works (as expected) if I switch to sUSD -> sLINK, for example. Would expect to be able to input values in both fields.

I forgot to mention I disabled input on "into" field for synthswap assets as it's going to be a bit more effort to make the conversion the other way. Could we break into another ticket?

> sUSD -> ELK does not calculate a value for ELK
I just checked and this gives an insufficient liquidity error. Do we have a standard error notification or something we can show for things like this?

>sUSD -> WBTC, sUSD -> USDT, sUSD -> USDC values are extremely low and/or with a 100% slippage, probably no liquidity here, do we still want to offer these tokens or hide them from the list given it's basically useless to swap them?
This is a good catch and actually because I'm not using the correct asset decimals

>Showing Slippage is cool, but we should maybe make it clear that the yellow % value refers to Slippage. It wasn't immediately clear to me when I saw it first :) Maybe add a "Slippage: " string in front of them, or something similar that fits
Agree this is not very obvious, it was just in the existing UI. I will see how it looks with "slippage" label.

platschi:
> > When selecting e.g. sUSD -> AELIN, I cannot type in values in the "Into" input field. However, this works (as expected) if I switch to sUSD -> sLINK, for example. Would expect to be able to input values in both fields.
> 
> I forgot to mention I disabled input on "into" field for synthswap assets as it's going to be a bit more effort to make the conversion the other way. Could we break into another ticket?

All right will create another ticket to fix it.

> > sUSD -> ELK does not calculate a value for ELK
> 
> I just checked and this gives an insufficient liquidity error. Do we have a standard error notification or something we can show for things like this?

Currently just showing the error warning within the "Swap" button, you can test by forcing an insufficient balance error. We can place it there for now until we have a better error handling in place (part of #1052, #491)

> 
> > sUSD -> WBTC, sUSD -> USDT, sUSD -> USDC values are extremely low and/or with a 100% slippage, probably no liquidity here, do we still want to offer these tokens or hide them from the list given it's basically useless to swap them?
> 
> This is a good catch and actually because I'm not using the correct asset decimals

:+1:

> > Showing Slippage is cool, but we should maybe make it clear that the yellow % value refers to Slippage. It wasn't immediately clear to me when I saw it first :) Maybe add a "Slippage: " string in front of them, or something similar that fits
> 
> Agree this is not very obvious, it was just in the existing UI. I will see how it looks with "slippage" label.

Thanks!

containers/Convert/Convert.tsx
```diff
@@ -3,16 +3,19 @@ import axios from 'axios';
 import { ethers } from 'ethers';
 import { useRecoilValue } from 'recoil';
 import { useCallback } from 'react';
-
-import erc20Abi from 'lib/abis/ERC20.json';
+//@ts-ignore
+import getFormatedSwapData from '@jaredborders/synthswap-utils';
```
Tburm, line 7: Should we find a way to publish this from Kwenta? @JaredBorders

queries/coingecko/useCoinGeckoTokenPricesQuery.ts
```diff
@@ -1,24 +1,29 @@
...
+		QUERY_KEYS.CoinGecko.TokenPrices(tokenAddresses, platform),
 		async () => {
 			const response = await axios.get<PriceResponse>(
-				`${CG_BASE_API_URL}/simple/token_price/ethereum?contract_addresses=${tokenAddresses.join(
+				`${CG_BASE_API_URL}/simple/token_price/${platform}?contract_addresses=${tokenAddresses.join(
 					','
 				)}&vs_currencies=usd`
 			);
-
+			console.log('response', response);
```
Tburm, line 26: Can remove

sections/shared/modals/SelectCurrencyModal/SelectCurrencyModal.tsx
```diff
@@ -106,88 +111,196 @@ export const SelectCurrencyModal: FC<SelectCurrencyModalProps> = ({
...
+						<EmptyDisplay>{t('modals.select-currency.search.empty-results')}</EmptyDisplay>
+					)}
+					{oneInchTokensPaged.length ? (
+						<>
+							<TokensHeader>
+								<span>
+									{assetSearch ? (
+										<span>{t('modals.select-currency.header.search-results')}</span>
+									) : synthCategory != null ? (
+										t('modals.select-currency.header.category-synths', {
```
Tburm, line 255: This will display "Crypto Synths" or "Forex Synths" above the 1inch tokens when there is a filter selected

image

Tburm (commented):
Swapping from ETH -> sUSD gave me an error. Maybe we need to handle the ETH case differently than other ERC20s?

image

Tburm:
This title also looks odd for non-sUSD pairs...

image

For reference, an sUSD pair:
image

containers/Convert/Convert.tsx
JaredBorders: We can. Will put on todo list for tmr

avclarke: Also could be good to add the types or convert the code to ts, maybe a nice devdao task

sections/shared/modals/SelectCurrencyModal/SelectCurrencyModal.tsx
avclarke: Good spot, I think I'll hide 1inch tokens when forex is selected and then update the label for when crypto is selected

avclarke:
> This title also looks odd for non-sUSD pairs...
> 
> image
> 
> For reference, an sUSD pair: image

I was just looking into this and it seems an existing issue in `formatCryptoCurrency`, it's adding currency key as prefix and sign as suffix. I will switch them unless there is a reason for this that I am missing.

Screenshot 2022-07-01 at 12 32 48

Tburm (dismissed):

containers/Convert/Convert.tsx
JaredBorders: Latest version of Synthswap package no longer uses that

JaredBorders: 1.0.3

Tburm (approved):

PR #1042, opened 2022-06-22
Feat/lazy load

Tburm:
New PR for lazy loading of the transaction history table


platschi (approved):
Tested on testnet and mainnet, works as expected!

Issue #1043, opened 2022-06-22
Fix empty state for History tables


platschi:
### Subject of the issue
When a wallet is connected to L2 but hasn't done any trades yet, the History tables for both Futures History and Spot history should show an empty table with a CTA message, similar to other tables on Kwenta. Currently, there is no such thing and the table looks broken.

### How it is currently
<img width="1064" alt="Screen Shot 2022-06-22 at 18 45 49" src="https://user-images.githubusercontent.com/548702/175093251-b80ca679-3392-44de-a7ec-a7d39c4042c1.png">

### Example of how it should be
<img width="952" alt="Screen Shot 2022-06-22 at 18 54 29" src="https://user-images.githubusercontent.com/548702/175094564-3912d02b-4751-4862-b54a-7c7b11298685.png">

### Deliverables
When a wallet is connected to Optimism but has no trades:

- [ ] On Futures History Tab: "You have no trades yet. Trade Perpetual Futures on Kwenta now!" 

The second sentence should be below the first, being a link pointing to `https://kwenta.io/markets/sETH`. See for example screenshot above.

- [ ] When user has no trades on Spot History Tab: "You have no trades yet. Visit the Kwenta exchange to swap synths."   

The second sentence should be below the first, being a link pointing to `https://kwenta.io/exchange`. See example above

Note: Make sure to use same layout as in second example image above.

### Rewards
25 SNX

### Deadline:
28 June 2022

PR #1050, opened 2022-06-24
Added text into Futures History Table and Spot History Table.

sebfk:
Fix to ticket #1043 

## Description
Added text into the tables in Future History Table and Spot History Table, if no trades have been made.

## How Has This Been Tested?
Ran it in local enviorment, and tested it with both logged into L2 and not logged into L2. Also tried doing some testnet transaction and tested if the text would then disappear.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,7 @@
 		],
 		"jsx-a11y/anchor-is-valid": "off",
 		"jsx-a11y/alt-text": "off",
-		"import/no-anonymous-default-export": "off"
+		"import/no-anonymous-default-export": "off",
+		"@next/next/no-img-element": "off"
 	}
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,11 +3,7 @@
 	"plugins": ["prettier"],
 	"rules": {
 		"react/react-in-jsx-scope": "off",
-		"prettier/prettier": "error", [
-		{
-		"endOfLine: "auto"
-		}
-		],
+		"prettier/prettier": "error",
 		"no-mixed-spaces-and-tabs": ["warn", "smart-tabs"],
 		"comma-dangle": [
 			"error",
@@ -21,7 +17,6 @@
 		],
 		"jsx-a11y/anchor-is-valid": "off",
 		"jsx-a11y/alt-text": "off",
-		"import/no-anonymous-default-export": "off",
-		"@next/next/no-img-element": "off"
+		"import/no-anonymous-default-export": "off"
 	}
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,11 @@
 	"plugins": ["prettier"],
 	"rules": {
 		"react/react-in-jsx-scope": "off",
-		"prettier/prettier": "error",
+		"prettier/prettier": "error", [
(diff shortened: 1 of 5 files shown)


.prettierrc
```diff
@@ -3,5 +3,6 @@
   "printWidth": 100,
   "semi": true,
   "singleQuote": true,
-  "trailingComma": "es5"
+  "trailingComma": "es5",
+  "endOfLine": "crlf"
```
koredefashokun, line 7: I think we should leave this as the default (`lf`). Is there a specific reason you had to change it to `crlf`?

.eslintrc
```diff
@@ -3,7 +3,11 @@
 	"plugins": ["prettier"],
 	"rules": {
 		"react/react-in-jsx-scope": "off",
-		"prettier/prettier": "error",
+		"prettier/prettier": "error", [
+		{
+		"endOfLine: "auto"
+		}
+		],
```
koredefashokun, line 10: Just wanted to check if there was a reason for changing the `endOfLine` config.

.prettierrc
LeifuChen: @koredefashokun the dev is using windows so it requires adding this to make the lint work properly. however, there is no need to commit to the PR. I already asked the dev to reset this file.

.eslintrc
LeifuChen: @koredefashokun same reason. it's the setting for the dev who is working with windows but there is no need to commit to the PR.  @sebfk please reset this file in your commit.

LeifuChen (changes requested):
@sebfk Don't include `.eslintrc` and `.prettierrc` in your commit. Keep them locally.

sections/dashboard/SpotHistoryTable/SpotHistoryTable.tsx
```diff
@@ -251,6 +262,22 @@ const StyledText = styled.div`
...
+	justify-items: center;
+	grid-gap: 10px;
+	color: ${(props) => props.theme.colors.selectedTheme.button.text};
+	font-size: 20px;
+	font-family: ${(props) => props.theme.fonts.bold};
+	div {
+		text-decoration: underline;
+		font-size: 16px;
+		font-family: ${(props) => props.theme.fonts.regular};
+	}
```
LeifuChen, line 278: Please add `cursor: pointer;` into `div {...}`.
Same css change for the `FutureHistoryTable.tsx` (it's not your problem but it's good that you can fix it in this PR).

LeifuChen (changes requested):
@sebfk One css change is requested in both the future history table and the spot history table.

translations/en.json
```diff
@@ -538,7 +538,9 @@
 				"date-time": "Date/Time",
 				"from": "From",
 				"to": "To",
-				"usd-value": "USD Value"
+				"usd-value": "USD Value",
+				"no-trade-history": "You have no trades yet",
```
platschi, line 542: Can you add a dot to the end of this sentence, so it's in line with the other one? :)

LeifuChen (approved):
@sebfk  Thanks for the contribution. Looks good to me.

PR #1060, opened 2022-06-29
Fix the market tab on the smaller screen

LeifuChen:
## Description
The text in the table might be cut out on a smaller screen.
image

## Motivation and Context
Make the text in table adapt the smaller screen.

## How Has This Been Tested?
1. Switch to responsive mode and change the size of the container. No cut out of the texts are observed.

image
image


koredefashokun (approved):
The changes look good.

PR #1064, opened 2022-06-30
Fix Liquidation Pnl

Tburm:
We calculate pnl net of fees on the frontend. New subgraph changes implemented the `pnlWithFeesPaid` value directly to avoid calculations on the frontend.

## Description
* Add `pnlWithFeesPaid` to the position types and queries
* Implement the `pnlWithFeesPaid` value for realized pnl and pnl % values

## Related issue
* [Issue](https://github.com/Kwenta/kwenta-subgraph/issues/58)


avclarke (approved):
Looks good!

PR #1070, opened 2022-07-01
Adjust max leverage for NP orders

### LINKED ISSUE #1012: Next Price:  Do not allow NP orders at maxLeverage value
### Subject of the issue
Users often open next price orders at 10x leverage, which can result in excess of maxLeverage at the next price update, which results in Keepers dropping the order.

This results in users loosing money. We have a warning text in place if the leverage value is within 1x of the maxLeverage, however the message either seems unclear or is often ignored or misunderstood.

To avoid this, we should limit the maximum allowed leverage for next price orders to a threshold n below maxLeverage, e.g. max 9x for a market with 10x maxLeverage.

### Deliverables
- Investigate maximum allowable leverage value that makes most sense for next price orders
- Implement limitation for next price orders

Tburm:
Adds an adjustment to max leverage amount when the user is submitting a next price order.

## Description
* Adds constant for leverage adjustment (99.75% of current leverage)
* Adjusts current max leverage on NP orders by the constant

## Related issue
- [ ] closes #1012 

## Motivation and Context
Trader next price orders fail when the specified position size and next oracle update push them past max leverage. This change will adjust the available max leverage by 99.75% since oracle updates should be <.2%.

### DIFF
--- a/constants/defaults.ts
+++ b/constants/defaults.ts
@@ -35,3 +35,6 @@ export const DEFAULT_SLIPPAGE = 1;
 
 // for Trading History
 export const DEFAULT_NUMBER_OF_TRADES: number = 50;
+
+// leverage adjustment
+export const DEFAULT_NP_LEVERAGE_ADJUSTMENT: number = 0.9975;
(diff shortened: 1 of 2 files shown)


platschi:
@Tburm Tested this on mainnet with a 50$ NP Long on UNI-PERP with maxLeverage (9.97). For some reason the Keeper didn't pick it up, had to execute manually (successfully). Not sure this is related to this ticket or it would need more than the 50$ to pick an order?

Market was calm, no volatility nor other fees at the moment of submit. See last two txs at https://optimistic.etherscan.io/address/0xc2ecd777d06ffdf8b3179286beabf52b67e9d991

LeifuChen (approved):
The solution looks good to me.

Issue #1002, opened 2022-06-13
Improve leverage slider


Tburm:
## General
It still is possible to push the leverage slider past 10x, which is causing `maxLeverage` reached errors. This might be an issue due to rounding the `maxLeverage` with `toNumber()`, making us lose precision. A possible solution might be to  set the max slider value to `maxLeverage - 0.1`

## On Desktop
Leverage slider is not visible on desktop (see screenshot). This should be fixed:

<img width="391" alt="Screen Shot 2022-06-14 at 09 57 39" src="https://user-images.githubusercontent.com/548702/173582723-281a82e3-8736-48bf-9c63-f9cbec05a90b.png">

## On Mobile
The current leverage slider shows the value below the slider. On mobile this would be below the finger used to move the slider.

- [x] Move the leverage value above the leverage slider on mobile

PR #1071, opened 2022-07-01
Fix the missing slider trail

LeifuChen:
## Description
Fix the following issues:
1. Leverage slider is not visible on the desktop.
2. Move the leverage value above the leverage slider on mobile.
3. Fix the displayed value of the slider label
4. When the user reaches the maximum leverage, the `up to xx` and max leverage label will show the negative value (in this case, the slider is disabled). The original solution is trying to use the `10x` as a placeholder but it's not working as expected because it only uses the placeholder value when the value is null (in this case, it's a negative value). I changed the logic a little bit to use the `10x` as the placeholder when the slider is disabled.

## Related issue
#1002 

## Motivation and Context
Improved the UI

## How Has This Been Tested?
1. Open the market page on the desktop and see the slider rail
2. Open the market page on the mobile and see the leverage value above the leverage slider

image
image
image

### DIFF
--- a/components/Slider/Slider.tsx
+++ b/components/Slider/Slider.tsx
@@ -53,7 +53,7 @@ const styledMarkLabel = css`
 const SliderContainer = styled.div`
 	width: 100vw;
 	height: 24px;
-	padding: 0 6px;
+	padding: 0px 12px 0px 4px;
 	box-sizing: border-box;
 	${media.lessThan('sm')`
 		padding: 0 4px 0 4px;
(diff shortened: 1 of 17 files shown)


platschi:
Slider works as expected, however the slider "slides" over the slider rail when setting maximum 10x which should be fixed. Looks a bit off with the 10x value being active when the slider slides past the label. See below what I mean:

https://user-images.githubusercontent.com/548702/177357421-e6804a58-a884-4d64-bd7a-f00998ce4b1f.mov

Tburm:
@platschi I might disagree with #0. The reason I requested the number above the slider on mobile was because your thumb covers the number while you adjust it. If we want them the same for _consistency_ it makes sense, but this is the reason why they are different.

platschi (approved):

Tburm (approved):

PR #1095, opened 2022-07-07
Error fixes

Tburm:
Hotfixing some errors that are displaying in the live version of the app right now.

## Description
* Fixes an error resulting from a missing connector to synthetix contracts
* Fixes errors resulting from missing data in the skew bar
* Fixes errors resulting from missing wallet address when not connected


LeifuChen:
Here is a note for myself. 

In the original implementation, `data` in `SkewInfo.tsx` could be undefined if the `currencyKey` doesn't find a match in `futuresMarkets`.

Issue #1097, opened 2022-07-08
Bug: Trade history empty state


Tburm:
Empty state for spot and futures history is displayed even when a trader has trades.

- [ ] Empty state only shows when the account has no spot trades
- [ ] Empty state only shows when the account has no futures trades

Example:
![Image](https://user-images.githubusercontent.com/10401554/178024841-9a76e85e-597f-4c6c-a669-7008c03a5230.png)

PR #1098, opened 2022-07-08
Fix "no result" on Table component

Tburm:
Fixes a bug in the `Table` component that prevents displaying the "no result" message. Additionally, there are fixes for the spot and futures history tables to improve how and when we display these messages.

## Description
* Adds loading spinner to spot/futures history tables
* Adds check for length of data on "No result" logic in the table component
* Adds a `pageSize` parameter to trigger pagination on the history tables
* Fix an unrelated bug in the `CurrencyPrice` component that results in throwing errors when the `price` value doesn't exist

## Related issue
- #1097 

## Motivation and Context
Bug fixes

## How Has This Been Tested?
* Tested display of tables on accounts that have history, and ones that do not have history

No pagination:
image

Pagination:
image

No result:
image


LeifuChen (approved):
@Tburm Good job! Looks good to me.

Issue #1089, opened 2022-07-06
Replace default node provider


platschi:
### Subject of the issue
Replace the current default infura provider with the one we have from blast, a decentralized node provider. Contact @JChiaramonte7 for details.

### Deliverables
- Replace default node provider (infura) with blast
- tbd

PR #1115, opened 2022-07-12
Update default provider to BlastAPI

Tburm:
Changes default provider from Infura to BlastAPI, a decentralized node provider.

## Description
* Adds a new environment variable for Blast API access: `NEXT_PUBLIC_BLASTAPI_PROJECT_ID`
* Updates default provider from Infura to BlastAPI

## Related issue
- #1089 

## Motivation and Context
Infura is a centralized node provider. This change moves Kwenta forward in decentralizing the frontend, even while users do not have their wallets connected.

## How Has This Been Tested?
Tested app with wallet disconnected. Data on the dashboard and futures markets all populates.


utils/network.ts
```diff
@@ -42,6 +45,40 @@ export type GasInfo = {
...
+	}
+
+	return loadProvider({
+		networkId,
+		infuraId: process.env.NEXT_PUBLIC_INFURA_PROJECT_ID,
+	});
+};
+
+const loadBlastProvider = (networkId: NetworkId) => {
+	const networkSlug = BLAST_NETWORK_LOOKUP[networkId];
```
avclarke, line 60: Should we add the same check as above on the env param existing and throw when missing?

avclarke (approved):
Looks good, just a small suggestion if you think it's worth adding.

PR #1122, opened 2022-07-14
Fix ETH swaps failing with 1inch

avclarke:
Some 1inch swaps which include ETH sometimes fail when slippage is only 1%. Raising to 3% seems to be the minimum to ensure these swaps don't error.


Tburm (approved):
:+1:

PR #1128, opened 2022-07-15
Hotfix: L1 atomic swaps

Tburm:
L1 atomic swaps are failing due to an error thrown by `useSynthetixTxn` with our gas object. The transaction estimates gas itself, so we can remove our gas specification completely and the transaction will succeed.

[Test txn](https://etherscan.io/tx/0x46dda8b75fbad85c0609de3ebabb81991d300e6dc6034d9c2847e04ebf66f1cc)


LeifuChen (approved):
Good job! Work like a charm.

PR #1140, opened 2022-07-19
Re-add transaction notifier for useExchange txns

koredefashokun:
## Description
This PR makes the transaction notifier show up for exchange-related transactions.

## Related issue
N/A

## Motivation and Context
N/A

## How Has This Been Tested?
N/A


sections/exchange/hooks/useExchange.ts
```diff
@@ -602,6 +601,30 @@ const useExchange = ({
 		}
 	);
 
+	useEffect(() => {
+		if (exchangeTxn.hash) {
+			monitorTransaction({
```
avclarke, line 606: Could be nice to extract this to a separate function to reduce duplicated code, it can then be called from both here and from the other submit tx.

avclarke (commented):
Looks good, just one place we can remove some code duplication.

sections/exchange/hooks/useExchange.ts
koredefashokun: Yeah. Saw that, but was too lazy to change it. Will fix now.

avclarke: I was thinking we could do the whole thing couldn't we, like a function called `monitorSubmitTx(hash)` as there's still some unnecessary duplication?

avclarke (approved):
Looks good!

Issue #1081, opened 2022-07-04
Improve Image handling


platschi:
### Subject of the issue
Improve image handling at different spots within the Kwenta codebase.

### Relevance
Part of #875 

### Deliverables

- [x] Remove stale pngs and svgs
- [x] Add SNX assets repo images locally and compress/optimize them to reduce external requests / repo size
- [x] Replace` <img>` tag with `Image` everywhere to improve performance
- [x] Remove dead/unused code

PR #1149, opened 2022-07-20
Clean up the image assets

LeifuChen:
## Description
Improve image handling at different spots within the Kwenta codebase.

## Related issue
#1081

## Motivation and Context
- [x] Remove stale pngs and svgs
- [x] Add SNX assets repo images locally and compress/optimize them to reduce external requests / repo size
- [x] Replace` ` tag with `Image` everywhere to improve performance
- [x] Remove dead/unused code


LeifuChen:
Note that we have to use `` tag to show ens avatar because we must add external URL to domains in `next.config.js`, which does not support wildcard.
image

components/Currency/CurrencyIcon/CurrencyIcon.tsx
```diff
@@ -79,11 +94,77 @@ const CurrencyIcon: FC<CurrencyIconProps> = ({ currencyKey, type, isDeprecated,
 
 	if (!firstFallbackError) {
 		switch (currencyKey) {
-			case CRYPTO_CURRENCY_MAP.ETH: {
-				return <ETHIcon {...props} />;
+			case FuturesMarketKey.sETH: {
```
avclarke, line 97: One other option here is we could create a mapping to get the image src and then just return the one Image tag like:
```
const synthIcons = {
  sETH: ETHIcon,
  sUNI: UNIIcon,
  ...
}

const src = synthIcons[currencyKey]
return src ? <Image src={src} layout="raw" {...props} /> : <TheDefaultImage />
```

avclarke (dismissed):
Looks good, just one suggestion but not a blocker.

components/Currency/CurrencyIcon/CurrencyIcon.tsx
LeifuChen: Good idea. Let me make a change.

LeifuChen: @avclarke Change has been made. Please have a re-review. Thanks.

Tburm (commented):
Looks good, just one issue in the market dropdown when there is no image found:
image

LeifuChen:
> Looks good, just one issue in the market dropdown when there is no image found: image

Thanks for the double-check. Yes. There is no AXS image inside `synthetix-assets` repo.
https://github.com/Synthetixio/synthetix-assets/tree/master/synths/png

The placeholder of AXS is same as prod.
image

Tburm:
@LeifuChen I was pointing out that the image is not sized properly when we use a placeholder image. I pushed a commit to inherit the props into the styled-component css instead of using a `style` tag and that fixed it.

Tburm (approved):

LeifuChen:
> @LeifuChen I was pointing out that the image is not sized properly when we use a placeholder image. I pushed a commit to inherit the props into the styled-component css instead of using a `style` tag and that fixed it.

Ok. I misunderstood your comments. Thanks for fixing it.

PR #1154, opened 2022-07-21
Refactor: Trade size bug

Tburm:
Fixes a division-by-zero error when inputting a trade size of zero.

## Description
* Fix trade size bug
* Add query keys to futures positions to fix margin updates on an account or network change


LeifuChen (approved):
Tested in Optimism Kovan. The code looks good to me.

Issue #1155, opened 2022-07-21
Filter current market from `FuturesPositionTable`


Tburm:
The `FuturesPositionTable` should filter the current market on the markets page.

![Image](https://user-images.githubusercontent.com/10401554/180318915-708b9d2a-0f18-404d-aeb8-397e30c84d1a.png)

Issue #1078, opened 2022-07-04
Refactor translation file


platschi:
### Subject of the issue
The translation file should be structured similarly to the navigation structure of the app, as opposed to similar text. This will make it easier to find copy, and also know where new copy should go.

### Relevance
Part of #875 

### Deliverables

- [x] Refactor translation file according to app navigation structure
- [x] Remove unused/obsolete translations
- [x] Add relevant information to Kwenta documentation

PR #1162, opened 2022-07-22
Bug: Futures positions table

Tburm:
Updates the futures positions table component to filter the current market if specified. This required other changes to state management and translations, including some opinionated copy changes.

## Description
* Add `showCurrentMarket` parameter to `FuturesPositionsTable`
* Remove all `useGetFuturesMarkets` hooks from components
* Use `futuresMarkets` recoil value where relevant
* Changes to copy on empty table states

## Related issue
* #1078 
* #1155 

## Motivation and Context
1. A bug causing the current market's position to show on the "markets" page
2. Some inconsistent copy between tables
3. Improved state management, reducing refetching significantly since the `useGetFuturesMarkets` hook will be called only once at the context level.

## How Has This Been Tested?
* Tested desktop table states with and without open positions
* Tested mobile table states with and without open positions

image

image

image

image

### DIFF
--- a/sections/dashboard/FuturesMarketsTable/FuturesMarketsTable.tsx
+++ b/sections/dashboard/FuturesMarketsTable/FuturesMarketsTable.tsx
@@ -35,17 +35,17 @@ const FuturesMarketsTable: FC = () => {
 
 	const futuresMarkets = useRecoilValue(futuresMarketsState);
 
-	const synthList = futuresMarkets?.map(({ asset }) => asset) ?? [];
+	const synthList = futuresMarkets.map(({ asset }) => asset);
 	const dailyPriceChangesQuery = useLaggedDailyPrice(synthList);
 
 	const futuresVolumeQuery = useGetFuturesTradingVolumeForAllMarkets();
 
 	const fundingRates = useGetAverageFundingRateForMarkets(
-		futuresMarkets?.map(({ asset, price, currentFundingRate }) => ({
+		futuresMarkets.map(({ asset, price, currentFundingRate }) => ({
 			currencyKey: asset,
 			assetPrice: price.toNumber(),
 			currentFundingRate: currentFundingRate.toNumber(),
-		})) ?? [],
+		})),
 		PERIOD_IN_SECONDS[Period.ONE_HOUR]
 	);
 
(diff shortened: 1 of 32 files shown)


sections/dashboard/FuturesMarketsTable/FuturesMarketsTable.tsx
```diff
@@ -27,62 +28,60 @@ import {
...
-}: FuturesMarketsTableProps) => {
+const FuturesMarketsTable: FC = () => {
 	const { t } = useTranslation();
 	const router = useRouter();
 	const { synthsMap } = Connector.useContainer();
 
-	const synthList = futuresMarkets.map(({ asset }) => asset);
+	const futuresMarkets = useRecoilValue(futuresMarketsState);
+
+	const synthList = futuresMarkets?.map(({ asset }) => asset) ?? [];
```
koredefashokun, line 38: Maybe we can make `futuresMarketState` default to `[]`, so we don't need to do the optional chaining every time we want to use it.

sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
```diff
@@ -332,6 +373,23 @@ const NoPositionsText = styled.div`
 	margin: 20px 0;
 	font-size: 16px;
 	text-align: center;
+	text-decoration: underline;
+`;
+
+const TableNoResults = styled(GridDivCenteredRow)`
```
koredefashokun, line 379: Not sure if we want to, but maybe we can put the styles for the "no results" state in the `Table` component itself, so we only have to pass the message to display, and the CTA text and link route (could be optional).

Tburm: Right now these are actually two different styles since there is a type with bigger padding, so this would be a good improvement

Tburm: I decided to add and export a style from the Table component, so we can import that inside components in case we want more custom styling.

koredefashokun (approved):

platschi (approved):
Works fine everywhere.

PR #1163, opened 2022-07-24
Fix: reject the txn may crash the app

### LINKED ISSUE #1165: Rejecting txs results in a crash
### Subject of the issue
Sending a transaction to confirm in MetaMask then rejecting it resulted in a crash.

Image

### How to reproduce

- Open a new Futures position
- From `TxConfirmationModal`, choose Confirm Order
- Reject in MetaMask. App crashes.

LeifuChen:
## Description
As @Tburm reported, the app may crash if the user rejects a future market transaction in the wallet.

## Related issue
closes #1165

## Motivation and Context
Improve the UX and stability of app

## How Has This Been Tested?
1. Open the future market page and open a position
2. From `TxConfirmationModal`, choose `Confirm Order`
3. The app is working normally after choosing `Reject` in wallet

### DIFF
--- a/components/Error/Error.tsx
+++ b/components/Error/Error.tsx
@@ -12,7 +12,7 @@ export const Error: FC<ErrorProps> = ({ message, formatter }) => {
 	const formattedMessage = useMemo(() => {
 		switch (formatter) {
 			case 'revert':
-				return formatRevert(message);
+				return formatRevert(message ?? '');
 			default:
 				return message;
 		}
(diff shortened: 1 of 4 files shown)


platschi (approved):
Bug has been fixed, app does not crash anymore upon rejecting txs. Tested with deposit/withdraw, open/modify, close orders on Futures as well as Exchange swaps on L1 and L2.

PR #1164, opened 2022-07-25
Fix tx error handling

avclarke:
- Adds a check if the tx error is user denied transaction and doesn't display the error view in that case or if there is no message.

- Cleans up error handling on manage position.

### DIFF
--- a/components/Error/Error.tsx
+++ b/components/Error/Error.tsx
@@ -1,7 +1,7 @@
 import React, { FC, useMemo } from 'react';
 import styled from 'styled-components';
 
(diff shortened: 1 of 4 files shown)


components/Error/Error.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { FC, useMemo } from 'react';
 import styled from 'styled-components';
 
-import { formatRevert } from 'utils/formatters/error';
+import { formatRevert, isMMUserDeniedError } from 'utils/formatters/error';
```
LeifuChen, line 4: Shall we change the name of this function to `isUserDeniedError`?

LeifuChen (approved):

PR #1175, opened 2022-07-25
Futures mobile fixes

koredefashokun:
## Description
This PR fixes a number of reported issues with futures on mobile.

## Related issue
N/A

## Motivation and Context
Conclusion of mobile compatibility development (until new features are added).

## How Has This Been Tested?
N/A


platschi:
Looks better now! A few comments:





- [ ] Top buttons on Markets (Price and Trades) shift layout when switching from either Account or Stats (see screen recording attached). Note #1184 adds more data to the Stats, so maybe merge that one first to avoid fixing the layout shift twice.
- [ ] Price button should be moved into the first place and load on default
- [ ] When scrolling down on Dashboard and selecting a market (e.g. try with the last one in the list), page jumps to the middle of the market page, not to the top (as would be expected)
- Still being discussed: Make Dashboard menu link clickable (link straight to Overview)
- Need to set up individual Markets page (similar to Desktop), as currently "Overview" and "Markets" links in navigation point to the same page (Dashboard). This will be tracked in a separate ticket.

https://user-images.githubusercontent.com/548702/181279515-a00654e0-410f-47aa-9c8c-4dec6e8b2fe9.MP4

platschi:
- [ ] Left-align "Oracle/Entry" column data in Open Position table to avoid shifts with higher amounts (see Discord channel comments):

### How it is
shift1


### How it should be
shift

platschi:
Looks great now, much better with the buttons not jumping around! Re: markets not jumping to top, we'll create a separate ticket for this.

My two last comments:

1) Skew values >1M create additional rows which makes the Stats table border overlap with the buttons. see screenshot:

skew


2) Open Position price box is now nicely aligned, however leverage values >=10.00x create an obsolete new line. Maybe expand the left column width a bit so it fits nicely. See screnshot:

Screen Shot 2022-08-01 at 15 08 53

platschi (dismissed):
Thanks, all changes work good now!

sections/futures/UserInfo/OpenOrdersTable.tsx
```diff
@@ -92,101 +95,145 @@ const OpenOrdersTable: React.FC = () => {
...
+									</MobilePositionSide>
+									<div>{cellProps.row.original.orderType}</div>
+								</div>
+							),
+							width: 100,
+						},
+						{
+							Header: <StyledTableHeader>Size</StyledTableHeader>,
+							accessor: 'size',
+							Cell: (cellProps: CellProps<any>) => <div>{cellProps.row.original.size}</div>,
```
Tburm, line 224: This size isn't formatted correctly on mobile:
image

sections/leaderboard/Leaderboard/Leaderboard.tsx
```diff
@@ -307,6 +307,8 @@ const Leaderboard: FC<LeaderboardProps> = ({ compact }: LeaderboardProps) => {
 							<StyledTable
 								compact={compact}
 								data={data}
+								showPagination
+								pageSize={20}
```
Tburm, line 311: Can we set this constant somewhere else?

sections/leaderboard/Leaderboard/Leaderboard.tsx
```diff
@@ -307,6 +307,8 @@ const Leaderboard: FC<LeaderboardProps> = ({ compact }: LeaderboardProps) => {
 							<StyledTable
```
Tburm, line 307: This table should have the same "loading" state as above so we avoid this state:
image

platschi:
From @Marko-roy: 

1) "Orders" Size amount on mobile remove asset name (should be the same as on desktop) to just show the Size value.
2) "Orders" tab table Type not aligned on mobile, same wit "Short" box, text should be centered and box padding aligned. See screenshot:

IMG_4097

Tburm (approved):

platschi (approved):

Issue #1006, opened 2022-06-14
Add Skew to mobile Futures Markets Stats


platschi:
### Subject of the issue
The initial mobile design didn't include any Skew information (Open Interest Long and Short values and %). We want to add this to the Market Stats on mobile.

### Deliverables
- Remove ETH-PERP price from mobile Market Stats box, rename External Price to Live Price
- Move data around to match order as shown in mockup (see below) 
- Add Skew information to Market Stats on Mobile (see mockup)

### Rewards
40 SNX

asherism:
<!-- Uploading "image.png"... —>

asherism:
<img width="712" alt="image" src="https://user-images.githubusercontent.com/11524787/177832940-f70b9fe1-34ee-4da0-929a-92c7ac4db579.png">

PR #1184, opened 2022-07-26
Add Skew data in mobile market stats tab

Xiaolei-Shawn:
## Description

- Render the market stats for mobile according to design
- Introduce Skew data hook
- Refactor `MarketDetails` to use hooks 

## Related issue

[Issue](https://github.com/Kwenta/kwenta/issues/1006)

## How Has This Been Tested?

Test it locally:
- Open Kwenta in Device simulation mode in Chrome developer tool/mobile
- Open `Futures`
- Open `Stats` tab

image


koredefashokun (commented):
Generally, everything looks good. We should be moving most data manipulations to hooks. Found a couple of things though:

- I think the hook should be a `.ts` file, and not a `.tsx` one, to discourage contributors from putting React components in the hook.
- The `enableTooltip` function can be refactored. It duplicates a lot of code. Ideally, we can create a map from key to the tooltip content. However, I understand that you may not be able to make this change in this PR given the added complexity that would add for handing the custom styles for the `TimerTooltip` and the `OneHrFundingRateTooltip`.
- We only use the `futureRates` thing to get the rate for the current market. In this case, it might make more sense to get the rate for only the selected market using `.find(...)` instead of `.reduce(...)`.

Xiaolei-Shawn:
> * I think the hook should be a `.ts` file, and not a `.tsx` one, to discourage contributors from putting React components in the hook.

It makes sense. Hook should not produce any React component. 👍 

> * We only use the `futureRates` thing to get the rate for the current market. In this case, it might make more sense to get the rate for only the selected market using `.find(...)` instead of `.reduce(...)`.

It was in original implementation but I can surely update it.

platschi:
> The enableTooltip function can be refactored. It duplicates a lot of code. Ideally, we can create a map from key to the tooltip content. However, I understand that you may not be able to make this change in this PR given the added complexity that would add for handing the custom styles for the TimerTooltip and the OneHrFundingRateTooltip.

I will scope this out in a separate ticket!

koredefashokun (approved):

Issue #1051, opened 2022-06-25
Fix price feed for atomic exchange


platschi:
### Subject of the issue
Atomic Swaps on Kwenta Exchange use the chainlink price. We should use the `value` number instead.

> that `value` number is the price user will be getting when exchange atomically, `systemValue` is the chainlink price `systemSourceRate` is chainlink price of source and `systemDestinateRate` is the chainlink price of destination
> `ExchangeRatesWithDexPricing`

### Deliverables
- Make sure atomic exchange price feed uses `value`

platschi:
Note: The current implementation of `useExchange.ts` needs to be cleaned up first before adding the Atomic Price Feed. Ticket has been paused until then.

LeifuChen:
> Note: The current implementation of `useExchange.ts` needs to be cleaned up first before adding the Atomic Price Feed. Ticket has been paused until then.
> 
> / Edit: Refactor done, re-evaluate this PR

I will create a new PR to add this feature.

PR #1186, opened 2022-07-27
Add a new price feed for atomic swap

LeifuChen:
## Description
The atomic swap and the fee reclamation may use different exchange rates. We want to use the different price feed for the different swap types.

## Related issue
#1051

## Motivation and Context
Show the more accurate amount that the user could receive when the atomic swap could be triggered.

## How Has This Been Tested?
1. Open the market page and switch to L1
2. Choose two different currencies except for sLink
3. See the amount that the user will receive on UI and compare it with https://dev.kwenta.io/exchange


platschi:
Note: PR blocked/paused until #1222 is solved. See latest comment in #1051 for context.

LeifuChen:
The exchange page has been refactored so this PR is out-of-date.

Issue #1177, opened 2022-07-25
Remove unused data hooks


platschi:
- [x] Remove unused data hooks
- [x] Remove polling from data hooks which don't need it

PR #1191, opened 2022-07-27
Improve futures data hooks

Tburm:
Partially refactor work, but also includes other needed improvements to data fetching. Removed unused hooks, and improves others where we can use recoil values instead of passing parameters.

## Description
* Consolidated data from other hooks into `useGetFuturesMarkets`
  * Resulted in deleting the `useNextPriceDetails` query
  * Resulted in deleting the `useGetFuturesMarketLimit` query
  * Resulted in deleting the `useGetFuturesOpenInterest` query
  * Removes `useGetFuturesMarket` query and replaces the result with a selector from `futuresMarkets`
* Updated funding rate queries to use recoil values for market info
* Updated daily trade stats query to use generated subgraph to avoid recursive fetching
* Updated `useGetFuturesOpenOrders` to use market info from recoil

## Related issue
- #1177 

## Motivation and Context
Many of our hooks are not using recoil values. They still rely on parameters being fetched and passed within components. These updates remove old hooks, and update others to follow current patterns for state management.

## How Has This Been Tested?
* Left market page open for a long time (check for updates)
* Left dashboard open for a long time (check for updates)
* Submitted market orders on testnet
* Submitted next price orders on testnet


koredefashokun (dismissed):
Looks good to me.

queries/futures/useGetFuturesDailyTrades.ts
```diff
@@ -1,31 +1,28 @@
...
 import { calculateTimestampForPeriod } from 'utils/formatters/date';
-import { FuturesMarketAsset } from 'utils/futures';
 import logError from 'utils/logError';
 
 import { DAY_PERIOD } from './constants';
 import { getFuturesTrades } from './subgraph';
 import { getFuturesEndpoint } from './utils';
 
-const useGetFuturesDailyTradeStatsForMarket = (
-	marketAsset: FuturesMarketAsset | null,
```
avclarke, line 17: I feel like it makes sense for this hook to take a market as a param so it can then be re-used in any place, otherwise it's restricted to only being used for the single market in recoil.

Tburm: good point, I added it back. I was trying to avoid passing recoil values as params for the hooks we are still calling from components. This one is only used in one place (`MarketDetails`) so it wouldn't make a difference unless we reuse this hook elsewhere.

Tburm: Right now we're able to pull all or most contract data for all markets in a single hook and store in a recoil object. I'm thinking we could use a similar pattern for subgraph data so we don't call the hooks all over the place in components.

sections/futures/MarketDetails/MarketDetails.tsx
```diff
@@ -136,7 +108,7 @@ const MarketDetails: React.FC<MarketDetailsProps> = ({ mobile }) => {
 						{children}
 					</TimerTooltip>
 				);
-			case fundingTitle:
+			case 'Inst. Funding Rate' || '1H Funding Rate':
```
avclarke, line 111: Is there something else we can use here other than label strings as this would break on other languages?

Tburm: The issue is that we use the titles as keys here, but the title is conditional. All of these keys are based on the title that is shown, so we would need to update the entire component to display translations.

Tburm: Let me write a separate issue for that and resolve it separately from this PR.

avclarke: ok sounds good

avclarke (approved):
Looks good!

sections/futures/MarketDetails/MarketDetails.tsx
Tburm: Issue here: https://github.com/Kwenta/kwenta/issues/1207

PR #1201, opened 2022-07-29
fix: truncated 'size' value at Future Positions above 1M.

### LINKED ISSUE #993: Futures Positions table Size issue
### Subject of the issue
For orders of a size >1 Million $, the Futures Position tables Size and Leverage values overlap, making it difficult to read values. The table should be more responsive in order to avoid this issue for whales, see screenshot:

whales

### Deliverables
- Truncate **Size** values from and above $1.000.00,00 in `FuturesPositionTable` (see conversation below)

### Reward
25 SNX

zhyd1997:
## Related issue

Fixes #993


platschi (approved):
Great, truncated values work as expected. Tested on OP Kovan:

Screen Shot 2022-07-31 at 11 22 24

LeifuChen (approved):
Looks good to me.

Issue #1024, opened 2022-06-16
Improve MarketDetails for smaller screens


platschi:
### Subject of the issue
a) On smaller/tablet size screens, the `MarketDetails` container metrics wrap out of their container. This should be avoided by improving the layout of the container. See comments below in this ticket.

![Image](https://user-images.githubusercontent.com/548702/180887154-6fa7c3a5-bd64-4eca-a82d-dabac8265bb4.png)

b) The padding of the tooltip for the Oracle price ("Last Update: nn:nn") is off compared to all other default tooltips. This should be fixed. See screenshots below:

Current padding of tooltip:
<img width="591" alt="Screen Shot 2022-07-29 at 11 20 53" src="https://user-images.githubusercontent.com/548702/181780649-844d1e3e-915c-4303-b4c8-a27a0ebed323.png">

Padding should be the same as other tooltips, e.g.:
<img width="591" alt="Screen Shot 2022-07-29 at 11 20 57" src="https://user-images.githubusercontent.com/548702/181780728-276f21f2-21ff-4983-9d32-dc159ed486fe.png">

### Deliverables
- Fix the "Last update" oracle price tooltip padding
- Add a nowrap white-space property to all metrics
- Make sure container scrolls horizontally when space is scarce
- Make sure tooltips are not clipped by horizontal scrolling and show correctly (see comment below)

### Reward
40 SNX

platschi:
This should be avoided:

![Image](https://user-images.githubusercontent.com/548702/180887154-6fa7c3a5-bd64-4eca-a82d-dabac8265bb4.png)

asherism:
All metrics should have a nowrap white-space property and the container should simply scroll horizontally when space is scarce. Note: scrolling the container contents will clip the tooltips- final fix should account for this.

PR #1209, opened 2022-07-31
fix/smaller screens

KngZhi:
- fix: align tooltip padding
- feat: nowrap
- feat: vertical scroll on MarketDetails
- feat: add margin while less than xl

## Related issue
- #1024

https://user-images.githubusercontent.com/11361147/182017366-68ed59c0-04f0-4197-ab2b-93ea5f6ed144.mov


platschi:
Thanks! Scrolling vertically on smaller screens works fine! A few things:

1) Tooltips on Firefox are still way off, while they do work fine on Chrome.

2) On Firefox, the `MarketDetails` now scrolls horizontal, too. This seems to be a Firefox related flexbox bug and could be mitigated by removing the set height of 55px. Not sure if this affects the box otherwise, though. Chrome seems unaffected of this bug.

@asherism might have a better understanding of the issues at hand.

asherism:
> 2\. On Firefox, the `MarketDetails` now scrolls horizontal, too. This seems to be a Firefox related flexbox bug and could be mitigated by removing the set height of 55px. Not sure if this affects the box otherwise, though. Chrome seems unaffected of this bug.

i'm confused- we want it to scroll horizontally, no?

KngZhi:
On Firefox:

image

platschi (approved):
Thanks, fixes works fine now on both Chrome and Firefox!

platschi:
> > 2. On Firefox, the `MarketDetails` now scrolls horizontal, too. This seems to be a Firefox related flexbox bug and could be mitigated by removing the set height of 55px. Not sure if this affects the box otherwise, though. Chrome seems unaffected of this bug.
> 
> i'm confused- we want it to scroll horizontally, no?

Yeah, I mixed up horizontal and vertical in my comment. Looks like it's all working fine now.

LeifuChen (approved):
@Junwei#7322 The root cause of this issue is that, in Chrome, the `fixed` position element (tooltip in our case) is relative to the viewport (browser window) while, in Firefox, the `fixed` position element is relative to the parent element (e.g. `MarketDetailContainer`). 

One possible solution is setting a new stacking context to make the `fixed` position element in Chrome relative to the parent element too. 

e.g.
image

I don't think it's better than your solution so let's keep your solution then.

Issue #1220, opened 2022-08-02
Futures Hooks Improvements


Tburm:
Improve hooks and data fetching by removing data and hook logic from components.

AC:
- [x] Create recoil value for funding rates by market
- [x] Improve the `useLaggedDailyPrice` hook by fetching a block number and querying `LatestRates`
- [x] Add `marketName` to `futuresMarkets`
- [x] Call rates ones and pull from state everywhere

PR #1221, opened 2022-08-02
Fix market name

Tburm:
Add the `[ASSET]-PERP` market name to the market object, and replace all instances to use the market info or util function.

## Description
Replace all market name instances with a version created by utility function.

## Related issue
- #1220


platschi (dismissed):

sections/futures/ProfitCalculator/ProfitDetails.tsx
```diff
@@ -86,7 +86,7 @@ const ProfitDetails: React.FC<ProfitDetailsProps> = ({
 							? wei(marketAssetPositionSize).toNumber().toFixed(2)
 							: ''}
 					</RowText>
-					<RowText className="gray-font-color">{`${marketAsset}-PERP`}</RowText>
+					<RowText className="gray-font-color">{marketAsset}</RowText>
```
koredefashokun, line 89: Did you mean to wrap this in `getMarketName`, or does the `marketAsset` value have the "-PERP" suffix already?

koredefashokun (commented):
Looks good. Just left one comment about a place where we might still need to import the `getMarketName` function. Was also wondering if we could maybe make the `getMarketName` function only take in a `FuturesMarketAsset`, to make sure that we don't misuse it in the future. I understand that that might lead to a number of type fixes that might be beyond the scope of this PR though.

sections/futures/ProfitCalculator/ProfitDetails.tsx
Tburm: Fixed this with a new commit

platschi (approved):

PR #1233, opened 2022-08-04
Fix the missing futures markets on the landing page

LeifuChen:
## Description
The user cannot see the future market cards on the landing page.

## How Has This Been Tested?
1. Open the landing page and the user can see the future market cards again.


platschi (approved):
Thanks, fix resolves issues on landing page!

Issue #1240, opened 2022-08-05
Profit Calculator: Resize window


platschi:
### Subject of the issue

From a Discord user:
> any thoughts on making the calculator resizable or smaller, can't see current position stats when it's open

The current profit calculator modal is static, as in using a lot of space on the screen without the option for users to resize/move the modal, as indicated in the user comment above, for example to have the calculator not cover currently open position data.

### Deliverables
- Make Profit Calculator modal resizable and draggable

### Notes
For an example on resize and drag, [see this code example](https://wwan5803.github.io/react-modal-resizable-draggable/)

### Bounty
40 SNX

PR #1249, opened 2022-08-09
Enable dragging and resizing on profit calculator

Xiaolei-Shawn:
## Description

- introduce new dependency [react-rnd](https://github.com/bokuweb/react-rnd) for modal resizing and dragging
- Provide a way to enable resizable and draggable for `BaseModal`.  As default, only enable them on `ProfitCalculator`.
- The way to enable resizing and dragging on `BaseModal` is to pass
```
resizableAnDraggableProps={{
	enableResizing: true,
	disableDragging: false,
}}
```

## Related issue

[Issue](https://github.com/Kwenta/kwenta/issues/1240)

Original:
image

New: 
Screen Recording 2022-08-09 at 17 20 29 2022-08-09 17_22_49


platschi (commented):
Looks great, however two points to make the Profit Calculator less "huge"/taking over the whole screen:

- Try to reduce padding on left/right 
- For `ModalWindow`, lower the default size, maybe to something like the below?

```
const ModalWindow = styled.div`
	height: 700px;
	width: 440px;
...
```

platschi (dismissed):

components/BaseModal/BaseModal.tsx
```diff
@@ -93,6 +98,8 @@ const StyledCardBody = styled(Card.Body)`
 	overflow-y: scroll;
 `;
 
+const StyledResizableAndDraggableContainer = styled(Rnd)``;
```
koredefashokun, line 101: Since we don't use this for styling, I'm not sure we need to wrap the `Rnd` component with `styled`.

sections/futures/ProfitCalculator/ProfitCalculator.tsx
```diff
@@ -131,10 +131,15 @@ const ProfitCalculator: FC<ProfitCalculatorProps> = ({ marketAsset, setOpenProfi
 
 	return (
 		<>
-			<BaseModal
+			<StyledBaseModal
 				onDismiss={() => setOpenProfitCalcModal(false)}
 				isOpen
 				title={t('futures.modals.profit-calculator.title')}
+				resizableAnDraggableProps={{
```
koredefashokun, line 138: This seems to have a typo. I think it should be `resizableAndDraggableProps`. Moreover, it should be fine to just name it `dndProps`.

koredefashokun (commented):
Looks pretty good. Left a few comments.

components/BaseModal/BaseModal.tsx
Xiaolei-Shawn: It's just for sake of consistency on naming convention. Let me remove it.

sections/futures/ProfitCalculator/ProfitCalculator.tsx
Xiaolei-Shawn: I will name it as `rndProps`  since the component's name is `ResizableAndDraggable`

platschi (approved):

PR #1252, opened 2022-08-09
Add past prices to state

Tburm:
Add past prices to state instead of fetching from components

## Description
* Add `pastRates` to futures state variables
* Get rates from state in each component
* Update `useGetLaggedDailyRates` component to fetch a block and get all latest rates
* Improves `ChangePercent` component to handle zeros and null values

## Related issue
- #1191 

## Motivation and Context
Continue moving data fetching from components to state


components/ChangePercent/ChangePercent.tsx
```diff
@@ -14,21 +15,31 @@ type ChangePercentProps = {
 };
 
 export const ChangePercent: FC<ChangePercentProps> = ({ value, decimals = 2, ...rest }) => {
-	const isPositive = wei(value ?? 0).gt(0);
+	const isValid = value;
+	const isZero = value && wei(value).gt(0);
```
koredefashokun, line 19: I think this should be:
```suggestion
	const isZero = value && wei(value).eq(0);
```

components/ChangePercent/ChangePercent.tsx
```diff
@@ -14,21 +15,31 @@ type ChangePercentProps = {
 };
 
 export const ChangePercent: FC<ChangePercentProps> = ({ value, decimals = 2, ...rest }) => {
-	const isPositive = wei(value ?? 0).gt(0);
+	const isValid = value;
```
koredefashokun, line 18: Maybe this can be:
```suggestion
	const isValid = !!value;
```
to ensure the value is an actual boolean.

Tburm:
@platschi @koredefashokun conflicts are fixed, ready for review

koredefashokun (approved):

PR #1261, opened 2022-08-11
Set Competition Data

platschi:
## Description
- Sets competition to start on August 16th, 00:00 and ending on August 23rd, 00:00
- Sets competition banner link to published blog post with information

### DIFF
--- a/constants/competition.ts
+++ b/constants/competition.ts
@@ -4,6 +4,6 @@
  */
 export const COMPETITION_DATES = {
 	// note: month starts at 0, not 1.
-	START_DATE: new Date(2022, 7, 18),
-	END_DATE: new Date(2022, 7, 22),
+	START_DATE: new Date(2022, 7, 16),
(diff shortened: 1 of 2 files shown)


constants/competition.ts
```diff
@@ -4,6 +4,6 @@
  */
 export const COMPETITION_DATES = {
 	// note: month starts at 0, not 1.
-	START_DATE: new Date(2022, 7, 18),
-	END_DATE: new Date(2022, 7, 22),
+	START_DATE: new Date(2022, 7, 16),
```
LeifuChen, line 7: Note that it's different from the description `Sets competition to start on August 16th, 00:00 and ending on August 23rd, 00:00`

LeifuChen (changes requested):

constants/competition.ts
LeifuChen: Got it. It's `monthIndex`

LeifuChen (approved):

PR #1277, opened 2022-08-16
Fix swap conversion error

avclarke:
Some 1inch assets were not getting the correct conversion rate and returning an error when trying to swap. This was caused by not using the correct token decimal in some places during the quote and swap.

This also fixes an issue with gas estimation not displaying.


koredefashokun (approved):
Changes look good. Great to finally get rid of `newGetTransactionPrice`

Issue #1273, opened 2022-08-15
Spot Exchange: Title Bug


platschi:
### Subject of the issue
The page title does not show the exchange rate when selecting synth/token pairs. E.g., selecting sMATIC/sINR shows the correct exchange rate in title, while DAI/sINR wrongly reverts to 0.0000:

![Image](https://user-images.githubusercontent.com/548702/184692337-f9f81705-5d6d-4122-acfd-42d213476d8a.png)

### Related issues
A similar issue was fixed in the scope of ticket #1234 

### Deliverables
- Make sure Title shows the correct exchange rate for all trading pairs

### Rewards
150 $OP

PR #1291, opened 2022-08-17
Resolve rate from coinGecko if exchange rate is not found

Xiaolei-Shawn:
## Description

- Resolve currency rate from `coinGecko` if its key is not found in `exchangeRates`.
- Fix a bug when calculating exchange rate that base and quote currency key are passed to function `newGetExchangeRatesForCurrencies` with wrong order: https://github.com/Kwenta/kwenta/blob/51068f9900e9f597adc85aabd9825ea014179a5c/hooks/useExchange.ts#L278
## Related issue
 
[Issue](https://github.com/Kwenta/kwenta/issues/1273)

## Motivation and Context

The title in exchange won't show correct rate if one or both currencies are not found in exchange rates results.

## How Has This Been Tested?

- Go to `Exchange` view
- Select `sINR` and `DAI`
- Tab title in browser should show correct exchange rate
- Switch currency to see title change

image
image


platschi:
@Xiaolei-Shawn Kindly check lint errors :)

Xiaolei-Shawn:
> @Xiaolei-Shawn Kindly check lint errors :)



> @Xiaolei-Shawn Kindly check lint errors :)

My bad! Fixed already.

LeifuChen:
@Xiaolei-Shawn 
I think the function signature is ordered like this by some reason. The rate on title has been fixed but it caused the issue for the calculation of the received amount.
image

pages/exchange.tsx
```diff
@@ -32,7 +32,7 @@ const Exchange: ExchangeComponent = () => {
 						? t('exchange.page-title-currency-pair', {
 								baseCurrencyKey,
 								quoteCurrencyKey,
-								rate: formatCurrency(quoteCurrencyKey, inverseRate, {
+								rate: formatCurrency(quoteCurrencyKey, rate, {
```
LeifuChen, line 35: As you may notice, it used the `inverseRate` on page title. It explains why the order of the function signature is a bit strange.

LeifuChen (changes requested):

pages/exchange.tsx
Xiaolei-Shawn: Now revert to inverted rate

LeifuChen:
@Xiaolei-Shawn The rate of synthetic assets (the assets with the prefix `s`, which is fetching the price from `rate` not via coincgecko api) is not working now.
image

LeifuChen (approved):
Looks good to me.

platschi (approved):
Perfect, works as expected.

PR #1300, opened 2022-08-18
Bug/exchange rates

Tburm:
Fix exchange rates on swap for synths that have futures markets (sEUR)

## Description
- change the `rates` object in state to include synth and asset names (`sEUR` and `EUR`)
- Update the ETH address for swap pricing to the WETH contract on Optimism

### DIFF
--- a/constants/currency.ts
+++ b/constants/currency.ts
@@ -77,7 +77,7 @@ export const COMMODITY_SYNTHS = new Set<CurrencyKey | 'XAU' | 'XAG' | 'WTI'>(['X
 export const sUSD_EXCHANGE_RATE = new Wei(1);
 export const SYNTH_DECIMALS = 18;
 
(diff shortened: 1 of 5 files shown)


platschi:
Trying ETH -> sUSD swap clicking "Max" on ETH results in 0 instead of expected max ETH balance of connected wallet

Screen Shot 2022-08-18 at 15 08 40

Tburm:
@platschi the issue is because we "buffer" ETH swaps by 0.1 ETH which seems really extreme. Since the balance is 0.017 it will default to 0, but if you had .2 ETH it would display .1 ETH

Tburm:
https://github.com/Kwenta/kwenta/blob/150de1163e8e97ae7597400f95a75094d6937638/hooks/useExchange.ts#L1117

koredefashokun (approved):

platschi (approved):

Issue #1287, opened 2022-08-17
Leaderboard Improvements


platschi:
### Subject of the issue
Connected wallets are currently pinned to the top of the leaderboard, leaving a gap in the ranking where your position would be. 

### Deliverables
- [x] Pin current user to top AND show in ranking
- [x] Hide leaderboard search bar on mobile

Tburm:
We decided to show the search bar on mobile, but display it below the selection buttons

PR #1301, opened 2022-08-18
Leaderboard improvements

Tburm:
Improve leaderboard styling on mobile and fix issues with pinning

## Description
- Add pinned row into tables at it's original place
- Fix search bar styling on mobile to display below buttons

## Related issue
- #1287


platschi (approved):
Works great now!

PR #1318, opened 2022-08-22
update kwenta links

platschi:
## Description
Update README, links and metadata throughout the dApp to point to `kwenta.eth.limo` rather than `kwenta.io`

Note: Docs will still be available via `docs.kwenta.io` until further notice


constants/links.ts
```diff
@@ -1,6 +1,6 @@
 import { CurrencyKey } from './currency';
 
-export const PROD_HOSTNAME = 'kwenta.io';
+export const PROD_HOSTNAME = 'kwenta.eth.limo';
```
platschi, line 3: Not 100% sure this is fine/needed or doesn't break any dev or CI flows

Tburm: this won't cause any problem, it's only used in one place in case the system is under maintenance

Tburm (approved):

PR #1322, opened 2022-08-25
Feat/order history

Tburm:
Changes the current "open orders" table into an "order history" table

## Description
- Updates the hook that requests orders to get all orders (not just pending)
- Adds a general checkbox component
- Implements an "open orders only" checkbox to filter the table

## Motivation and Context
This will give more information on the futures page without crowding it or adding anything new in the default state. It also sets up the frontend for the release of limit orders and stop losses when those order types are available.

### DIFF
--- a/queries/futures/useGetFuturesOrders.ts
+++ b/queries/futures/useGetFuturesOrders.ts
@@ -30,7 +30,11 @@ const useGetFuturesOpenOrders = (options?: UseQueryOptions<any>) => {
 					futuresEndpoint,
 					gql`
 						query Orders($account: String!, $market: String!) {
-							futuresOrders(where: { account: $account, market: $market }) {
+							futuresOrders(
+								where: { account: $account, market: $market }
+								orderBy: timestamp
+								orderDirection: desc
+							) {
 								id
 								account
 								size
--- a/queries/futures/useGetFuturesOrders.ts
+++ b/queries/futures/useGetFuturesOrders.ts
@@ -1,0 +1,71 @@
+import Wei from '@synthetixio/wei';
+import { utils as ethersUtils } from 'ethers';
+import request, { gql } from 'graphql-request';
+import { useQuery, UseQueryOptions } from 'react-query';
+import { useRecoilValue } from 'recoil';
+
+import { ETH_UNIT } from 'constants/network';
+import QUERY_KEYS from 'constants/queryKeys';
+import { appReadyState } from 'store/app';
+import { futuresAccountState, marketInfoState } from 'store/futures';
+import { isL2State, networkState } from 'store/wallet';
+import logError from 'utils/logError';
+
+import { getFuturesEndpoint } from './utils';
+
+const useGetFuturesOpenOrders = (options?: UseQueryOptions<any>) => {
+	const { selectedFuturesAddress } = useRecoilValue(futuresAccountState);
+	const isAppReady = useRecoilValue(appReadyState);
+	const isL2 = useRecoilValue(isL2State);
+	const network = useRecoilValue(networkState);
+	const futuresEndpoint = getFuturesEndpoint(network);
+	const marketInfo = useRecoilValue(marketInfoState);
+
+	return useQuery<any>(
+		QUERY_KEYS.Futures.Orders(network.id, selectedFuturesAddress),
+		async () => {
+			try {
+				const marketAddress = marketInfo?.market;
+				const response = await request(
+					futuresEndpoint,
+					gql`
+						query Orders($account: String!, $market: String!) {
+							futuresOrders(where: { account: $account, market: $market }) {
+								id
+								account
+								size
+								market
+								asset
+								targetRoundId
+								timestamp
+								orderType
+								status
+							}
+						}
+					`,
+					{ account: selectedFuturesAddress, market: marketAddress }
+				);
+
+				const orders = response
+					? response.futuresOrders.map((o: any) => ({
+							...o,
+							asset: ethersUtils.parseBytes32String(o.asset),
+							targetRoundId: new Wei(o.targetRoundId, 0),
+							size: new Wei(o.size).div(ETH_UNIT),
+					  }))
+					: [];
+				return orders;
+			} catch (e) {
+				logError(e);
+				return [];
+			}
+		},
+		{
+			enabled: isAppReady && isL2 && !!marketInfo?.market && !!selectedFuturesAddress,
+			refetchInterval: 5000,
+			...options,
+		}
+	);
+};
+
+export default useGetFuturesOpenOrders;
--- a/sections/futures/UserInfo/OrderHistory.tsx
+++ b/sections/futures/UserInfo/OrderHistory.tsx
@@ -93,6 +93,8 @@ const OrderHistory: React.FC = () => {
 					wei(marketInfo?.currentRoundId ?? 0).gte(wei(order.targetRoundId).add(2)) &&
 					order.status === 'Pending'
 						? 'Expired'
+						: order.status === 'Pending'
+						? 'Open'
 						: order.status,
 				market: getMarketName(order.asset),
 				marketKey: MarketKeyByAsset[order.asset as FuturesMarketAsset],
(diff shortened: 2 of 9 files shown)


queries/futures/useGetFuturesOrders.ts
```diff
@@ -0,0 +1,77 @@
...
+import { ETH_UNIT } from 'constants/network';
+import QUERY_KEYS from 'constants/queryKeys';
+import Connector from 'containers/Connector';
+import useIsL2 from 'hooks/useIsL2';
+import { futuresAccountState, marketInfoState } from 'store/futures';
+import logError from 'utils/logError';
+
+import { getFuturesEndpoint } from './utils';
+
+const useGetFuturesOpenOrders = (options?: UseQueryOptions<any>) => {
```
avclarke, line 17: I noticed this is a separate query to `useGetFuturesOpenOrders` so worth renaming to `useGetFuturesOrders`. But also is there a reason we have two very similar queries now in those two? Can we combine them and query by passing accountType to differentiate between cross margin and isolated margin, similar to here: https://github.com/Kwenta/kwenta/blob/dev/queries/futures/useGetFuturesTradesForAccount.ts

avclarke (commented):
Looks great, appreciate it's a draft so ignore if still WIP but left a couple of comments.

queries/futures/useGetFuturesOrders.ts
Tburm: One or the other will be removed. We don't need to pass accountType because I'm using `selectedFuturesAddress` which takes the type into account.

avclarke: Might have been my misunderstanding of abstractAccount, so this will either be wallet address for isolated margin trades and cross margin address for cross margin trades?

Tburm: Exactly, which makes this so we can just check if `selectedFuturesAddress == abstractAccount`. If cross margin, they are both the smart contract. If isolate margin, they are both the EOA address.

This leaves open the option for us to query where `account` is equal to the EOA address, and we will get open orders for isolated _and_ cross margin.

avclarke: Ah perfect, I can update the other queries as well then instead of using accountType. I initially thought abstract account was only assigned on CM trades.

sections/futures/UserInfo/OrderHistory.tsx
```diff
@@ -0,0 +1,368 @@
...
+
+	const gasSpeed = useRecoilValue(gasSpeedState);
+	const currencyKey = useRecoilValue(currentMarketState);
+	const marketInfo = useRecoilValue(marketInfoState);
+	const { selectedFuturesAddress } = useRecoilValue(futuresAccountState);
+
+	const { handleRefetch } = useRefetchContext();
+
+	const [action, setAction] = React.useState<'' | 'cancel' | 'execute'>('');
+	const [selectedOrder, setSelectedOrder] = React.useState<any>();
```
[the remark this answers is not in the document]
Tburm: I created a type on the relevant component since this one is deleted

queries/futures/types.ts
```diff
@@ -26,7 +26,7 @@ export type PositionDetail = {
 	profitLoss: Wei;
 };
 
-export type FuturesOrder = {
+export type FuturesPositionOrder = {
```
avclarke, line 29: Is this type only used inside the `FuturesPosition` type? Might be better to remove export if so just so it's easier to differentiate between `FuturesPositionOrder` and `FuturesOrder`

queries/futures/useGetFuturesOpenOrders.ts
```diff
@@ -9,61 +9,91 @@ import { ETH_UNIT } from 'constants/network';
...
-							targetRoundId: new Wei(o.targetRoundId, 0),
-							size: new Wei(o.size).div(ETH_UNIT),
-					  }))
+				const openOrders: FuturesOrder[] = response
+					? response.futuresOrders.map((o: FuturesOrder) => {
+							const asset: FuturesMarketAsset = ethersUtils.parseBytes32String(
+								o.asset
+							) as FuturesMarketAsset;
+							const size = new Wei(o.size).div(ETH_UNIT);
+							const targetPrice = new Wei(o.targetPrice).div(ETH_UNIT);
```
avclarke, line 62: I'm not sure it's best naming but I added a util for this: `weiFromWei` in number utils.

queries/futures/useGetFuturesOpenOrders.ts
```diff
@@ -9,61 +9,91 @@ import { ETH_UNIT } from 'constants/network';
...
+								targetPrice: targetPrice.gt(0) ? targetPrice : null,
+								size: size,
+								market: getMarketName(asset),
+								marketKey: MarketKeyByAsset[asset],
+								orderType: o.orderType === 'NextPrice' ? 'Next-Price' : o.orderType,
+								sizeTxt: formatCurrency(asset, size.abs(), {
+									sign: asset ? synthsMap[asset]?.sign : '',
+									minDecimals: size.abs().lt(0.01) ? 4 : 2,
+								}),
+								targetPriceTxt: formatCurrency(asset, targetPrice, {
```
avclarke, line 77: can use formatDollars here

queries/futures/useGetFuturesOpenOrders.ts
```diff
@@ -9,61 +9,91 @@ import { ETH_UNIT } from 'constants/network';
...
+									sign: '$',
+									minDecimals: targetPrice.lt(0.01) ? 4 : 2,
+								}),
+								side: size.gt(0) ? PositionSide.LONG : PositionSide.SHORT,
```
avclarke, line 81: Can the size be 0 if you're only adjusting margin or is this the size after the order is executed? We might need to take into account existing size if it's order size delta.

queries/futures/useGetFuturesOpenOrders.ts
```diff
@@ -9,61 +9,91 @@ import { ETH_UNIT } from 'constants/network';
...
+								isStale:
+									o.orderType === 'NextPrice' &&
+									wei(marketInfo?.currentRoundId ?? 0).gte(wei(o.targetRoundId).add(2)),
```
avclarke, line 84: As there's a few of these `wei(marketInfo?.currentRoundId ?? 0)` it might be nice to do a single `const currentRound = wei(marketInfo?.currentRoundId ?? 0)` above

sections/futures/MobileTrade/drawers/OrderDrawer.tsx
```diff
@@ -2,21 +2,30 @@ import React from 'react';
...
 
 const OrderDrawer: React.FC<OrderDrawerProps> = ({ open, order, closeDrawer, setAction }) => {
 	const items = React.useMemo(() => {
-		if (!order) return [];
+		if (!order || !order.side) return [];
+
+		const price = order.targetPrice
+			? [
+					{
+						label: 'Price',
```
avclarke, line 24: Should probably translate this

sections/futures/UserInfo/OpenOrdersTable.tsx
```diff
@@ -163,12 +131,29 @@ const OpenOrdersTable: React.FC = () => {
...
+								return (
+									<div>
+										<div>{cellProps.row.original.sizeTxt}</div>
+										{cellProps.row.original.targetPrice && (
+											<Currency.Price
+												currencyKey={'sUSD'}
+												price={cellProps.row.original.targetPrice}
+												sign={'$'}
+												conversionRate={1}
+												formatOptions={formatOptions}
```
avclarke, line 152: If this is `sUSD` then we don't need formatOptions for other assets do we?

sections/futures/UserInfo/OpenOrdersTable.tsx
```diff
@@ -229,9 +218,32 @@ const OpenOrdersTable: React.FC = () => {
...
```
avclarke, line 241: Also here do we need formatOptions on isDecimalFour if it's sUSD

avclarke (changes requested):
Looks good, just a few comments and questions.

queries/futures/useGetFuturesOpenOrders.ts
Tburm: Is this possible, to set a "limit order" where the action is to simply top up margin?

If so, we would need a third option here for when the order size is zero but `marginDelta` is nonzero.

Tburm: Actually looks like this is not possible: https://github.com/Kwenta/margin-manager/blob/cc9da276957679c5af887d3f31e40216ff27f851/contracts/MarginBase.sol#L789

Order will fail if the `sizeDelta` is zero.

avclarke: ok all good then

avclarke (approved):
Looks good!

PR #1324, opened 2022-08-26
Use ethcall for positionDetailsForMarketKey

koredefashokun:
## Description
This PR batches contract calls in the `useGetFuturesPositionForMarkets` hook using ethcall. It should improve performance on the app quite a bit, since the query was getting refetched quite a bit.

## Related issue
N/A

## Motivation and Context
Performance improvements

## How Has This Been Tested?
N/A

### DIFF
--- a/lib/abis/FuturesMarket.json
+++ b/lib/abis/FuturesMarket.json
@@ -1,0 +1,1109 @@
+[
+	{
+		"inputs": [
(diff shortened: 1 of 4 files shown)


Tburm (approved):
Works great, tested on OP mainnet with 2 positions open

Issue #1295, opened 2022-08-18
Spot History: Consistent Currency symbols


platschi:
### Subject of the issue
The Spot history is not consistent when it comes to currency symbols. US Dollar values are indicated differently when showing four decimals, e.g. `2.2123 $` while two decimal values are shown correctly as `$8.74`, making the table look inconsistent. The currency symbol should be uniform with the rest of the UI being in front of the numeric value.

![Image](https://user-images.githubusercontent.com/548702/185265123-0990590e-fa49-4880-ab55-26ae68d93183.png)

### Deliverables

- [ ] Fix the Spot History `USD Value` column and make sure values with four decimals are shown with the $ symbol in front of the value, too. In the example above it should be: `$2.2123`.

### Bounty
65 $OP

PR #1325, opened 2022-08-26
Always format USD value of spot trades with a sign instead of a currencyKey

0xwalde:
## Description
Updates the Spot History table to always format the USD Value column with a `sign` (on the left) instead of a `currencyKey` (on the right). Previously the value was being formatted on the left for trades involving fiat synths and on the right for non-fiat synths, but that doesn't make a whole lot of sense since USD Value will always be in fiat terms.

## Related issue
https://github.com/Kwenta/kwenta/issues/1295

## Motivation and Context
Inconsistent currency symbol display is confusing and doesn't look great

## How Has This Been Tested?
Tested locally against a mainnet fork

## Screenshot
Kwenta-1295-screenshot


LeifuChen (changes requested):
Good work. Please clean the code as `Code scanning` suggested.

0xwalde:
Ah thanks! I missed that lint. Should be fixed now.

LeifuChen (approved):

PR #1327, opened 2022-08-27
Provide a default provider

bachstatter:
Provide a default provider when not connected


LeifuChen:
@bachstatter Thanks. However, we get this error in the deployment.
You may check it by yourself: https://kwenta-git-provide-default-provider-kwenta.vercel.app/market/?asset=sETH
image

bachstatter:
Thanks for reporting. Not sure why we're not running into this. I've made a PR to js-monorepo which will fix it.. 
https://github.com/Synthetixio/js-monorepo/pull/277

bachstatter:
@LeifuChen should be fixed now

LeifuChen:
@bachstatter Thanks for the solution. It has been added to #1211.

Issue #1314, opened 2022-08-22
Fix Spot market names on landing page


platschi:
### Subject of the issue
On the Kwenta Landing Page, when selecting Spot Markets, sUSD is just labeled as "US", Indian Rupees as "Indian". This should be fixed.

### Deliverables

- [x] Make sure the US Dollar is labelled as "US Dollar", not "US"
- [x] Make sure Indian Rupees

![Image](https://user-images.githubusercontent.com/548702/185943521-36ea65c4-6980-4a26-80c1-57aa7592170b.png)

PR #1330, opened 2022-08-29
fix forex spot names on landing page

platschi:
## Description
Forex **Spot** descriptions on landing page are cut off, both on L1 and L2. For example, the South Korean Won is just displayed as "South", and so forth.

Screen Shot 2022-08-29 at 17 39 40

## Related issue
#1314 

## Motivation and Context
Order on the landing page

## How Has This Been Tested?
localhost

Screen Shot 2022-08-29 at 17 42 14


LeifuChen (approved):
Good catch. Thanks.

Issue #1332, opened 2022-08-29
Hide Competition Elements


platschi:
Hide the competition elements from the UI. 

Important: We do not want to remove the existing code because we are expecting the next competition to run in October 2022. We only want to hide elements from the UI.

### Deliverables

Note: All elements apply to both desktop and mobile! Make sure both are hidden!
- [x] Dashboard: Hide Competition Banner
- [x] Dashboard: Hide Competition Leaderboard in right sidebar (Desktop only)
- [x] Leaderboard: Hide Competition Banner
- [x] Leaderboard: Hide Bronze, Silver, Gold competition buttons and make sure the leaderboard defaults to "All"

### Bounty
150 $OP

### Deadline:
Wednesday August 31st, 2022

PR #1334, opened 2022-08-30
Add environment variable to toggle competition UI on/off

0xwalde:
Hides the competition UI elements while leaving them configurable for use in later competitions.

## Description
Introduces an environment variable `NEXT_PUBLIC_COMPETITION_ACTIVE` to control whether or not the competition UI elements display. Allows these elements to be shown or hidden independent of the `COMPETITION_DATES` set in [constants/competition.ts](https://github.com/Kwenta/kwenta/blob/dev/constants/competition.ts) so that the countdown and "Competition ended" states can still be displayed.

## Related issue
https://github.com/Kwenta/kwenta/issues/1332

## How Has This Been Tested?
Tested against a local fork of Optimism mainnet, on both desktop and mobile displays. I checked the Dashboard/Overview and Leaderboard pages since that seemed to be mainly where the competition elements show up, but just let me know if there's anywhere else I missed!

Kwenta-1332-1
Kwenta-1332-2
Kwenta-1332-3


.env.local.example
```diff
@@ -3,4 +3,5 @@ NEXT_PUBLIC_PORTIS_APP_ID="PORTIS_APP_ID"
 NEXT_PUBLIC_BN_NOTIFY_API_KEY="BLOCKNATIVE_NOTIFY_API_KEY"
 NEXT_PUBLIC_BN_ONBOARD_API_KEY="BLOCKNATIVE_ONBOARD_API_KEY"
 NEXT_PUBLIC_INFURA_PROJECT_ID="INFURA_PROJECT_ID"
-NEXT_PUBLIC_BLASTAPI_PROJECT_ID="BLASTAPI_PROJECT_ID"
\ No newline at end of file
+NEXT_PUBLIC_BLASTAPI_PROJECT_ID="BLASTAPI_PROJECT_ID"
+NEXT_PUBLIC_COMPETITION_ACTIVE=false
```
0xwalde, line 7: I made this a new environment variable because that felt like the right level of configurability (e.g. can easily change it without having to deploy new code), but just let me know if anyone thinks this should be handled differently (aka a server-side env var or just hardcoded).

avclarke (approved):
Nice work, looks good! I think the env var for config makes sense and allows us to merge and deploy future competition changes before we actually set it live.

platschi (approved):
All elements disappeared on deploy, works like a charm. Thank you!

Issue #1328, opened 2022-08-29
Revert to using Synths constant instead of currency key strings


avclarke:
### Subject of the issue
When the Synthetix package was updated, the Synths constant was removed and we replaced it with raw strings for currency keys but this has an impact on type security and code cleanliness so it would be good to add the Synths constant back into our own codebase. 

It was previously mostly used across the codebase to format sUSD values in `formatCurrency(Synths.sUSD, value, { sign: '$' })`, we should add a wrapper function for this `formatDollars(value, options)` so we not longer need to imports Synths everywhere and receptively specify '$' sign.

## Deliverables

- [ ] Re-introduce Synths constant
- [ ] Replace `formatCurrency(Synths.sUSD, value, { sign: '$' })` with `formatDollars`

## Rewards
200 $OP

PR #1343, opened 2022-09-01
ref/revert synths constant

KngZhi:
- refactor: remove hardcode sUSD

## Related issue

#1328


utils/formatters/number.ts
```diff
@@ -151,6 +152,9 @@ export const formatCurrency = (
 		? formatFiatCurrency(value, options)
 		: formatCryptoCurrency(value, options);
 
+export const formatDollars = (value: WeiSource, options?: FormatCurrencyOptions) =>
+	formatCurrency(Synths.sUSD?.name as string, value, { sign: '$', ...options });
```
avclarke, line 156: I think we can hardcode 'sUSD' in this one place instead of casting to string because of the optional.

avclarke (changes requested):
Looks good, just a minor change. I know the initial task involved reverting to using the mapping instead of lots of string keys everywhere but now Synthetix has changed the interface, I think we'd need to recreate the original in our code, so we could do `Synths.sUSD` instead of dealing with `Synths.sUSD?.name` etc. But with removing all the 'sUSD' occurrences with formatDollar wrapper I'd be happy to just use a string in that one place and we can add Synths constant later if we find we're hardcoding more keys.

utils/formatters/number.ts
KngZhi: Okay, got it, and thanks for your thorough explanation :)

platschi (approved):

PR #1352, opened 2022-09-01
Fix enabled networks

Tburm:
This change updates the "enabled networks" for testnet


queries/futures/useGetFuturesMarkets.ts
```diff
@@ -45,12 +51,15 @@ const useGetFuturesMarkets = (options?: UseQueryOptions<FuturesMarket[]>) => {
 				FuturesMarketData.globals(),
 			]);
 
-			const enabledMarkets = markets.filter((m: any) => {
+			const filteredMarkets = markets.filter((m: any) => {
 				const asset = utils.parseBytes32String(m.asset) as FuturesMarketAsset;
-				return !!MarketKeyByAsset[asset];
+				const market = enabledMarkets.find((market) => {
```
avclarke, line 56: Just to check I understand correctly, there are certain markets that will be returned from the contract but we don't want to enable, so we have a hardcoded copy of what markets to enable on each network?

avclarke (approved):
Looks good, just had a question about why we filter like this.

queries/futures/useGetFuturesMarkets.ts
platschi: Yes, because usually Synthetix pushes new market contracts in a paused state but it takes weeks (e.g. XMR-PERP is available but still paused since ~3 weeks) for them to be live. This avoids us having paused new markets on the UI for indefinite amounts of time.

avclarke: ok makes sense and we can't read the paused state from the contracts?

PR #1333, opened 2022-08-30
feat: mobile theme switch support.

zhyd1997:
## Related issue

Closes #1219

## Screenshots (if appropriate):

https://user-images.githubusercontent.com/31362988/187456683-37a271b4-b868-43e5-9bc4-2f03e9bf1365.mov

PR #1354, opened 2022-09-02
hotfix: disabled theme switch if expected theme is already set.

zhyd1997:
## Related issue

N/A

## Motivation and Context

This PR fixes the bug introduced in #1333 

If the current theme is already `light`, then clicked the button should do nothing **instead of** change the theme to `dark`, which is unexpected, and the same to the `dark` button.


platschi (approved):
Good catch, thanks! Works as expected now.

LeifuChen (approved):
Good catch. Thanks for the fix.

Issue #1331, opened 2022-08-29
Landing Page: L1 Spot Markets incomplete


platschi:
### Subject of the issue
When connected to L1, the spot markets do not display charts nor recent trade data. See screenshot. Even if there's no trade volume, the Chart and 24H price change should display.

![Screen Shot 2022-08-29 at 17 39 40](https://user-images.githubusercontent.com/548702/187295630-058a74c6-e4b9-46bc-87cf-62e73e08e867.png)

### Your environment
* L1 Ethereum connected to wallet
* Firefox

### Steps to reproduce
Visit kwenta, go to dashboard, connect a wallet and switch to L1, then click the logo to visit the landing page in order to see the Spot markets available on L1. Most pairs have no data.

PR #1363, opened 2022-09-07
Show L2 spot data only on landing page

LeifuChen:
## Description
The spot tab will show L1 spot data if the wallet is connected to Ethereum mainnet. However, the price feed is hardcoded to L2 so the L1 spot is missing most of the info. The PR will disable the query of L1 spot data so the landing page will only show L2 spot data.

## Related issue
#1331

## How Has This Been Tested?
1. Open the landing page
2. Switch to `spot`  tab, see L2 spot data
3. Go to the market page and connect the wallet. Switch the network to Ethereum
4. Go to  the landing page and switch to `spot` tab, see L2 spot data.


platschi (approved):
Works fine now, only shows L2 synths on landing page :+1:

Issue #1373, opened 2022-09-09
Fix ETH/BTC/LINK labels in Futures Markets input field


platschi:
### Subject of the issue
The input field on Futures Markest still shows the "s" for the ETH, BTC and LINK markets, whereas all other markets labels the input in the respective token name (LINK, APE, etc.). We should make sure ETH, BTC and LINK also are labeled as such, without the "s" synth indicator.

### How it is
![Image](https://user-images.githubusercontent.com/548702/189371584-ae566f15-d683-478d-9b86-b7a86c702157.png)

### Howe it should be

![Image](https://user-images.githubusercontent.com/548702/189371971-6b5872c0-58ff-4e90-9f3d-fdc7f322e40b.png)

### Deliverables
- Fix the input labels for ETH-PERP, BTC-PERP  and LINK-PERP  to make sure they only show ETH, BTC and LINK respectively.

### Reward
65 $OP

PR #1376, opened 2022-09-09
Remove leading "s" from ETH/BTC/LINK assets display in `OrderSizing` component

0xwalde:
Makes the ETH, BTC, and LINK asset names in the order size input consistent with the other Futures Market assets by removing the leading "s". 

## Description
This change is just a UI override of the underlying market asset name to remove the leading "s" for display. I initially wanted to update the `FuturesMarketAsset` representations in `utils/futures.ts` and the Recoil state to exclude the "s" so it would be consistent throughout the app, but decided against it since these markets are still called sETH/sBTC/sLINK in the underlying `FuturesMarketData` contract.

Happy to go another direction with it if anyone thinks the simple UI override isn't enough though!

## Related issue
https://github.com/Kwenta/kwenta/issues/1373

## Motivation and Context
Inconsistency between the futures markets

## How Has This Been Tested?
Test locally against Optimism Goerli

kwenta-1373-screenshot

### DIFF
--- a/sections/futures/OrderSizing/OrderSizing.tsx
+++ b/sections/futures/OrderSizing/OrderSizing.tsx
@@ -113,7 +113,7 @@ const OrderSizing: React.FC<OrderSizingProps> = ({ disabled }) => {
 
 			<CustomInput
 				disabled={isDisabled}
-				right={marketAsset || 'sUSD'}
+				right={marketAsset?.replace(/^s/, '') || 'sUSD'}
 				value={assetValue}
 				placeholder="0.0"
 				onChange={onChangeAssetValue}


sections/futures/OrderSizing/OrderSizing.tsx
```diff
@@ -113,7 +113,7 @@ const OrderSizing: React.FC<OrderSizingProps> = ({ disabled }) => {
 
 			<CustomInput
 				disabled={isDisabled}
-				right={marketAsset || 'sUSD'}
+				right={marketAsset?.replace(/^s/, '') || 'sUSD'}
```
koredefashokun, line 116: Maybe we could use the `getDisplayAsset` function from `utils/futures` for consistency.

koredefashokun (commented):
Looks good. Just added a small comment to make sure code is consistent.

sections/futures/OrderSizing/OrderSizing.tsx
0xwalde: Oh duh, thanks! I was looking for something like this and completely missed it 🤦

platschi (approved):

PR #1391, opened 2022-09-15
Fix funding display

Tburm:
Fixes a display issue with the hourly funding rate in `MarketDetails`.

Expected behavior is:
* Display 1H while the page loads
* If the 1H funding rate query fails, display the instant funding rate from the contract

## Description
The order that we retrieve data is causing a display bug in `MarketDetails` where it flashes the instant funding rate before displaying the hourly funding rate. The instant rate is used as a backup if the hourly query fails. This change checks if the funding query has a value of `null` before displaying the instant rate, which makes sure the query was attempted and failed.

## How Has This Been Tested?
* Forced null values for 1 market
* Forced null values for all markets

Both situations behaved as expected.

### DIFF
--- a/queries/futures/useGetAverageFundingRateForMarkets.ts
+++ b/queries/futures/useGetAverageFundingRateForMarkets.ts
@@ -4,9 +4,10 @@ import request, { gql } from 'graphql-request';
 import { useQuery, UseQueryOptions } from 'react-query';
 import { useRecoilValue, useSetRecoilState } from 'recoil';
 
+import { Period, PERIOD_IN_SECONDS } from 'constants/period';
 import QUERY_KEYS from 'constants/queryKeys';
 import Connector from 'containers/Connector';
-import { fundingRatesState, futuresMarketsState } from 'store/futures';
+import { fundingRatesState, futuresMarketsState, marketAssetsState } from 'store/futures';
 import { FuturesMarketKey, MarketKeyByAsset } from 'utils/futures';
 import logError from 'utils/logError';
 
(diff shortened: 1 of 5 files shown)


platschi:
Works fine on Optimism. When switching network to OP Goerli, it shows Instant again until I refresh or go to dashboard and return to a market. On L1, shows Instant only. 

Not sure it's a blocker. When switching back to L2 Optimism, 1h Funding shows correctly again.

platschi (dismissed):
Works now as expected 👍🏼

sections/futures/MarketDetails/useGetMarketData.ts
```diff
@@ -45,14 +45,15 @@ const useGetMarketData = (mobile?: boolean) => {
 
 	const pastPrice = pastRates.find((price) => price.synth === marketAsset);
 
-	const fundingTitle = React.useMemo(
-		() => `${!fundingRate?.fundingRate && !!marketInfo ? 'Inst.' : '1H'} Funding Rate`,
-		[fundingRate, marketInfo]
-	);
+	const fundingTitle = React.useMemo(() => `${fundingRate?.fundingTitle ?? '1H Funding Rate'}`, [
```
avclarke, line 48: Should probably add this to translations

avclarke (commented):
Looks good, just a label we should probably add to translations while making a change there.

sections/futures/MarketDetails/useGetMarketData.ts
Tburm: agreed, I updated this

queries/futures/useGetAverageFundingRateForMarkets.ts
```diff
@@ -46,8 +49,12 @@ const useGetAverageFundingRateForMarkets = (
 		}
 	);
 
+	const periodLength = PERIOD_IN_SECONDS[period];
+
+	const periodTitle = period === Period.ONE_HOUR ? '1H Funding Rate' : 'Funding Rate';
```
avclarke, line 54: And noticed another place where we should probably translate

Tburm: I fixed this, and another for the instantaneous funding rate. That should be all (except the keys, which need to be refactored entirely)

avclarke (approved):

Issue #1393, opened 2022-09-16
Replace tippy with tooltip component


platschi:
### Subject of the issue
Kwenta currently uses two different ways to show tooltips, one being the `tippyjs/react´ package and the tooltip component. To reduce dependencies, we should remove all instances of tippy and replace them with our tooltip component.

### Deliverables

- [ ] Remove `tippyjs/react` package
- [ ] Replace tooltip instances using tippy with tooltip component
- [ ] Make sure new tooltips are displayed correctly within the modals)

### Bounty
150 $OP

PR #1399, opened 2022-09-18
Replace tippy with tooltip component

leovct:
## Description
Kwenta currently uses two different ways to show tooltips, one being the `tippyjs/react` package and the tooltip component. To reduce dependencies, we should remove all instances of tippy and replace them with our tooltip component.

- [x] Remove tippyjs/react package
- [x] Remove old `tippy-content` CSS class
- [x] Remove `GasPriceSummaryItem` component because it wasn't used anywhere
- [x] Replace tooltip instances using tippy with tooltip component
- [x] Add new field: `visible: boolean` to Tooltip component
- [x] Remove `arrow`, `trigger` and `interactive`properties
- [x] Make sure new tooltips are displayed correctly within the modals)

## Related issue
#1393


sections/shared/modals/TxConfirmationModal/TxConfirmationModal.tsx
```diff
@@ -223,22 +220,20 @@ export const TxConfirmationModal: FC<TxConfirmationModalProps> = ({
 								components={[<NoTextTransform />]}
 							/>
 							&nbsp;
-							<StyledTooltip
-								placement="top"
+							<CustomStyledTooltip
+								preset="top"
```
platschi, line 224: Make sure this tooltip stays within the borders of the modal. Also it should not disappear as soon as a user tries to move the mouse away from the (i) icon so that they can click the actual "Learn more" link :)

platschi: Example:

Screen Shot 2022-09-19 at 10 07 45

sections/shared/modals/TxConfirmationModal/TxConfirmationModal.tsx
```diff
@@ -181,13 +180,11 @@ export const TxConfirmationModal: FC<TxConfirmationModalProps> = ({
 										]}
```
platschi, line 180: Not sure the `ExternalLink` is still necessary for the `exchange-fee-hint`?

sections/shared/components/FeeRateSummary/FeeRateSummary.tsx
```diff
@@ -35,10 +35,8 @@ const FeeRateSummaryItem: FC<FeeRateSummaryItemProps> = memo(({ totalFeeRate, ba
 							<>
 								<DynamicFeeLabel>+</DynamicFeeLabel>
 								<DynamicFeeRateTooltip
+									preset="bottom"
```
platschi, line 38: `DynamicFeeRateToolTip` does not work

platschi (dismissed):
Works as expected, one dependency less :+1:

LeifuChen (dismissed):
Looks good to me.
The only concern is that now the tooltip couldn't stay a while for the user to interact with the link (https://atomiks.github.io/tippyjs/v6/all-props/#interactive). However, I agree that it's out of the scope of this ticket, and it seems unnecessary as we don't have any critical info to show via the link.

sections/shared/components/FeeRateSummary/FeeRateSummary.tsx
leovct: It should be fixed now!

sections/shared/modals/TxConfirmationModal/TxConfirmationModal.tsx
leovct: It has been fixed also :)

platschi (approved):

Issue #1402, opened 2022-09-19
Add Tally Wallet to Rainbowkit


platschi:
Tally Wallet now supports Optimism since v0.16.1

### Deliverables
- Add Tally Wallet to rainbowkit as an option to connect to

PR #1407, opened 2022-09-20
Added tally ho to rainbowkit

LeifuChen:
## Description
Add `Tally Ho` to the wallet list.
> Note that `Tally Ho` needs to be set as a default wallet in the setting.
image

## Related issue
#1402

## How Has This Been Tested?
1. Open the page and click on the `connect wallet`
2. Select `Tally Ho` and confirm on the pop-up

image


components/Rainbowkit/Tally/Tally.ts
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const Tally = ({ chains, shimDisconnect }: TallyOptions): Wallet => ({
+	id: 'tally',
+	iconBackground: '#FFF',
+	name: 'Tally Ho',
+	iconUrl: async () => TallyIcon,
+	downloadUrls: {
+		android: 'https://play.google.com/store/apps/details?id=io.gnosis.Tally',
+		ios: 'https://apps.apple.com/us/app/gnosis-Tally/idid1515759131',
```
platschi, line 18: These links don't work, should probably set them to empty strings as Tally does not have Android/iOS apps?

platschi (changes requested):

platschi (commented):
Works good with Tally Ho in Brave Browser. Two points:

- Switching to Optimistic Goerli / Goerli does not work (looks like Tally does not support them?) - no blocker though for mainnet usage, just an observation
- The round Tally logo makes it look not in line with the other logos in the rainbow onboarding modal. Can we update it, maybe change the white background of the logo to match the Tally bg color?

components/Rainbowkit/Tally/Tally.ts
LeifuChen: Yes. Good catch. I will update these links.

Tburm:
Pushing a commit to fix the icon. We needed to specify a background color @LeifuChen @platschi

components/Rainbowkit/Tally/Tally.ts
Tburm: I pushed a commit to remove these links

Tburm (dismissed):

LeifuChen:
> Pushing a commit to fix the icon. We needed to specify a background color @LeifuChen @platschi

@Tburm Thanks for the fix.

components/Rainbowkit/Tally/Tally.ts
LeifuChen: @Tburm I added the link to `Tally Ho`'s chrome browser extension.  
image

Tburm (approved):

platschi:
Perfect, works like a charm now!

platschi (approved):

PR #1411, opened 2022-09-20
Improve Trading Volume

Tburm:
Making our trading volume query more efficient by avoiding some aggregation. Moves 24 hr volumes to a state variable.

## Description
* Create state variable for `futuresVolumes` storing 24h volume
* Uses state in all components
* Removes hooks to pull single market volume

## Related issue
#1220


koredefashokun (approved):
Looks good. Nice that we can reduce the number of queries.

PR #1414, opened 2022-09-20
Improve Rates

Tburm:
Get exchange rates from the `ratesState` variable everywhere it is needed. The query will be called from both the `ExchangeContext` and the `RefetchContext`.

## Description
* Update all instances of `useExchangeRatesQuery` to use the state variable

## Related issue
* #1220

### DIFF
--- a/hooks/useExchange.ts
+++ b/hooks/useExchange.ts
@@ -50,6 +50,7 @@ import {
 	sourceCurrencyKeyState,
 	txErrorState,
 } from 'store/exchange';
+import { ratesState } from 'store/futures';
 import { ordersState } from 'store/orders';
 import { hasOrdersNotificationState, slippageState } from 'store/ui';
 import { gasSpeedState } from 'store/wallet';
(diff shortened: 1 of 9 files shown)


platschi (commented):
Guess we can remove `queries/rates/useExchangeRatesQuery` if it's replaced everywhere?

koredefashokun (approved):

Tburm:
@platschi we can't remove it, it's still called from `useExchange` and `RefetchContext`. Replacing the queries elsewhere makes sure it is only called once, then everything points at state.

hooks/useExchange.ts
```diff
@@ -117,6 +118,7 @@ const useExchange = ({ showNoSynthsCard = false }: ExchangeCardProps) => {
 	const setHasOrdersNotification = useSetRecoilState(hasOrdersNotificationState);
 	const quoteCurrencyAmountBN = useRecoilValue(quoteCurrencyAmountBNState);
 	const baseCurrencyAmountBN = useRecoilValue(baseCurrencyAmountBNState);
+	useExchangeRatesQuery({ refetchInterval: 15000 });
```
avclarke, line 121: I know this is already merged but noticed it as looking through for something unrelated and I'm not sure we need this extra query do we, as it's already called in RefetchProvider which wraps the whole app?

Issue #1405, opened 2022-09-20
Disable Deposit Margin / Withdraw Margin / Confirm Order / Close Position buttons when Unsupported Network


awbvious:
Environment:

- Brave browser
- Metamask extension

Steps to reproduce:

1. have an open sETH position on Optimism
2. Go to https://kwenta.io/market/?asset=sETH
3. Switch to Optimism (if needed) and Connect Wallet (if needed)
4. Open another tab and go to https://app.uniswap.org/#/swap in tab
5. Connect Wallet (if needed)
6. Switch to Polygon via Uniswap widget in top right corner (confirm chain switch in Metamask)
7. Go back to tab for https://kwenta.io/market/?asset=sETH (top corner will display "Switch Networks" and "Unsupported Network")
8. All buttons will still be available/active, such as the Deposit, Withdraw, Modify, Close
9. Clicking any button from step 8 will still show pop-up with available/active button Deposit Margin, Withdraw Margin,  Confirm Order, Close Position--respectively
10. Clicking any button from step 9 will open Metamask to try to Submit the transaction with Matic
11. Click Submit
12. "Your transaction has been sent to the network" will show in the bottom of Kwenta
13. Transaction will be "Success" in polygonscan and cost the user Matic, but will not affect user's position or margin balance

The problem is when you change the network while the tab is active, it works fine.  The wallet disconnects, then the user is presented with connect wallet pop-up, and users can't accidentally submit.  But the UI/UX is not prepared for changing the network while the tab is not active: The tab is Kwenta left open, the active tab changes, and then the chain is changed in wallet to an unsupported network, and then the Kwenta tab is made active again.

The solution, I suggest, is either:

a) Make inactive / disabled the buttons from Step 9 and any other button that requires submitting a transaction from Metamask.  Replace the text the button from Deposit Margin, Withdraw Margin,  Confirm Order, Close Position, etc. to "Unsupported Network".

b) Replace the button from Step 9, or any other button that requires submitting a transaction, with "Switch Networks" button that works identically to the button from Step 7.

My preference is for B, but either would be an improvement over the current behavior which could waste gas and time and cause confusion.

PR #1416, opened 2022-09-22
Disabled the trade panel if Optimism is not connected

LeifuChen:
## Description
Disable the trade panel if the wallet is not connected or the network is unsupported

## Related issue
#1405

## Motivation and Context
Make sure the user can only interact with the trade panel when connecting the optimism mainnet or optimism goerli.

## How Has This Been Tested?
Case 1: No wallet connected
1. Visit the futures page
2. The trade panel is replaced by the message `connect wallet`
3. Click on the `connect wallet`
4. The trade panel is visible after connecting the wallet

Case 2: Connected to Polygon
1. Visit the futures page 
2. The trade panel is replaced by the message `Switch to L2`
3. Click on the `Switch to L2`
4. The trade panel is visible after switching to L2

image
image


platschi:
Works fine so far, just one small nitpick (button colors should be responsive to theme:)

Screen Shot 2022-09-23 at 09 02 28

avclarke (commented):
Looks good, the only other thing I was thinking is we can also replace the `CrossMarginUnsupported` with this so we have a single component to handle this across both futures types.

sections/futures/Trade/FuturesUnsupported.tsx
```diff
@@ -0,0 +1,63 @@
...
+			)}
+		</MessageContainer>
+	);
+};
+
+const UnsupportedMessage = styled.div`
+	margin-top: 12px;
+`;
+
+const IsolatedLink = styled.div`
```
avclarke, line 42: I think we can rename this to just Link or something more generic.

LeifuChen: We use `Link` from `next/link` in the code repo. So maybe name this as `LinkContainer`? 
image

avclarke: Yea sounds good

LeifuChen: Done.

LeifuChen:
> Looks good, the only other thing I was thinking is we can also replace the `CrossMarginUnsupported` with this so we have a single component to handle this across both futures types.

Sounds like a good idea. @platschi do you want to have a more generic component in the new PR or in this PR?

avclarke (approved):

avclarke:
> > Looks good, the only other thing I was thinking is we can also replace the `CrossMarginUnsupported` with this so we have a single component to handle this across both futures types.
> 
> Sounds like a good idea. @platschi do you want to have a more generic component in the new PR or in this PR?

I don't mind updating the cross margin part actually as I have a couple of ideas to improve the state there in general.

PR #1426, opened 2022-09-26
Disable all the info of trade tabs on the unsupported network

LeifuChen:
## Description
There is no need to show any trade info by the given wallet address if the user is connected to the wrong network.

## Motivation and Context
Make the current network much clearer to users.

## How Has This Been Tested?
Case 1. Connect to the `Ethereum` and go to the `Futures` page.
All the tabs under the Trading View Chart shows `Perpetual futures are available on Optimism L2` with the link `Switch to L2`

Case 1. Connect to the `Optimism` and go to the `Futures` page.
All the tabs under the Trading View Chart shows either the data or `You have no xx` in the table

image
image


platschi (approved):
This works fine and should help with pointing new users to Optimism. 

One thing I noticed is that the "Spot History" (under Dashboard -> History) persists when switching to unsupported networks, we should make sure to catch this and also show an unsupported network message, though the logic is slightly different as spot is available on L1 and L2. Will ticket this out separately.

Regarding this ticket, all good.

Tburm (approved):
Approving but discovered another (unrelated) issue.

When I try to use this switch to L2 button in Brave it throws an error. The issue is that the `@synthetixio/optimism-networks` call to switch networks does not include a currency of ETH and instead adds the currency `UNKNOWN`. The parameters [here](https://github.com/Synthetixio/js-monorepo/blob/master/packages/optimism-networks/src/constants.ts#L19) do not include it
image

PR #1434, opened 2022-09-29
Position state update

Tburm:
Update how we store position and portfolio data in state. This PR mostly splits isolated from cross margin position data, allowing us to easily access a particular account type.

## Description
* Create `positionHistory` state value to hold subgraph data for positions
* Create `portfolio` selector to calculate portfolio value from other state values
* Update `positions` state value to an object containing isolated and cross margin positions.
* Display isolated and cross margin positions in the leaderboard trader history
* Adds a second tab to the dashboard to display isolate and cross margin positions separately
* Adds cross margin positions to the balance dropdown in the nav


platschi:
Awesome. Testing on OP Goerli (see screenshot):

- [ ] Portfolio value does not count up
- [ ] Isolated Margin shows a "2" badge however only 1 isolated market is traded

Screen Shot 2022-10-03 at 18 09 45

queries/futures/useGetFuturesPositionForMarkets.ts
```diff
@@ -28,24 +28,21 @@ const useGetFuturesPositionForMarkets = (options?: UseQueryOptions<FuturesPositi
...
-	const selectedFuturesAddress = useRecoilValue(selectedFuturesAddressState);
-
+	const [positions, setPositions] = useRecoilState(positionsState);
+	const { walletAddress, crossMarginAddress, crossMarginAvailable, status } = useRecoilValue(
+		futuresAccountState
+	);
 	const assets = futuresMarkets.map(({ asset }) => asset);
 
-	return useQuery<FuturesPosition[] | []>(
+	return useQuery<any>(
```
avclarke, line 38: If we're not returning anything I think we should change to undefined rather than any.

queries/futures/useGetFuturesPositionForAccount.ts
```diff
@@ -1,54 +1,108 @@
...
 				logError(e);
-				return null;
+				if (!account) setPositionHistory(DEFAULT_POSITION_HISTORY);
+				return DEFAULT_POSITION_HISTORY;
 			}
 		},
 		{
-			enabled: !!selectedFuturesAddress,
-			refetchInterval: 5000,
+			enabled: !!walletAddress,
```
avclarke, line 105: Having this condition means the state reset on line 32 won't be called as the query will be disabled when no wallet, I think we can just remove this check. I ended up doing this for a couple of other queries to ensure state clears when wallet disconnected.

queries/futures/useGetFuturesPositionForMarkets.ts
```diff
@@ -28,24 +28,21 @@ const useGetFuturesPositionForMarkets = (options?: UseQueryOptions<FuturesPositi
...
 		QUERY_KEYS.Futures.MarketsPositions(
 			network?.id as NetworkId,
 			assets || [],
-			selectedFuturesAddress ?? ''
+			walletAddress ?? '',
+			crossMarginAddress ?? ''
 		),
 		async () => {
-			if (!assets || (!provider && !l2Provider) || !selectedFuturesAddress) {
-				setFuturesPositions([]);
```
avclarke, line 45: I think we'd still want to clear state when wallet disconnected

queries/futures/useGetFuturesPositionForMarkets.ts
```diff
@@ -28,24 +28,21 @@ const useGetFuturesPositionForMarkets = (options?: UseQueryOptions<FuturesPositi
...
-				return [];
-			}
-
 			await ethCallProvider.init(isL2 ? provider : l2Provider);
```
avclarke, line 46: Not related to your changes but do we need this conditional here as shouldn't this query only be made when connected to l2? I think we should try and keep the `l2Provider` only for use on the landing page if possible.

queries/futures/useGetFuturesPositionForMarkets.ts
```diff
@@ -57,34 +54,64 @@ const useGetFuturesPositionForMarkets = (options?: UseQueryOptions<FuturesPositi
...
+					const asset = assets[ind];
+					return mapFuturesPosition(position, canLiquidate, asset);
+				})
+				.filter(({ remainingMargin }) => remainingMargin.gt(0));
+
+			const crossPositions =
+				crossMarginAvailable && crossMarginAddress
+					? futuresMarkets
+							.map((futuresMarket: FuturesMarket, ind: number) => {
+								const position = positionDetails[futuresMarkets.length + ind];
```
avclarke, line 99: This was a bit confusing to follow, I wonder if it might be better to separate isolatedMargin and crossMargin position calls.

queries/futures/utils.ts
```diff
@@ -290,92 +294,83 @@ export const mapMarginTransfers = (
...
+			feesPaid,
+			netFunding,
+			pnlWithFeesPaid,
+			netTransfers,
+			totalDeposits,
+			entryPrice,
+			avgEntryPrice,
+			exitPrice,
+		}: FuturesPositionResult) => {
+			const entryPriceWei = weiFromWei(entryPrice).div(ETH_UNIT);
```
avclarke, line 329: How come we need to divide by ETH in these conversions? `weiFromWei` is a util for converting a BigNum wei to Synthetix wei, e.g. so BigNumber('1000000000000000000') gets set as wei('1')

sections/dashboard/Overview/Overview.tsx
```diff
@@ -24,62 +20,56 @@ import PortfolioChart from '../PortfolioChart';
...
 	const [activeMarketsTab, setActiveMarketsTab] = useState<MarketsTab>(MarketsTab.FUTURES);
 
-	const totalSpotBalancesValue = formatDollars(wei(synthBalances?.totalUSDBalance ?? zeroBN));
-
-	const totalFuturesPortfolioValue = formatDollars(wei(portfolioValue ?? zeroBN));
-
-	const POSITIONS_TABS = useMemo(
-		() => [
+	const POSITIONS_TABS = useMemo(() => {
+		const crossPositions = positions.cross_margin.filter(({ position }) => !!position).length;
```
avclarke, line 42: Do you think it would be better to filter out null positions at the source or we need to also show market positions with empty positions in places?

queries/futures/useGetFuturesPositionForAccount.ts
```diff
@@ -1,54 +1,108 @@
...
+import { getFuturesEndpoint, mapFuturesPositions } from './utils';
 
-const useGetFuturesPositionForAccount = (options?: UseQueryOptions<any>) => {
+const DEFAULT_POSITION_HISTORY: PositionHistoryState = {
+	[FuturesAccountTypes.CROSS_MARGIN]: [],
+	[FuturesAccountTypes.ISOLATED_MARGIN]: [],
+};
+
+const useGetFuturesPositionForAccount = (
+	account?: string,
```
avclarke, line 20: Having this optional account and conditional state setting feels a bit off, I think ideally we avoid setting state in this query and have a higher level wrapper query that calls this with users wallet and sets positionHistory state.

avclarke (changes requested):
Nice improvements! Just a few comments and suggestions

queries/futures/useGetFuturesPositionForAccount.ts
```diff
@@ -1,54 +1,108 @@
...
+import { FuturesAccountTypes, PositionHistory, PositionHistoryState } from './types';
+import { getFuturesEndpoint, mapFuturesPositions } from './utils';
 
-const useGetFuturesPositionForAccount = (options?: UseQueryOptions<any>) => {
+const DEFAULT_POSITION_HISTORY: PositionHistoryState = {
+	[FuturesAccountTypes.CROSS_MARGIN]: [],
+	[FuturesAccountTypes.ISOLATED_MARGIN]: [],
+};
+
+const useGetFuturesPositionForAccount = (
```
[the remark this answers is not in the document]
Tburm: I still like `positionHistory` for subgraph data and `positions` or `futuresPositions` for contract data. The `positionHistory` object stores more than just the open positions, and stores some historical data like avg entry, open timestamp, and summary of margin transfers.

I will update all of these to use the `positionHistory` naming and leave `futuresPositions` type naming for raw contract data.

queries/futures/useGetFuturesPositionForMarkets.ts
Tburm: I think this is a good solution when we move to redux. I opted out of that solution here because they both need to call before the dashboard loads, but if they're running at the same time the can't write to the same object in recoil because of a race condition. 

The other option I considered is to call them separately and store on separate objects in state, then combine using a selector. This avoids the race condition, but makes it so we need to duplicate the query and adds n+1 state variables for the number of futures market types we have.

sections/dashboard/Overview/Overview.tsx
Tburm: We need to keep those positions to total the remaining margin. This handles situations where you have margin but no open position. They are included so we can total margin, but not count them as a position on the badge.

queries/futures/useGetFuturesPositionForMarkets.ts
avclarke: I was thinking more just within this query to separate them in terms of calls, so we could have `isolatedPositionCalls`, `crossPositionCalls`, which would then allow to map from `isolatedPositionDetails` and `crossPositionDetails`, rather than mapping from futuresMarkets both times with offset index. Could then become:

```
const isolatedPositions = isolatedPositionDetails
    .map((position, ind) => {
        const canLiquidate = canLiquidateStateIsolated[ind];
        const asset = assets[ind];
        return mapFuturesPosition(position, canLiquidate, asset);
    })
    .filter(({ remainingMargin }) => remainingMargin.gt(0));

const crossPositions = crossPositionDetails
    .map((position, ind) => {
        const canLiquidate = canLiquidateStateCross[ind];
        const asset = assets[ind];
        return mapFuturesPosition(position, canLiquidate, asset);
    })
    .filter(({ remainingMargin }) => remainingMargin.gt(0));
```

More just a readability thing as it wasn't very clear when I first followed it but definitely not essential.

queries/futures/useGetFuturesPositionForAccount.ts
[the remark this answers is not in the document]
avclarke: Sounds good

queries/futures/useGetFuturesPositionForMarkets.ts
Tburm: Need to return `void`, but good call

queries/futures/useGetFuturesPositionForAccount.ts
Tburm: This will be fixed now with the "two query" setup. The one that sets state runs a `useEffect` if the query changes, and if the query data is not available it resets to the default with no positions. This allows us to pass the `walletAddress` parameter which might be null, but know that nothing will run and the state will be reset if this happens.

queries/futures/useGetFuturesPositionForAccount.ts
```diff
@@ -1,54 +1,108 @@
...
+
+						// make sure it's not a duplicate before adding to the list
+						const existingPositionId = acc[accountType].findIndex((pos) => pos.id === position.id);
+						if (
+							existingPositionId === -1 ||
+							position.timestamp > acc[accountType][existingPositionId].timestamp
+						)
+							acc[accountType] = [
+								position,
+								...acc[accountType].filter(({ id }) => id !== existingPositionId),
```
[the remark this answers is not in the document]
Tburm: We don't need this check at all since we start with the default. I think this was left over from when I was reducing with existing positions.

queries/futures/useGetFuturesPositionForAccount.ts
Tburm: Fixed this by splitting into two queries. One is a regular `react-query` which fetches the data for a specified account. The other calls that query, and has a `useEffect` to set the state value when the data changes.

queries/futures/useGetFuturesPositionForAccount.ts
```diff
@@ -1,54 +1,108 @@
...
+				const positionHistory = response ? mapFuturesPositions(response) : [];
+
+				const positionHistoryByType = positionHistory.reduce(
+					(acc: PositionHistoryState, position: PositionHistory) => {
+						const accountType = position.accountType;
+
+						// make sure it's not a duplicate before adding to the list
+						const existingPositionId = acc[accountType].findIndex((pos) => pos.id === position.id);
+						if (
+							existingPositionId === -1 ||
```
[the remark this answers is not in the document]
Tburm: This was all unnecessary based on another comment, basically removed the whole check instead of fixing this.

queries/futures/useGetFuturesPositionForMarkets.ts
Tburm: I added a function to clear and updated the `enabled` parameter

queries/futures/useGetFuturesPositionForMarkets.ts
Tburm: This makes sense. I'll slice the array we get back and rename them, but don't want to split the calls up because it will double our RPC calls.

queries/futures/utils.ts
Tburm: Looks like this is because the original value is a `Wei` value with a precision of 0. So the function just returns the same value again, with a new precision of 18.

To get the "real" value we need to divide here, do you think that means we should update the `weiFromWei` function to handle this when precision values are different?

avclarke: I just tried here changing this to only `weiFromWei(entryPrice)` and it seemed to work as expected. The `weiFromWei` func is really just a helper for the `new Wei(entryPrice, 18, true)` calls that were replaced. Unless there's some other change that impacts these as I was only trying on dev.

koredefashokun (approved):

queries/futures/utils.ts
Tburm: That's not what I'm seeing. When I remove the divisor I get this number value for the avg entry price: `2.0122782e+22`
image

Tburm: This function _does_ work for orders, where the values returned from the query are a string. This query returns `Wei` values with a precision of 0 instead of 18, so they are simply converted to `Wei` with a precision of 18 without doing the conversion.

Tburm: I just added an if statement to that function to check if the input is already a `Wei` value. If it is, we convert from the given precision to 18.

queries/futures/useGetFuturesPositionForMarket.ts
```diff
@@ -53,6 +55,7 @@ const useGetFuturesPositionForMarket = (options?: UseQueryOptions<FuturesPositio
 			return position;
 		},
 		{
+			enabled: isL2,
```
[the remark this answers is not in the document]
Tburm: That's right, I fixed this and added the condition to clear on L1.

avclarke (approved):

koredefashokun (approved):

Issue #1339, opened 2022-08-30
Add pagination to Transfers tab table


platschi:
Currently the "Transfers" tab table on Futures Markets does not include pagination, instead showing an endless list of deposits and withdrawals. This should be reduced to show maximum 5 rows with deposit/withdraw information per page and include pagination, to match the design of other tabs (such as Trades) on desktop.

### How it is

![Image](https://user-images.githubusercontent.com/548702/193340364-d412231c-a5a7-4e45-a68c-14e4bc5e0613.png)

### How it should be

![Image](https://user-images.githubusercontent.com/548702/193340480-9ed3bd36-f2ba-4610-9bd2-3bc365aa4fb2.png)

### Deliverables
- Add pagination to `Transfers` table
- Limit rows to 5 

Note that this is already the case on mobile, so this ticket only affects desktop view.

### Bounty
80 $OP

PR #1442, opened 2022-10-01
Paginate Transfers table with 5 rows per page

0xwalde:
Adds pagination to the Transfers table to match the behavior of the other tables, with 5 rows per page.

Also added the `highlightRowsOnHover` prop to the Trades table since the other 3 tabs already had it and it sounded like the goal here was consistency between tabs.

## Related issue
https://github.com/Kwenta/kwenta/issues/1339

## How Has This Been Tested?
Tested locally against Optimism Goerli. 

Kwenta-1339

### DIFF
--- a/sections/futures/Trades/Trades.tsx
+++ b/sections/futures/Trades/Trades.tsx
@@ -56,6 +56,7 @@ const Trades: React.FC<TradesProps> = ({ history, isLoading, isLoaded, marketAss
 	return (
 		<Card>
 			<StyledTable
+				highlightRowsOnHover
 				columns={[
 					{
 						Header: (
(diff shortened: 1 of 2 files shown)


sections/futures/Trades/Trades.tsx
```diff
@@ -56,6 +56,7 @@ const Trades: React.FC<TradesProps> = ({ history, isLoading, isLoaded, marketAss
 	return (
 		<Card>
 			<StyledTable
+				highlightRowsOnHover
```
platschi, line 59: Nice catch :)

platschi (approved):
Works great

PR #1444, opened 2022-10-03
New state management and SDK #1

koredefashokun:
## Description
This PR sets up the new architecture for the Kwenta frontend, which includes:

- Moving state management to Redux
- Abstracting application logic to an `sdk` folder, which will be moved into a standalone package in the near future

## Related issue
N/A

## Motivation and Context
The new architecture solves a myriad of performance problems with the existing application, separates presentation and business logic, and makes it easier to refactor existing components and add new features.

## How Has This Been Tested?
N/A

N/A


sdk/data/synths.ts
```diff
@@ -0,0 +1,236 @@
...
+			.filter(booleanTypeGuard)
+	);
+
+const mainnetSynths = synthsByNetwork(1);
+const optimismSynths = synthsByNetwork(10);
+const goerliSynths = synthsByNetwork(5);
+const optimismGoerliSynths = synthsByNetwork(420);
+
+export const getSynthsForNetwork = (networkId: NetworkId) => {
+	switch (NetworkNameById[networkId]) {
```
avclarke, line 224: Would it make more sense to just do
```
const synths = synthsByNetwork(networkId);
if (!synths.length) throw new Error('We do not support synths on the selected network.');
return synths

```

sdk/contracts/index.ts
```diff
@@ -0,0 +1,134 @@
...
+	| 'Synth';
+
+export type ContractDetails = {
+	name: ContractName;
+	networks: Partial<Record<NetworkId, { address: string; contract: Contract }>>;
+};
+
+export type AllContractsMap = Record<ContractName, Partial<Record<NetworkId, Contract>>>;
+
+export type ContractMap = Partial<Record<ContractName, Contract>>;
```
avclarke, line 36: Will be important the contract map includes the specific types for each one once the typechain contracts are implemented as commented above.

sdk/common/balances.ts
```diff
@@ -0,0 +1,47 @@
...
+import { wei } from '@synthetixio/wei';
+import { ethers } from 'ethers';
+import { orderBy } from 'lodash';
+
+import { notNill } from 'queries/synths/utils';
+import { zeroBN } from 'utils/formatters/number';
+
+type SynthBalancesTuple = [string[], ethers.BigNumber[], ethers.BigNumber[]];
+
+export const getSynthBalances = async (walletAddress: string, contracts: ContractsMap) => {
```
avclarke, line 12: I was thinking this could also be a class / service similar to exchange / futures etc, as it will also be api's we'd like to expose to anyone using the sdk rather than internal methods that need to be passed contract instances.

avclarke (commented):
Looking great so far! I dropped a few comments and suggestions but ignore them if you just haven't got to something yet as appreciate it's still WIP.

sdk/exchange.ts
```diff
@@ -0,0 +1,793 @@
...
+
+import { getSynthBalances } from './common/balances';
+import type { ContractMap } from './contracts';
+import SynthRedeemerABI from './contracts/abis/SynthRedeemer.json';
+import { getSynthsForNetwork, SynthsMap, SynthSymbol } from './data/synths';
+
+type CurrencyRate = ethers.BigNumberish;
+type SynthRatesTuple = [string[], CurrencyRate[]];
+
+const protocols =
```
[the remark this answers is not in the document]
koredefashokun: Good catch. Will fix.

sdk/exchange.ts
```diff
@@ -0,0 +1,793 @@
...
+			throw new Error('Exchanger does not exist on the currently selected network.');
+		}
+
+		return await this.contracts.Exchanger.feeRateForExchange(
+			ethers.utils.formatBytes32String(sourceCurrencyKey),
+			ethers.utils.formatBytes32String(destinationCurrencyKey)
+		);
+	}
+
+	private isCurrencyETH(currencyKey: string) {
```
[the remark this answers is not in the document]
koredefashokun: Okay. Just switched it to do private before public, but I'll invert this.

sdk/contracts/index.ts
```diff
@@ -0,0 +1,134 @@
+import { NetworkId, NetworkNameById } from '@synthetixio/contracts-interface';
```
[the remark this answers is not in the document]
koredefashokun: Yeah. I'm going to create native types for `NetworkId` and the other types we get from synthetix.

sdk/contracts/index.ts
```diff
@@ -0,0 +1,134 @@
...
+import { Contract } from 'ethers';
```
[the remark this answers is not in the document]
koredefashokun: Yeah, we can. Just trying to get these to be functional first though. Will add that once the new exchange is up and running.

sdk/contracts/index.ts
```diff
@@ -0,0 +1,134 @@
...
+	networks: Partial<Record<NetworkId, { address: string; contract: Contract }>>;
+};
+
+export type AllContractsMap = Record<ContractName, Partial<Record<NetworkId, Contract>>>;
+
+export type ContractMap = Partial<Record<ContractName, Contract>>;
+
+export const contracts: AllContractsMap = {
+	Exchanger: {
+		1: new Contract('0xD64D83829D92B5bdA881f6f61A4e4E27Fc185387', ExchangerABI),
```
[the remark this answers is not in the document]
koredefashokun: That's interesting. In the future, we will still need to copy the addresses into a file somehow, so it might be better to just have them all in one place. We can definitely talk more about this when we start refining this, though.

sdk/data/synths.ts
koredefashokun: I considered doing that. However, this way is slightly more performant since we actually populate the synths for each network at build time. If we do it that way, we would be running the function that filters the contracts for each network everytime the network changes while the app is running.

sdk/contracts/index.ts
koredefashokun: For sure. This will be added once we ascertain that everything works, and we start cleaning stuff up and adding DX improvements.

sdk/contracts/index.ts
```diff
@@ -0,0 +1,134 @@
...
+		69: new Contract('0xbdb2Bf553b5f9Ca3327809F3748b86C106719C95', SynthABI),
+	},
+};
+
+export const booleanTypeGuard = <T>(x: T | null): x is T => Boolean(x);
+
+const contractsByNetwork = (id: NetworkId): ContractMap =>
+	Object.fromEntries(
+		Object.entries(contracts)
+			.map(([name, config]) => {
```
avclarke, line 100: It's a bit cleaner to use `reduce` for these types of mapping rather than filtering out nulls after.

sdk/data/synths.ts
```diff
@@ -0,0 +1,236 @@
...
+// SNX
+// 1 - 0xC011a73ee8576Fb46F5E1c5751cA3B9Fe0af2a6F
+// 5 - 0x51f44ca59b867E005e48FA573Cb8df83FC7f7597
+// 10 - 0x8700dAec35aF8Ff88c16BdF0418774CB3D7599B4
+// 420 - 0x2E5ED97596a8368EB9E44B1f3F25B2E813845303
+
+const synthsByNetwork = (id: NetworkId): Partial<SynthsMap> =>
+	Object.fromEntries(
+		Object.entries(synths)
+			.map(([symbol, config]) => {
```
avclarke, line 199: Same here I think reduce would be more appropriate

sdk/data/synths.ts
avclarke: Yea that makes sense, just not sure we need the explicit name mapping and also not sure we should throw an error when not returning a promise, I think we can just return an empty map, so could maybe be like:

```
ALL_SYNTHS = {
  1: synthsByNetwork(1);
  10: synthsByNetwork(10);
  ...
}

export const getSynthsForNetwork = (networkId: NetworkId) => {
  return ALL_SYNTHS[networkId] ?? {};
}
```

platschi:
Vercel deploy error message:

Screen Shot 2022-10-06 at 12 36 08

pages/_app.tsx
```diff
@@ -112,15 +127,21 @@ const App: FC<AppProps> = (props) => {
...
-				<QueryClientProvider client={new QueryClient()}>
-					<WagmiConfig client={wagmiClient}>
-						<WithAppContainers>
-							<InnerApp {...props} />
-						</WithAppContainers>
-					</WagmiConfig>
-				</QueryClientProvider>
-			</RecoilRoot>
+			<Provider store={store}>
+				{/* <SDKContext.Provider
```
avclarke, line 131: Yea I think like you identified, we don't really need the sdk context as can reference it directly due to being a singleton

state/store.ts
```diff
@@ -0,0 +1,34 @@
...
+import balancesReducer from './balances/reducer';
+import exchangeReducer from './exchange/reducer';
+import walletReducer from './wallet/reducer';
+
+// TODO:
+// - Make sure that using this static value (wagmiClient.provider)
+//   is actually responsive to provider changes.
+// - Figure out how to get the signer out of wagmi
+
+export let sdk = new KwentaSDK(10, wagmiClient.provider, window.ethereum as any);
```
avclarke, line 17: Was there a reason this changed to `let`?

sdk/common/balances.ts
avclarke: We could follow another pattern from Yearn for this, where they pass the root sdk class through to each service: https://github.com/yearn/yearn-sdk/blob/main/src/common.ts#L24

avclarke: Each class then extends the service interface: https://github.com/yearn/yearn-sdk/blob/main/src/interfaces/token.ts#L29
With `this` provided on init: https://github.com/yearn/yearn-sdk/blob/main/src/yearn.ts#L117
allowing each service to access any other part of the sdk: https://github.com/yearn/yearn-sdk/blob/main/src/interfaces/token.ts#L39

hooks/useExchange.ts
```diff
@@ -33,6 +34,11 @@ type ExchangeCardProps = {
 type ExchangeModal = 'settle' | 'confirm' | 'approve' | 'redeem' | 'base-select' | 'quote-select';
 export type SwapRatio = 25 | 50 | 75 | 100;
 
+sdk.events.on('signer_connected', () => {
```
avclarke: This is a nice solution but I feel it would be good if we could keep the sdk interaction isolated to connector and redux and have state changes drive redux actions. So in this case the connector can be the source of truth for state of wallet connection, when a change happens it would update the sdk with `await setSigner` and then update state with connection info and status. I think then, a page component that is managing a data query can listen to a `useEffect` change on connection status / wallet address for triggering fetches, or maybe an `onConnected` callback we add to the connector.

And if the connector is awaiting setSigner before any state is updated hopefully there should be no need for listening to connected event.

state/exchange/actions.ts
```diff
@@ -0,0 +1,382 @@
...
+
+export const submitApprove = createAsyncThunk<any, void, ThunkConfig>(
+	'exchange/submitApprove',
+	async (_, { getState, dispatch, extra: { sdk } }) => {
+		const {
+			exchange: { quoteCurrencyKey, baseCurrencyKey },
+		} = getState();
+
+		if (quoteCurrencyKey && baseCurrencyKey) {
+			const hash = await sdk.exchange.handleApprove(quoteCurrencyKey, baseCurrencyKey);
```
avclarke, line 147: I'm not sure about the UX requirements here but we might want to store the transactions in state like:
```
{
  hash: string,
  type: 'approve' ... etc,
  status: 'pending' | 'complete' | 'failed'
}
```
and then this gives us the opportunity to display and handle different transaction states in our UI as well as just the notifier, thinking here like we want to disable a button / show approving before being able to perform a swap. 

And probably good practice to do this for each transaction type.

sections/exchange/TradeCard/Cards/QuoteCurrencyCard.tsx
```diff
@@ -1,36 +1,60 @@
...
+	}));
 
-	const openQuoteModal = useCallback(() => setOpenModal('quote-select'), [setOpenModal]);
+	const quoteBalance = useAppSelector(selectQuoteBalanceWei);
+	const quotePriceRate = useAppSelector(selectQuotePriceRateWei);
+	const quoteCurrencyName = useAppSelector(selectQuoteCurrencyName);
+
+	const dispatch = useAppDispatch();
+
+	const quoteAmountDebounced = useDebouncedMemo(() => quoteAmount, [quoteAmount], 300);
```
avclarke, line 31: Another place where it seems like a lot of logic is duplicated between mobile and desktop, would be good to consolidate, maybe we can add a TODO

sections/exchange/TradeCard/CurrencyCard/CurrencyCardInput.tsx
```diff
@@ -105,6 +93,29 @@ const CurrencyCardInput: FC<CurrencyCardInputProps> = memo(
...
+			current = getPriceAtCurrentRate(current);
+		}
+
+		return current;
+	}, [priceRate, selectPriceCurrencyRate, getPriceAtCurrentRate, amount]);
+
+	return (
+		<CurrencyAmountValue data-testid="amount-value">
+			{currencyKeySelected && tradeAmount != null
+				? formatCurrency('sUSD', tradeAmount, { sign: '$' })
```
avclarke, line 113: Could use `formatDollars`

state/exchange/actions.ts
```diff
@@ -0,0 +1,417 @@
...
+	async (_, { extra: { sdk } }) => {
+		const {
+			balances: redeemableBalances,
+			totalUSDBalance: totalRedeemableBalance,
+		} = await sdk.exchange.getRedeemableDeprecatedSynths();
+
+		return {
+			redeemableSynthBalances: redeemableBalances.map((r) => ({
+				...r,
+				balance: '0',
```
avclarke, line 21: Just checking this is intentional to be 0?

state/exchange/actions.ts
```diff
@@ -0,0 +1,417 @@
...
+				const needsApproval = sdk.exchange.checkNeedsApproval(baseCurrencyKey, quoteCurrencyKey);
+
+				if (needsApproval) {
+					// TODO: Handle case where allowance is not MaxUint256.
+					// Simplest way to do this is to return the allowance from
+					// checkAllowance, store it in state to do the comparison there.
+					const isApproved = await sdk.exchange.checkAllowance(
+						quoteCurrencyKey,
+						baseCurrencyKey,
+						'0'
```
avclarke, line 217: Yea as your comment above we shouldn't have approved being set bigger than 0, it should be checked against amount when making a swap, let's do that in this PR otherwise it's buggy logic.

state/exchange/reducer.ts
```diff
@@ -0,0 +1,291 @@
...
+			state.tokenListLoading = true;
+		});
+		builder.addCase(fetchTokenList.fulfilled, (state, action) => {
+			state.tokenListLoading = false;
+			state.synthsMap = action.payload.synthsMap;
+			state.tokensMap = action.payload.tokensMap;
+			state.tokenList = action.payload.tokenList;
+		});
+		builder.addCase(fetchTokenList.rejected, (state) => {
+			state.tokenListLoading = false;
```
avclarke, line 242: It would be nice if we kept a consistent pattern for query states, like you done above on balances with `status: error | loading | success` I think we should stick to the pattern for all query states.

state/wallet/selectors.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const selectIsMainnet = createSelector(
+	(state: RootState) => state.wallet.networkId,
+	(networkId) => networkId && (networkId === 5 || networkId === 420)
+);
+
+export const selectBaseUrl = createSelector(
+	(state: RootState) => state.wallet.networkId,
+	(networkId) => {
+		if (networkId === 10 || networkId === 420) {
```
avclarke, line 41: This seems the same as getBaseUrl in blockExplorer, could we use that instead?

avclarke (changes requested):
Looks good, mostly just the balance and signer issue needs fixing and then added a few suggestions and clean up comments.

state/exchange/actions.ts
```diff
@@ -0,0 +1,382 @@
...
+	async (_, { getState, extra: { sdk } }) => {
+		const {
+			exchange: { quoteCurrencyKey, baseCurrencyKey },
+		} = getState();
+
+		if (quoteCurrencyKey && baseCurrencyKey) {
+			const needsApproval = sdk.exchange.checkNeedsApproval(baseCurrencyKey, quoteCurrencyKey);
+
+			if (needsApproval) {
+				// TODO: Handle case where allowance is not MaxUint256.
```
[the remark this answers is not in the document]
koredefashokun: Done

state/exchange/actions.ts
```diff
@@ -0,0 +1,382 @@
...
+import { monitorTransaction } from 'contexts/RelayerContext';
+import { toWei } from 'utils/formatters/number';
+
+type ThunkConfig = {
+	dispatch: AppDispatch;
+	state: RootState;
+	extra: { sdk: KwentaSDK };
+};
+
+export const fetchBalances = createAsyncThunk<any, void, ThunkConfig>(
```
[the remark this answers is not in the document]
koredefashokun: Outdated.

sections/exchange/MobileSwap/MobileSwapCurrencies.tsx
```diff
@@ -1,12 +1,36 @@
...
+
+	const routeToMarketPair = useCallback(
+		(baseCurrencyKey: string, quoteCurrencyKey: string) =>
+			router.replace('/exchange', ROUTES.Exchange.MarketPair(baseCurrencyKey, quoteCurrencyKey), {
+				shallow: true,
+			}),
+		[router]
+	);
+
+	const handleCurrencySwap = useCallback(() => {
```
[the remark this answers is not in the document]
koredefashokun: Replied to this earlier, but seems like it disappeared somehow. Basically, using a shared component was leading to a number of performance issues. However, I added a TODO like you recommended to look closer into re-merging the two.

sections/exchange/ExchangeModals/ExchangeModals.tsx
```diff
@@ -1,30 +1,83 @@
...
+
+	const onQuoteCurrencyChange = useCallback(
+		(currencyKey: string) => {
+			dispatch(setQuoteCurrencyKey({ currencyKey }));
+
+			if (baseCurrencyKey && baseCurrencyKey !== currencyKey) {
+				routeToMarketPair(baseCurrencyKey, currencyKey);
+			}
+
+			dispatch(fetchRates());
```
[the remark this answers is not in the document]
koredefashokun: Yeah.

sections/exchange/ExchangeModals/ExchangeModals.tsx
```diff
@@ -1,30 +1,83 @@
...
+		(currencyKey: string) => {
+			dispatch(setBaseCurrencyKey({ currencyKey }));
+
+			if (!!quoteCurrencyKey && quoteCurrencyKey !== currencyKey) {
+				routeToMarketPair(currencyKey, quoteCurrencyKey);
+			} else {
+				routeToBaseCurrency(currencyKey);
+			}
+
+			dispatch(fetchRates());
```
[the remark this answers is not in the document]
koredefashokun: Done.

sdk/transactions.ts
```diff
@@ -0,0 +1,115 @@
...
+const DEFAULT_GAS_BUFFER = 0.2;
+
+export default class TransactionsService {
+	private sdk: KwentaSDK;
+
+	constructor(sdk: KwentaSDK) {
+		this.sdk = sdk;
+	}
+
+	public createContractTxn(
```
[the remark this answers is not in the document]
koredefashokun: This is mostly to add feature-parity with the `useExchange`.

sdk/synths.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export default class SynthsService {
+	private sdk: KwentaSDK;
+	private balances: any;
+
+	constructor(sdk: KwentaSDK) {
+		this.sdk = sdk;
+	}
+
+	public async getSynthBalances() {
```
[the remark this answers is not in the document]
koredefashokun: Added.

sdk/data/synths.ts
```diff
@@ -0,0 +1,261 @@
...
+		},
+		category: 'forex',
+	},
+};
+
+// SNX
+// 1 - 0xC011a73ee8576Fb46F5E1c5751cA3B9Fe0af2a6F
+// 5 - 0x51f44ca59b867E005e48FA573Cb8df83FC7f7597
+// 10 - 0x8700dAec35aF8Ff88c16BdF0418774CB3D7599B4
+// 420 - 0x2E5ED97596a8368EB9E44B1f3F25B2E813845303
```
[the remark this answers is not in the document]
koredefashokun: Kind of. We had them earlier for the synths map, but it is not obvious whether we will need them later. Removed it though.

sdk/common/gas.ts
```diff
@@ -0,0 +1,72 @@
...
+const OVMGasPriceOracle = getContractFactory('OVM_GasPriceOracle').attach(
+	predeploys.OVM_GasPriceOracle
+);
+
+const contractAbi = JSON.parse(
+	// @ts-ignore
+	OVMGasPriceOracle.interface.format(ethers.utils.FormatTypes.json)
+);
+
+export const getL1SecurityFee = async (isL2: boolean, metaTx: MetaTx, signer?: ethers.Signer) => {
```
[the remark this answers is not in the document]
koredefashokun: True. Removed it.

pages/exchange.tsx
```diff
@@ -1,26 +1,61 @@
...
+	const router = useRouter();
+	const dispatch = useAppDispatch();
+
+	useEffect(() => {
+		const quoteCurrencyFromQuery = router.query.quote as string | undefined;
+		const baseCurrencyFromQuery = router.query.base as string | undefined;
+
+		if (!!quoteCurrencyFromQuery || !!baseCurrencyFromQuery) {
+			dispatch(resetCurrencies({ quoteCurrencyFromQuery, baseCurrencyFromQuery })).then(() => {
+				dispatch(checkNeedsApproval());
```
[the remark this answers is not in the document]
koredefashokun: Good idea. Added.

components/Currency/CurrencyIcon/TokenIcon.tsx
```diff
@@ -15,34 +15,42 @@ export type TokenIconProps = {
...
-			<Placeholder {...props}>{currencyKey === 'sDebtRatio' ? 'DEBT' : currencyKey}</Placeholder>
+			<Placeholder $isDeprecated={isDeprecated} {...props}>
+				{currencyKey === 'sDebtRatio' ? 'DEBT' : currencyKey}
+			</Placeholder>
 		);
 	}
 };
 
-const TokenImage = styled.img<{ isDeprecated?: boolean }>`
+const TokenImage = styled.img<{ $isDeprecated?: boolean }>`
```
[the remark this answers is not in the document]
koredefashokun: Lol. Just found out that I didn't submit my responses. GitHub is weird.

sdk/exchange.ts
```diff
@@ -0,0 +1,1227 @@
...
+		}
+
+		return zeroBN;
+	}
+
+	public async swapSynthSwap(
+		fromToken: Token,
+		toToken: Token,
+		fromAmount: string,
+		metaOnly?: 'meta_tx' | 'estimate_gas'
```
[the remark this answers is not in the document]
koredefashokun: Yeah. I think we can add this to the list of exchange API changes once this is merged.

sdk/exchange.ts
```diff
@@ -0,0 +1,1227 @@
...
+
+			const usdValue = amounts.amountReceived.sub(amounts.fee);
+			synthAmountEth = ethers.utils.formatEther(usdValue);
+		}
+
+		const params = await this.getOneInchSwapParams(oneInchFrom, oneInchTo, synthAmountEth);
+
+		const formattedData = getFormattedSwapData(params, SYNTH_SWAP_OPTIMISM_ADDRESS);
+
+		const synthSwapContract = new ethers.Contract(
```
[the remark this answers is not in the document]
koredefashokun: Added this to the list

sdk/exchange.ts
```diff
@@ -0,0 +1,1227 @@
...
+		const approveAddress = await this.getApproveAddress(quoteCurrencyKey, baseCurrencyKey);
+
+		if (!!quoteCurrencyContract) {
+			const allowance = (await quoteCurrencyContract.allowance(
+				this.sdk.walletAddress,
+				approveAddress
+			)) as ethers.BigNumber;
+
+			return wei(ethers.utils.formatEther(allowance)).gte(quoteAmount);
+		}
```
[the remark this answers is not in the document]
koredefashokun: Removed the conditional, so the frontend can handle this one.

sdk/exchange.ts
```diff
@@ -0,0 +1,1227 @@
...
+	constructor(sdk: KwentaSDK) {
+		this.sdk = sdk;
+		this.getAllTokensMap();
+	}
+
+	private get isL2() {
+		return [10, 420].includes(this.sdk.networkId);
+	}
+
+	public getTxProvider(baseCurrencyKey: string, quoteCurrencyKey: string) {
```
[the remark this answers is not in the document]
koredefashokun: We should add this to a list of the next phase of changes.

sdk/exchange.ts
```diff
@@ -0,0 +1,1227 @@
...
+	private tokenList: Token[] = [];
+	private allTokensMap: any;
+	private sdk: KwentaSDK;
+
+	constructor(sdk: KwentaSDK) {
+		this.sdk = sdk;
+		this.getAllTokensMap();
+	}
+
+	private get isL2() {
```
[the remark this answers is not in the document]
koredefashokun: Yeah. It will be on the `Context` class.

containers/Connector/Connector.tsx
```diff
@@ -68,6 +69,16 @@ const useConnector = () => {
 		dispatch(setSigner(signer));
 	}, [signer, dispatch]);
 
+	useEffect(() => {
```
avclarke, line 72: Probably best to move this to a hook and initialise it only on market pages, I think futures has something more custom doesn't it for rates?

platschi (dismissed):
Last UI test issues have been resolved and the exchange works now as expected, tested with several swaps between synths and tokens, fast and without issues except for some 1inch failed txs, but they are out of scope for this refactor. 

Giving the green light to merge this and continue testing `dev` in case we've missed any minor bugs. 

Awesome job, well done :+1:

containers/Connector/Connector.tsx
```diff
@@ -70,12 +72,14 @@ const useConnector = () => {
...
 
 	useEffect(() => {
-		const rateInterval = setInterval(() => {
-			dispatch(setExchangeRates());
-		}, 15000);
+		const ratesListener = (exchangeRates: Rates) => {
+			dispatch(setExchangeRates(exchangeRates));
+		};
+
+		sdk.events.on('exchangeRates_updated', ratesListener);
```
avclarke, line 79: Would be nice if we improved the api here to something like `sdk.exchange.onRatesUpdated()` rather than having to call on directly and pass the keys.

containers/Connector/Connector.tsx
```diff
@@ -70,12 +72,14 @@ const useConnector = () => {
...
 
 		return () => {
-			clearInterval(rateInterval);
+			sdk.events.removeListener('exchangeRates_updated', ratesListener);
```
avclarke, line 82: same here so we're not referencing events directly but something like `sdk.exchange.removeListener(ratesListener)`

sdk/services/exchange.ts
```diff
@@ -60,10 +61,16 @@ export default class ExchangeService {
 	private tokenList: Token[] = [];
 	private allTokensMap: any;
 	private sdk: KwentaSDK;
+	private exchangeRates: Rates = {};
 
 	constructor(sdk: KwentaSDK) {
 		this.sdk = sdk;
 		this.getOneInchTokens();
+
+		startInterval(async () => {
```
avclarke, line 70: Wondering if we should hold a reference so this could be cleared, as not a problem for us but if other clients were creating and destroying sdk instances this could lead to a memory leak. Maybe a todo is ok for now as not something we need to be concerned about with a singleton.

containers/Connector/Connector.tsx
```diff
@@ -70,12 +72,14 @@ const useConnector = () => {
 	}, [signer, dispatch]);
 
 	useEffect(() => {
-		const rateInterval = setInterval(() => {
-			dispatch(setExchangeRates());
-		}, 15000);
+		const ratesListener = (exchangeRates: Rates) => {
```
avclarke, line 75: Still not sure this is the right location for this, seems out of place in the connector. Maybe we can create a new component called AppData or something, as there will be quite a bit of app wide data we need to manage, similar to how we have RefetchProvider now, just doesn't need to be a context but something that lives at the very top level.

sdk/services/exchange.ts
```diff
@@ -430,7 +430,7 @@ export default class ExchangeService {
 			ethers.utils.formatBytes32String(currencyKey)
 		);
 
-		return numEntries ?? 0;
+		return Number(numEntries.toString()) ?? 0;
```
avclarke, line 433: As you replaced numEntries ?? 0 above makes it seems like this can be null or undefined so .toString() could error?

sdk/services/exchange.ts
koredefashokun: I think this is a limitation of JS. The React component model makes this easy to think about in the context of effects, but we'll probably have to have a method called `clearListeners()` on the SDK class for this purpose.

koredefashokun: You can check this out for more context: https://stackoverflow.com/questions/29333017/ecmascript-6-class-destructor. I'm skeptical about using obscure APIs like `WeakRef` though.

sdk/services/exchange.ts
koredefashokun: Good catch. Replaced with `numEntries ? Number(numEntries.toString()) : 0`.

containers/Connector/Connector.tsx
koredefashokun: Good idea.

koredefashokun: Added a simple `removeRatesListeners` method.

containers/Connector/Connector.tsx
koredefashokun: Created an AppData component, that simply returns its children, but initializes the necessary hooks. It feels like an antipattern though. I hope we can switch to simply running those functions outside components in the future, because putting them inside is just overhead for the component tree.

avclarke (dismissed):
Amazing work!

platschi (approved):

PR #1448, opened 2022-10-03
Remove daily stats query

Tburm:
Remove a redundant query

## Description
Remove a query that is only used on the mobile dashboard to calculate trades and volume. We have this information stored elsewhere in state, so we can avoid the additional network call.


sections/dashboard/MobileDashboard/FuturesMarkets.tsx
```diff
@@ -14,17 +14,30 @@ const FuturesMarkets = () => {
...
+			) ?? null
 		);
 	}, [futuresMarkets]);
 
+	const [trades, volume] = useMemo(() => {
+		const totalTrades = Object.values(futuresVolumes).reduce(
+			(total, { trades }) => total.add(trades),
+			wei(0)
+		);
+		const totalVolume = Object.values(futuresVolumes).reduce(
```
avclarke, line 34: Not a big deal but could be nice to combine these in a single reduce to save repeating the loop.

avclarke (approved):
Looks good!

sections/dashboard/MobileDashboard/FuturesMarkets.tsx
Tburm: good call, I combined them

platschi (approved):

PR #1449, opened 2022-10-04
adjusted PositionDetailsTooltips padding and preset location to fixed

burtrico:
I fixed the PositionDetails Tooltips' styling to remove the blank space above and below the tooltip text. 

I also changed the preset location for each tooltip to be fixed so that it expands right and does not ever get cut off by the left side of the window.

## Description
I fixed the PositionDetails Tooltips' styling by changing the top and bottom padding to 0px (left/right remain 10px). See the image at the bottom where padding is 0px 10px 0px 10px. This removes the blank space above and below the tooltip text.
 
I also changed the preset location prop for each PositionDetails Tooltip to be 'fixed' instead of 'bottom'. This included removing the a-typical LeftMarginTooltip styles to make all of the tooltips consistent using the 'fixed' preset. This removed the issue where tooltips would be cut off on the left side when the window was a medium width.

## Related issue
[Fix PositionDetails tooltip #1438](https://github.com/Kwenta/kwenta/issues/1438)

## Motivation and Context
This makes the PositionDetails tooltips consistent with the MarketDetails tooltips. Before the PositionDetails tooltips had blank space above and below the text.

Second, the fixed position correction was necessary to prevent the tooltips from being cut off by the left side of the window on a narrower screen (example shown below).
Screen Shot 2022-10-03 at 7 44 08 PM

## How Has This Been Tested?
I adjusted the styling and then tested in Brave browser as well as mobile. I also pulled it up on firefox for a quick glance.
I used metamask on the optimism network. I had to test with an open position in order to see the position details on mobile.
I kept my changes isolated to the PositionCard component.

Screen Shot 2022-10-03 at 6 00 53 PM
Screen Shot 2022-10-03 at 8 08 43 PM

### DIFF
--- a/sections/futures/PositionCard/PositionCard.tsx
+++ b/sections/futures/PositionCard/PositionCard.tsx
@@ -282,7 +282,7 @@ const PositionCard: React.FC<PositionCardProps> = () => {
 					</InfoRow>
 					<InfoRow>
 						<PositionCardTooltip
-							preset="bottom"
+							preset="fixed"
 							height={'auto'}
 							content={t('futures.market.position-card.tooltips.position-side')}
 						>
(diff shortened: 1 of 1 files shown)


platschi (approved):
Tested and looks good, works now both on mobile and desktop 👍🏼

LeifuChen (approved):
@burtrico great work. looks good to me.

PR #1463, opened 2022-10-06
Cross margin feedback

### LINKED ISSUE #1429: Limit Order Feedback
### General Feedback

- [x] Orders: Show all open orders in the Orders table, regardless of which market one is visiting right now (currently only shows ETH limit orders on ETH-PERP)
- [x] Pin any orders in current market to top of this list 

> Need later design a way to add e.g. a checkbox to hide/show only orders in current market instead of all

### Cross-Margin on Dashboard
- [x] Separate isolated and cross margin positions on the dashboard. #1434 
- [x] Futures history on the dashboard is does not distinguish isolated vs. cross margin trades. Add ring icons to the "market" to differentiate them.
- [x] Rename "Stop" to "Stop-Market"

### Later (not MVP?)
- #1432 
- #1386

### In PR #1463

- [x] Cross-Margin margin is not resetting from OP Goerli when switching to OP Mainnet on the dashboard

- [x] Display "reserved margin" for a market somewhere: (@asherism) Depends on #1469

- [x] Consider making Cross-Margin FAQ accessible somewhere after creating an account (@asherism , 1 liner with link like next order?)
- [x] Cross-Margin allows you to decrease leverage and submit the transaction even thou it doesn't do anything & shouldn't be able to (need to disable this and replace with a CTA) I.E. user selects max leverage and goes all in he can successfully but unsuccessfully lower his leverage, when it should prompt them to lower their position size or add more margin. 
- [x] Fix possible regression in adjust leverage modal

- [x] Add Beta tag to Cross-Margin so it is clear it's a beta (@asherism)

- [x] Add light theme icons for cross-margin/isolated margin (@asherism)

- [x] Bug: Seems no event triggered when withdrawing ETH, was showing 0.00 ETH Account Margin when it actually has 0.1 ETH deposited.

- [x] Check refetch/refresh at deposit/withdraw margin. After deposit of 50k sUSD UI did not refresh (needed hard refresh to show up)

- [x] Keeper ETH balance: Rename to "Account ETH Balance", add tooltip to explain why it's there. This would allow us to show this all the time instead of conditionally when they have stop/limit order types selected.

- [x ] Keeper ETH should have some warning about execution. For example, if the user withdraws their ETH but has an open order

- [x] Add tooltip to Fee Rejection label explaining what it is

### Mobile

- [x] Mobile Navigation Menu button "Isolated Margin" does not switch to Isolated, but market stays on CM when not connected to wallet.
- [x] Cross-Margin "Deposit" button empty when not connected to wallet (see screenshot below)

Image

- [x] Once connected, clicking "Switch to isolated market" button does not work

### Done
- [x] Deposit to cross margin account doesn't update sUSD balance in nav component
- [x] Set Cross-Margin as default throughout UI
- [x] Switch currency button is a bit confusing. It is a down arrow, I think a "switch" arrow would make more sense 🔃 
- [x] Input fields: Still showing "**s**ETH", make sure it just shows all token names without the **s**.  This applies to both isolated and cross-margin inputs 
- [x] Language: We should call it everywhere "Cross-Margin" or "Cross Margin" (without the dash), currently we mix both ways of naming it which looks weird. 
- [x] Size in ETH: Make sure we use token abbreviations everywhere (ETH, BTC, AVAX, etc.), not symbols. Also add some space between token abbr. and value for better readability.
- [x] Preview fills in 0$s when adding an arbitrary value in the amount input, while it only starts calculating correct values once I _also_ fill in the Limit Price. Maybe it should make sure to calculate a preview only once both inputs are filled in?
- [x] Keeper ETH Deposit doesn't have a preview during limit orders that will deposit .1 ETH
### Deposit/Withdraw modals
- [x] Balance input has 40px `margin-bottom`, unnecessary space?
- [x] Consider to remove "Gas Fee" information, irrelevant on L2

avclarke:
Closes #1429

- [x] Consider making Cross-Margin FAQ accessible somewhere after creating an account

- [x] Cross-Margin allows you to decrease leverage and submit the transaction even thou it doesn't do anything & shouldn't be able to (need to disable this and replace with a CTA) I.E. user selects max leverage and goes all in he can successfully but unsuccessfully lower his leverage, when it should prompt them to lower their position size or add more margin. 

- [x] Fix possible regression in adjust leverage modal

- [x] Add Beta tag to Cross-Margin so it is clear it's a beta 

- [x] Add light theme icons for cross-margin/isolated margin

- [x] Bug: Seems no event triggered when withdrawing ETH, was showing 0.00 ETH Account Margin when it actually has 0.1 ETH deposited.

- [x] Check refetch/refresh at deposit/withdraw margin. After deposit of 50k sUSD UI did not refresh (needed hard refresh to show up)

- Display "reserved margin" for a market somewhere.
       Depends on #1469 (will address in separate PR)

### Keeper

- [x] Keeper ETH should have some warning about execution. For example, if the user withdraws their ETH but has an open order
- [x] Keeper ETH balance: Rename to "Account ETH Balance", add tooltip to explain why it's there. This would allow us to show this all the time instead of conditionally when they have stop/limit order types selected.

### Limit Orders
- [x] Add tooltip to Fee Rejection label explaining what it is

### Mobile

- [x] Mobile Navigation Menu button "Isolated Margin" does not switch to Isolated, but market stays on CM when not connected to wallet.
- [x] Cross-Margin "Deposit" button empty when not connected to wallet (see screenshot below)

Image

- [x] Once connected, clicking "Switch to isolated market" button does not work


sections/futures/CrossMarginOnboard/CrossMarginFAQ.tsx
```diff
@@ -1,11 +1,13 @@
 import { useTranslation } from 'react-i18next';
 import styled from 'styled-components';
 
+export const CROSS_MARGIN_FAQ_URL = 'https://docs.kwenta.io/products/futures/cross-margin-accounts';
```
Tburm, line 4: We have `constants/links` but tend to put them in components when relevant. Do you think we should set a standard here?

Not to block this PR, but a good question: Where/how should we store/retrieve links?

Tburm (commented):
Left a few comments. One thing I can't get working is cancelling limit/stop orders. It looks like the order types do not specify the new `Stop-Market` in this PR, but also doesn't look like this is always where it fails. Sometimes the transaction is not created, and there is no feedback to the user when this happens.

Tburm:
Order cancelling is the only thing I think needs fixing before approval/merge.

sections/futures/CrossMarginOnboard/CrossMarginFAQ.tsx
avclarke: Good call, this should be in constants.

sections/futures/UserInfo/OpenOrdersTable.tsx
```diff
@@ -112,12 +112,14 @@ const OpenOrdersTable: React.FC = () => {
 	}, [cancelNextPriceOrder.hash, executeNextPriceOrder.hash]);
 
 	const rowsData = useMemo(() => {
-		if (!cancelling) return openOrders;
-		const copyOrders = [...openOrders];
-		const cancellingIndex = copyOrders.findIndex((o) => o.id === cancelling);
-		copyOrders[cancellingIndex] = { ...copyOrders[cancellingIndex], isCancelling: true };
-		return copyOrders;
-	}, [openOrders, cancelling]);
+		const ordersWithCancel = openOrders.map((o) => ({ ...o, cancel: () => onCancel(o) }));
```
avclarke, line 115: @Tburm the cancel issue turned out to be a really strange one where the cell render method in the table was never getting the latest callback and so crossMarginAccountContract was always null, unless you changed tabs and back again. I tried a few things but could never get it to work as is, so I moved it into the table data itself like this.

sections/dashboard/Overview/Overview.tsx
```diff
@@ -46,6 +47,7 @@ const Overview: FC = () => {
 				name: PositionsTab.CROSS_MARGIN,
 				label: t('dashboard.overview.positions-tabs.cross-margin'),
 				badge: crossPositions,
+				titleIcon: <FuturesIcon type="cross_margin" />,
```
Tburm, line 50: This is wrapping now on my screen, 14" macbook window maximized:
image

Tburm (dismissed):
One design comment, not sure if this is expected so I'll approve and re-approve if you want to fix it.

Tburm (approved):

Issue #1469, opened 2022-10-07
Add marginDelta to FuturesOrder in the subgraph


avclarke:
Will be needed to track reserved margin in the orders table.

https://github.com/Kwenta/margin-manager/blob/main/contracts/MarginBase.sol#L104

Tburm:
`marginDelta` has been added in this [PR](https://github.com/Kwenta/kwenta-subgraph/pull/94)

I'm working on a frontend update to add to types and update the subgraph file.

PR #1473, opened 2022-10-07
Stop Orders

Tburm:
Updates order queries to add the `marginDelta` value and rename "Stop" orders to "Stop-Market" in the queries

## Description
- Update `subgraph.ts`
- Update types to include the new `marginDelta` field
- Improve data cleaning by isolating a mapping function

## Related issue
- #1469


avclarke (approved):
Looks good!

PR #1474, opened 2022-10-10
Add icons to cross margin trades

Tburm:
Add icons to differentiate isolated from cross margin trades in trade history (dashboard) and leaderboard.

## Description
* Add isolated and cross margin icons on two tables

## Related issue
- #1429


Tburm:
@asherism - We may want to consider adding these icons in other places but I didn't want to make the change without your feedback. Two places I considered:
image

image

avclarke (approved):

PR #1475, opened 2022-10-10
Update free margin

Tburm:
Fixing a bug with the cross margin account overview on network switching. The query was not enabled when cross margin is not available, which means it was not cleared on networks without deployed contracts.

## Description
* Remove `enabled` parameters and make sure `freeMargin` is cleared on network switch
* Update the query to call "account" and "settings" data async
* Get the account data in `useFuturesData` from recoil instead of calling the query from 2 locations

## Related issue
- #1429


avclarke:
Would you be ok to base this off of #1463 as I already made some changes here so will save running into conflicts.

Tburm:
@avclarke One thing I'm not positive about - the free margin seems to take a bit longer than the other queries to update when switching from OP -> OP Goerli. Maybe not a big deal, but I couldn't isolate why it was delayed compared to other queries.

avclarke:
@Tburm I added a check to only fetch the settings once in my PR: https://github.com/Kwenta/kwenta/pull/1463/files#diff-2cbc12d406fcdd2f0a3e944dadd7ae759b5c252137b65ecf506fa607f50d8cc8R42

Could you rebase off of there and check if that helps with the delay at all?

Tburm:
@avclarke I did the rebase. It should make merging easier when your feedback PR is complete, but doesn't look like it fixed the delay.

avclarke:
ok thanks, I can take a look at the delay.

avclarke:
I was thinking more to open your PR against my branch so it's easier to review separately

avclarke:
But it's ok, we could just close mine now and both review this one.

Tburm:
@avclarke I actually think the new branch is cleaner, I'll close this one and make a new branch.

Issue #545, opened 2022-03-29
Implement Stats page


platschi:
### Subject of the issue
Implement a responsive version (mobile / desktop) of the Kwenta Stats page.
### Design
https://www.figma.com/file/mcaXc01qAA1si3I8shXkeI/Kwenta---Website?node-id=1061%3A1350

### Deliverables

- [x] Add a "Stats" navigation link to the trading page (to the right of landing page)
- [x] Implement a responsive Stats page to cover both mobile and desktop layout as defined in design spec above. 

Add the follow Statistics (same as mockup):

- [x] Futures Trading Volume over time (/w filters 24h/1w/1m/max)
- [x] Total Trades / Trader by Period
- [x] Total Unique Traders / Traders by Period
- [ ] TVL per asset
- [x] Current Total Open Interest per asset 

### Notes
For more detailed questions, please contact the devDAO PM. All data can also be found on the Dune dashboard for reference.

### Deadline
August 11, 2022

### Rewards
500 SNX

platschi:
#1485

Tburm:
@platschi we should be able to close this one after the PR is merged. There is no `TVL` chart, but we don't have a TVL to display. Even with cross margin, sUSD on active positions is burned so we can't check contracts for deposited sUSD.

PR #1485, opened 2022-10-14
Stats page

Tburm:
Creating a stats page linked from the homepage. This is an extension of the devDAO PR addressing [Issue #545](https://github.com/Kwenta/kwenta/issues/545)

## Related issue
- #545


pages/index.tsx
```diff
@@ -21,6 +24,19 @@ type HomePageComponent = FC & { layout?: FC<AppLayoutProps> };
...
 const HomePage: HomePageComponent = () => {
 	const { t } = useTranslation();
+	const router = useRouter();
+
+	const queryPage = router.query.page;
+	const [currentPage, setCurrentPage] = useState<TPages>('landing-page');
+
+	useEffect(() => {
+		if (queryPage === 'stats') {
+			setCurrentPage('stats-page');
```
avclarke, line 34: I think it'd be better to move this into its own page `pages/stats` rather than have conditional layout in the home page.

Tburm: Good call, I had just implemented the same logic from the devDAO branch. I didn't realized `HomeLayout` keeps the homepage header in place, which is what I was worried about losing. I just pushed a commit to create a stats page and simplify the routing.

Tburm: No problem @zhyd1997 ! Great work on the initial version, if you have any questions about the changes I made just let me know.

zhyd1997: Sure

platschi (dismissed):

constants/routes.ts
```diff
@@ -13,6 +13,7 @@ export const formatUrl = (route: string, params: Record<string, string>) => {
 export const ROUTES = {
 	Home: {
 		Root: '/',
+		Stats: '/stats',
```
avclarke, line 16: I think we can remove this and just route to `ROUTES.Stats.Home` instead? Otherwise it feels like a sub route of home. And unrelated to this PR but same for these dashboard routes under Home, seems they should be moved to under dashboard?

avclarke (commented):
Looks good, just a couple of minor comments / suggestions.

constants/routes.ts
Tburm: Good catch - I can move these around. I originally left stats under "Home" because it could only be access from the homepage, but now that it has it's own route we can lay that out here. The dashboard ones were a similar reason, I think left over from before we had the `/dashboard` route and the complete homepage.

platschi (dismissed):

avclarke (approved):
Looks good!

platschi (dismissed):

PR #1491, opened 2022-10-20
Fetch cross margin account from logs instead of subgraph

avclarke:
## Description
Remove subgraph dependency when querying cross margin account and replace with log filtering.

## How Has This Been Tested?
- Creating account
- Switching accounts


platschi:
First thing after connecting to wallet:

Screen Shot 2022-10-20 at 13 06 52

The account in question only approved (step 1) before this PR.

Tburm (approved):
This worked properly for me, and the code looks good.

One issue I had is that if I don't approve during onboarding the "deposit" modal will display "Deposit margin" but queue up an approval. We should fix on the other branch.

PR #1494, opened 2022-10-20
Improve trade preview speed

avclarke:
Batches some smart contract calls together before calculating the preview


Tburm (approved):
Very nice, it feels much faster and makes a lot of sense to fetch everything at once :+1:

PR #1498, opened 2022-10-21
Fix queries being interrupted by Metamask error on network switch

avclarke:
Because of a Metamask error that bubbles up during network switching, some contract queries get interrupted and error out causing state to fail to update. These workarounds ensure the queries will retry on failure to successfully complete the query on the next pass.


platschi (approved):
Works now with a fresh wallet, can switch within UI from L1 to L2 and back, no error message shown, loads onboarding within a second on L2.

Occasionally however it just disconnects the whole wallet, so I need to click "Connect" again after I switched from L1 to L2 within MM.

PR #1503, opened 2022-10-21
Design fixes

asherism:
flat market dropdown: 
image

remove susd currency from info box(free account margin):
image

input icon on hover state:
image

action button on hover state & style fix:
image

space between order types:
image

fixed border radius on position buttons:
image


Tburm:
I just want to acknowledge that replacing `sUSD` with `$` means we're assuming `sUSD` is always equal to $1. This is true according to the contracts, but could get confusing if `sUSD` is ever significantly depegged from $1.

platschi (approved):

PR #1506, opened 2022-10-24
Fix slider value label style and fix input delay

avclarke:
- Fixes the slider value label displaying as white on light theme
- Fix lag on native position size input

### DIFF
--- a/components/Slider/Slider.tsx
+++ b/components/Slider/Slider.tsx
@@ -1,4 +1,9 @@
-import Slider, { SliderProps as DefaultSliderProps } from '@material-ui/core/Slider';
+import Slider, {
+	SliderProps as DefaultSliderProps,
+	ValueLabelProps,
+} from '@material-ui/core/Slider';
+import Tooltip from '@material-ui/core/Tooltip';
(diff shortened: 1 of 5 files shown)


platschi:
It now shows the value on light theme when selecting the slider and moving it, as well as on hover, however otherwise it is still white on white.


https://user-images.githubusercontent.com/548702/197532966-4c1bf7be-74d0-4acf-af38-706cbcb72e5f.mov

avclarke:
> It now shows the value on light theme when selecting the slider and moving it, as well as on hover, however otherwise it is still white on white.
> 
>  Screen.Recording.2022-10-24.at.10.07.49.mov

@platschi  I had to make a bit of a trade off with the issue of the white label, turned out to be not a simple css change and so we're now passing our own component but if we don't have the display="auto" (display on use) it will always display which is an issue when it gets to min and max. As it seemed unnecessary to always display anyway I thought having it fade out was ok but let me know if you'd like to keep trying, maybe we can move to another task once we have this release out the way.

platschi (approved):

PR #1515, opened 2022-10-25
Fix leverage slider modal layout on mobile

avclarke:
The slider inside the edit leverage modal was causing the content to fit to the edge of the screen, this now ensures it maintains the padding.


platschi (approved):
Works perfectly now, both on desktop and mobile

PR #1516, opened 2022-10-25
Add check for existing account during cross margin onboarding

avclarke:
There was a user that managed to remain in the onboarding state once they created an account, this could potentially be to do with the logs query failing. These changes ensure if a user does get into that state that they will not be able to create another account, as an existing account will always be queried before submitting to create a new one. If the logs query fails it will error and if the query succeeds with an account the state will be updated and move the user to the next step.


LeifuChen (approved):
Looks good to me.

platschi (approved):

PR #1517, opened 2022-10-25
Position HIstory: Add limit and sort

Tburm:
Add a limit and sorting to the position history query to ensure we get the most recent positions. This corrects a bug where "Avg Entry" and "Realized Pnl" and missing data on the position card for open positions.


platschi (approved):
Works now as expected

PR #1518, opened 2022-10-25
Revert to querying cross margin account from subgraph

avclarke:
Querying logs is throwing an error on some wallets such as Rainbow and Coinbase, maybe an issue with the wallets provider but currently can see no way to fix it so reverting to querying the cm account from the subgraph for now.


Tburm (approved):

platschi:
With rainbow wallet on mobile, it now works (detects cross margin account), however the frontend does not detect my deposited cross margin balance on the markets page, neither on the dashboard.

Issue #1513, opened 2022-10-24
CM: Include decimals in adv. order input fields


platschi:
### Subject of the issue
For markets such as Euro or OP, the limit or stop loss order default input value just shows 0 or 1 given to their <1$ values. The input fields should default to a value with at least 2 decimals in order to indicate the value to be entered. This should be the default for the input field for all markets (2 decimals, e.g. 0.98)

![Screen Shot 2022-10-24 at 16 02 42](https://user-images.githubusercontent.com/548702/197605086-635263c9-7628-40d9-bd4d-8d323e888bbf.png)

###

- [ ] Show default input value for "Limit Price" input field with 2 decimals
- [ ] Show default input value for "Stop Market Price" input field with 2 decimals

### Bounty
100 $OP

PR #1519, opened 2022-10-26
chore: make 2 digits decimal

abhisheknaiidu:
## Description
added decimals for 2 places

## Related issue
#1513 

Screenshot 2022-10-26 at 7 23 08 PM

Screenshot 2022-10-26 at 10 43 17 AM

### DIFF
--- a/sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
+++ b/sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
@@ -63,7 +63,7 @@ export default function TradeCrossMargin({ isMobile }: Props) {
 			if (!invalidLabel || !price) {
 				onTradeOrderPriceChange(price);
 			}
-			setOrderPrice(marketAssetRate.toString(2));
+			setOrderPrice(price);
 		},
 		[onTradeOrderPriceChange, setOrderPrice, leverageSide, marketAssetRate, orderType]
 	);
@@ -108,10 +108,10 @@ export default function TradeCrossMargin({ isMobile }: Props) {
 							const price =
 								(type === 'limit' && leverageSide === 'long') ||
 								(type === 'stop market' && leverageSide === 'short')
-									? floorNumber(marketAssetRate, 0)
+									? floorNumber(marketAssetRate, 2)
 									: (type === 'stop market' && leverageSide === 'long') ||
 									  (type === 'limit' && leverageSide === 'short')
-									? ceilNumber(marketAssetRate, 0)
+									? ceilNumber(marketAssetRate, 2)
 									: '';
 							onChangeOrderPrice(String(price));
 						}}
--- a/sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
+++ b/sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
@@ -110,10 +110,10 @@ export default function TradeCrossMargin({ isMobile }: Props) {
 							const price =
 								(type === 'limit' && leverageSide === 'long') ||
 								(type === 'stop market' && leverageSide === 'short')
-									? floorNumber(marketAssetRate, 2) - offset
+									? floorNumber(marketAssetRate, decimals) - offset
 									: (type === 'stop market' && leverageSide === 'long') ||
 									  (type === 'limit' && leverageSide === 'short')
-									? ceilNumber(marketAssetRate, 2) - offset
+									? ceilNumber(marketAssetRate, decimals) + offset
 									: '';
 							onChangeOrderPrice(String(price));
 						}}
--- a/sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
+++ b/sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
@@ -21,7 +21,7 @@ import {
 	showCrossMarginOnboardState,
 	crossMarginAccountOverviewState,
 } from 'store/futures';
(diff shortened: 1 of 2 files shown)


sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
```diff
@@ -63,7 +63,7 @@ export default function TradeCrossMargin({ isMobile }: Props) {
 			if (!invalidLabel || !price) {
 				onTradeOrderPriceChange(price);
 			}
-			setOrderPrice(price);
+			setOrderPrice(marketAssetRate.toString(2));
```
platschi, line 66: This would show the current market rate, which we don't want here. The default field output depends on the order type selected and varies.

Make sure to use `setOrderPrice(price)` and have a look how you could add two decimals to the existing output

platschi (changes requested):

platschi (approved):
Looks like it is working fine now 👍🏼

sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
```diff
@@ -108,10 +108,10 @@ export default function TradeCrossMargin({ isMobile }: Props) {
 							const price =
 								(type === 'limit' && leverageSide === 'long') ||
 								(type === 'stop market' && leverageSide === 'short')
-									? floorNumber(marketAssetRate, 0)
+									? floorNumber(marketAssetRate, 2)
```
avclarke, line 111: These were rounded to integers to avoid the default order price being the same as current price, which is an invalid input for limit and stop orders. It wasn't an ideal solution but worked for the most part. Having it at 2 decimal places increases the chances the price will be the same when rounded as prices are often like `1556.43000000`. I think a better solution is to also add or subtract $0.01 to ensure they are offset from current price.

abhisheknaiidu: But it doesn't makes sense if we've any price < 1$, right?

avclarke: 100.0000 for example just becomes 100.01 or 99.99

abhisheknaiidu: @avclarke i get you, but does it makes sense if we've any price <= 1$, then it becomes 0.9 or 1.1, that is not correct right?

abhisheknaiidu: @platschi wdyt ?

avclarke: Sorry I read that as > 1 fist time round. So when price is smaller than 1 say 0.9 then our order price would default to 0.89 or 0.91. It does become a problem when price is 0.01 as we cannot have 0.00. For this case we could try changing the decimal amount. There is a function in number.js called `suggestedDecimals` which we can use to determine the decimal places then always just increment or decrement the last digit to shift the price.

abhisheknaiidu: @avclarke i think we're already doing it in `floorNumber` and `ceilNumber` functions

Screenshot 2022-10-27 at 12 02 21 AM

avclarke: ok perfect, we can not pass any decimals and just increment last digit of what's returned there?

abhisheknaiidu: It's doing the same what you've just mentioned above.

avclarke: A rough idea of what I'm thinking would be:
```
const decimals = suggestedDecimals(marketAssetRate)
const offset = 1 / 10 ** (decimals ?? 1)
let adjustedPrice = down ? floorNumber(marketAssetRate) : ceilNumber(marketAssetRate)
adjustedPrice = down ? adjustedPrice - offset : adjustedPrice + offset
```

abhisheknaiidu: @avclarke i've made the changes as you requested.

sections/futures/TradeCrossMargin/TradeCrossMargin.tsx
```diff
@@ -105,13 +105,15 @@ export default function TradeCrossMargin({ isMobile }: Props) {
 						onChange={(index: number) => {
 							const type = CROSS_MARGIN_ORDER_TYPES[index];
 							setOrderType(type as FuturesOrderType);
+							const decimals = suggestedDecimals(marketAssetRate);
+							const offset = 1 / 10 ** (decimals ?? 1);
 							const price =
 								(type === 'limit' && leverageSide === 'long') ||
 								(type === 'stop market' && leverageSide === 'short')
-									? floorNumber(marketAssetRate, 0)
+									? floorNumber(marketAssetRate, 2) - offset
```
avclarke, line 113: It still has `floorNumber(marketAssetRate, 2)` where it should either now be `floorNumber(marketAssetRate, decimals)` or just `floorNumber(marketAssetRate)` and let the function dictate suggested decimals. Also it should be minus the offset on floor and plus the offset on ceil.

abhisheknaiidu: @avclarke my bad, have done the changes 👍🏽

avclarke (approved):
Looks good!

Issue #1509, opened 2022-10-24
CM Onboarding: Deposit sUSD (step 3) improvements


platschi:
### Subject of the issue
During the onboarding process, step 3 users are asked to deposit their first balance into the cross margin account. Users have to manually input a value. Currently, the balance amount does not do anything.

It would be great to be able to click on the balance (e.g. in example below 200.10 sUSD) and the input field automatically populates with the max available value/balance.

As an example, see the "Amount" input field for Cross Margin Market orders, which has a 'Max' button that fills the total available margin. The clickable balance amount should look and work similar.

![Screen Shot 2022-10-24 at 12 35 58](https://user-images.githubusercontent.com/548702/197566842-0158c6d0-0f46-49ee-a259-8d07dc50a2c9.png)

Additionally, we should disable the "deposit susd" button when the input value is <50 sUSD. You can see for an example the "Deposit Margin" modal on markets, which already includes the error message if value is below 50 sUSD

Last, we should not capitalize the button text for the currency (sUSD instead of SUSD)

### Deliverables

- [x] Make the balance amount a clickable button that fills in the input field with the total available sUSD balance
- [x] Disable the "Deposit sUSD" button if value is < 50 sUSD and show "50 sUSD minimum margin to open a position" error message inside button
- [x] Fix the capitalized `s` in sUSD in button text, it should read: `Deposit sUSD`

### Bounty
250 $OP

PR #1520, opened 2022-10-26
CM onboarding modal step 3 improvements

0xwalde:
- Adds the ability to set the deposit amount to the maximum sUSD balance available by clicking on the displayed balance.
- Disables the "Deposit sUSD" button and shows a warning message when the amount being deposited is less than 50 sUSD
- Fixes the capitalization on the deposit button text from "SUSD" -> "sUSD"

Also includes a minor refactor to extract the `wei(50)` minimum margin amount to a shared constant since it was being defined in a few places. Always happy to revert that if it isn't desired or if there are any concerns about sharing this value across isolated and cross-margin code, however!

## Related issue
https://github.com/Kwenta/kwenta/issues/1509

## How Has This Been Tested?
Tested locally against Optimism Goerli

### DIFF
--- a/components/Input/InputBalanceLabel.tsx
+++ b/components/Input/InputBalanceLabel.tsx
@@ -8,9 +8,10 @@ import { formatCurrency } from 'utils/formatters/number';
 type Props = {
 	balance: Wei;
 	currencyKey: string;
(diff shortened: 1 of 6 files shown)


avclarke (approved):
Nice work, looks good!

Issue #1527, opened 2022-10-26
Implement Eligibility Box


platschi:
### Subject of the issue
Implement a simple box to the right sidebar of the dashboard for traders to enter their address and check if the given address is listed in the eligible addresses list for the SNX Aelin raise or not.

### Mockup
<img width="537" alt="Screen Shot 2022-10-26 at 12 53 20" src="https://user-images.githubusercontent.com/548702/198076299-f3ed3194-960f-424a-ab18-2dd4a0074833.png">

### Deliverables
- Add box to desktop dashboard sidebar with input, check error/success message depending on if address is included in eligibility list
- Make sure box is responsive and shown on mobile correctly, too

- If address input not in list, show red "Not eligible" message
- If address is eligible, change border and text to green and display "Address eligible" & a link icon pointing to the Aelin Raise pool

PR #1549, opened 2022-11-01
Add eligibility box

koredefashokun:
## Description
This PR adds a simple form for users to check if their address is eligible for the ongoing distribution.

## Related issue
#1527 

## Motivation and Context
N/A

## How Has This Been Tested?
N/A

image


sections/dashboard/EligibilityBox/EligibilityBox.tsx
```diff
@@ -0,0 +1,82 @@
...
+const EligibilityBox = () => {
+	const [address, setAddress] = useState('');
+	const [isEligible, setIsEligible] = useState<boolean>();
+
+	const handleChangeAddress = useCallback((e) => {
+		setAddress(e.target.value);
+	}, []);
+
+	const handleCheck = useCallback(() => {
+		setIsEligible(!!(!!address && EligibilityMap[address]));
```
avclarke, line 19: Could remove the !! before address as have it on the wrapper anyway

sections/dashboard/EligibilityBox/EligibilityBox.tsx
```diff
@@ -0,0 +1,82 @@
...
+	}, [address]);
+
+	return (
+		<div>
+			<EligibilitySectionTitle>SNX Staker Distribution</EligibilitySectionTitle>
+			<EligibilityContainer>
+				<EligibilityText>Enter your wallet address to check eligibility</EligibilityText>
```
avclarke, line 26: Not sure how long we're planning on having this feature in but should probably translate this if it's in here for good.

sections/dashboard/EligibilityBox/EligibilityBox.tsx
```diff
@@ -0,0 +1,82 @@
...
+				<EligibilityInput placeholder="0x..." value={address} onChange={handleChangeAddress} />{' '}
+				<Button fullWidth size="sm" onClick={handleCheck}>
+					Check
```
avclarke, line 29: Same here, should translate unless it's temporary

sections/dashboard/EligibilityBox/EligibilityBox.tsx
```diff
@@ -0,0 +1,82 @@
...
+				</Button>
+			</EligibilityContainer>
+			{isEligible !== undefined && (
+				<EligibilityStatus $isEligible={isEligible}>
+					{`Address is ${!isEligible ? 'not' : ''} eligible`}
```
avclarke, line 34: And here

sections/dashboard/EligibilityBox/EligibilityMap.ts
```diff
@@ -0,0 +1,2956 @@
+type StringMap = Record<string, boolean>;
+
+const EligibilityMap: StringMap = {
```
avclarke, line 3: Quite a big file to be storing in source, do you think it might be worth putting on IPFS or something? Again only something worth considering if permanent. Also probably a `Set` would make more sense here rather than object mapping true.

sections/dashboard/EligibilityBox/EligibilityBox.tsx
```diff
@@ -0,0 +1,82 @@
+import { useCallback, useState } from 'react';
+import styled from 'styled-components';
+
+import Button from 'components/Button';
+import Input from 'components/Input/Input';
+import Text from 'components/Text';
+
+import EligibilityMap from './EligibilityMap';
```
avclarke, line 8: As it's a constant I think we should uppercase as `ELIGIBILITY_MAP`

avclarke (commented):
Looks good, just a few comments. Not sure if this is permanent or not, so not worth worrying so much if it's not.

sections/dashboard/EligibilityBox/EligibilityBox.tsx
koredefashokun: Yeah. I don't think we're going to have it around for a long time, so I decided it's not really worth it.

sections/dashboard/EligibilityBox/EligibilityBox.tsx
koredefashokun: Same as above

sections/dashboard/EligibilityBox/EligibilityBox.tsx
koredefashokun: Same as above.

sections/dashboard/EligibilityBox/EligibilityMap.ts
koredefashokun: Same as above. It would be a good idea to have in IPFS otherwise, though.

koredefashokun: Regarding using a set vs a hash table, I'm not sure it really matters in this scenario, since under the hood (and performance-wise), they work identically.

sections/dashboard/EligibilityBox/EligibilityBox.tsx
koredefashokun: Changed.

sections/dashboard/EligibilityBox/EligibilityBox.tsx
koredefashokun: Changed.

sections/dashboard/EligibilityBox/EligibilityMap.ts
avclarke: It's more just for keeping it a bit cleaner and reducing the bloat of unnecessary `: true`, also seems a more a fitting data type for unique strings with the only function of checking if one exists.

avclarke (approved):

sections/dashboard/EligibilityBox/EligibilityBox.tsx
```diff
@@ -0,0 +1,90 @@
...
+			<EligibilityContainer>
+				<EligibilityText>Enter your wallet address to check eligibility</EligibilityText>
+				<EligibilityInput placeholder="0x..." value={address} onChange={handleChangeAddress} />
+				<Button fullWidth size="sm" onClick={handleCheck}>
+					Check
+				</Button>
+			</EligibilityContainer>
+			{isEligible !== undefined && !isLonoEligible && (
+				<EligibilityStatus $isEligible={isEligible}>
+					{`Address is ${!isEligible ? 'not' : ''} eligible for the Aelin pool (Nov 1 - Nov 14)`}
```
platschi, line 37: Can we add one more condition here to show "Visit pool" at the end of the sentence (add a dot to the end of the first sentence, too) and link this out to the pool on Aelin (once live/set up) _if_ address is eligible

platschi (changes requested):

sections/dashboard/EligibilityBox/EligibilityBox.tsx
LeifuChen: Good idea.

platschi (approved):

PR #1552, opened 2022-11-02
fix: updated pnl-graphic.png image

Himnashu:


platschi:
Thanks! Two things:

1) Kindly follow the docs at https://docs.kwenta.io/developers/devdao-contribute#first-time-set-up, mainly step 4 and sign your commits! This is good practice and we encourage every dev to do so by default.

2) Instead of adding a new file, you can just replace the old PnL graphic as it won't be needed anymore. This way we keep the repo clean of unnecessary (old) files.

Himnashu:
> Thanks! Two things:
> 
> 1. Kindly follow the docs at https://docs.kwenta.io/developers/devdao-contribute#first-time-set-up, mainly step 4 and sign your commits! This is good practice and we encourage every dev to do so by default.
> 2. Instead of adding a new file, you can just replace the old PnL graphic as it won't be needed anymore. This way we keep the repo clean of unnecessary (old) files.

Fixed both issues.

platschi (approved):

PR #1556, opened 2022-11-02
truncate values

### LINKED ISSUE #1487: Futures Markets: Truncate high values on mobile
### Subject of the issue
On the Futures Markets mobile page, dollar values for **24H Volume** as well as **Open Interest** above 1 million USD should be truncated. 

E.g. a value of $2,651,263.58 should be shown as 2.6M

### Information
See for implementation reference PR #1201 

### Screenshot

Image

### Deliverables
- Truncate values for 24H Volume and OI (Open Interest) from and above $1.000.00,00 on the **mobile** Futures Markets page

### Bounty
100 $OP

abhisheknaiidu:
fixes #1487


platschi (dismissed):
Tested and works as expected on mobile.

avclarke (commented):
It would be nice if instead of repeating the truncation logic and specification in each place we modified the number formatter to just accept a boolean `truncate: boolean` and do the truncation formatting logic there as it is something that should be consistent across all numbers. There is only a few places we use `truncation` throughout the app currently and it's all the same format so should be simple to change.

platschi:
Vercel failed to build, seems like a bug introduced in the last commit. Please check it out and make sure to run `npm run lint` before committing.

Screenshot 2022-11-07 at 18 03 18

avclarke (approved):
Looks good!

PR #1564, opened 2022-11-04
Cross Margin Competition

Tburm:
Enable the leaderboard for the upcoming cross margin competition

## Description
* Adds a dropdown to the "Leaderboard" section on the nav
* Adds a competition leaderboard that will pull either round 1 or 2 depending on inputs
* Dynamic tab buttons based on leaderboard type (competition vs all time)
* Opts for a constant instead of an environment variable and recoil value to set competition
  * This shouldn't make any difference, since if we update an environment variable we would still need to redeploy

### DIFF
--- a/constants/competition.ts
+++ b/constants/competition.ts
@@ -4,6 +4,8 @@
  */
 export const COMPETITION_DATES = {
 	// note: month starts at 0, not 1.
-	START_DATE: new Date(Date.UTC(2022, 7, 16)),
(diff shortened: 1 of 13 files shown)


platschi:
Do not forget to give the navigation some love ;)

Screenshot 2022-11-07 at 12 54 21

platschi (changes requested):

avclarke (approved):
Looks good!

platschi (approved):

PR #1567, opened 2022-11-05
Fix a bug where users can create extra cross margin accounts

### LINKED ISSUE #1566: Users able to create two cross margin accounts when only one is supported
### Subject of the issue
Some users have been able to make two cross margin accounts while the platform only supports a single account currently. 

### Expected behaviour
Users should never be able to make more than one cross margin account

avclarke:
closes #1566


platschi (approved):
Tested with a fresh testnet wallet, While in the process of approving the disabled button spinning wheel showed until the tx was confirmed and it moved to the approve step 2.  LGTM

Issue #1573, opened 2022-11-07
Trade History: Remove pagination


platschi:
### Subject of the issue
When scrolling down the (lazyloading) trade history table, after ~5 loads it shows pagination (1 out of 1 pages) at the bottom. This should not be there as the table should theoretically lazyload infinitely. This happens both on desktop and mobile.

This _might_ be due to the Table component treating each load as a new row, hitting the 5 row limit to "show" the pagination: https://github.com/Kwenta/kwenta/blob/0ccdc1dff1912bddd5fd22eb12cb083ef9568a0a/components/Table/Table.tsx#L113-L119

### Screenshot of the bug
<img width="330" alt="Screenshot 2022-11-07 at 18 14 36" src="https://user-images.githubusercontent.com/548702/200416893-d0acabdc-bf5d-44f6-be11-547ba1888dfe.png">

Mobile:
<img width="436" alt="Screenshot 2022-11-07 at 18 18 32" src="https://user-images.githubusercontent.com/548702/200417570-0cabde5b-fbf7-476f-9923-5d420775744e.png">

### Steps to reproduce
- Make sure pagination on trade history table for both desktop and mobile does not show after 5 "loads"

### Bounty
100 $OP

PR #1578, opened 2022-11-08
Fix pagination showing up on lazy-loaded trading history table

0xwalde:
Updates `Table.tsx` to only ever show pagination at the bottom of the table if the `showPagination` prop is true. Without this change, the pagination will show any time there are more than 100 rows (the `MAX_PAGE_ROWS`), even if `showPagination` is explicitly set to false.

I also added a type for the `react-table` object returned when using `useTable` in conjunction with `usePagination`, since there was a TODO in there that mentioned it.

## Related issue
https://github.com/Kwenta/kwenta/issues/1573

## How Has This Been Tested?
Tested locally against Optimism Goerli


LeifuChen (approved):
Looks good to me.

platschi (approved):

PR #1591, opened 2022-11-09
SDK: Context & Minor fixes (1.5)

koredefashokun:
## Description
This PR adds a context class to the SDK, fixes some minor structure issues, makes the documentation more comprehensive and prepares the implementation of the futures service.

## Related issue
N/A

## Motivation and Context
This PR continues the implementation of the Kwenta SDK.

## How Has This Been Tested?
N/A

### DIFF
--- a/containers/Connector/Connector.tsx
+++ b/containers/Connector/Connector.tsx
@@ -56,7 +56,7 @@ const useConnector = () => {
 	);
 
 	useEffect(() => {
-		sdk.context.setProvider(provider).then(handleNetworkChange);
+		sdk.setProvider(provider).then(handleNetworkChange);
 		transactionNotifier = new BaseTN(provider);
 	}, [provider, dispatch, handleNetworkChange]);
 
--- a/containers/Connector/Connector.tsx
+++ b/containers/Connector/Connector.tsx
@@ -56,7 +56,7 @@ const useConnector = () => {
 	);
 
 	useEffect(() => {
-		sdk.setProvider(provider).then(handleNetworkChange);
+		sdk.context.setProvider(provider).then(handleNetworkChange);
 		transactionNotifier = new BaseTN(provider);
 	}, [provider, dispatch, handleNetworkChange]);
 
--- a/sdk/context.ts
+++ b/sdk/context.ts
@@ -63,6 +63,10 @@ export default class Context implements IContext {
 		return [10, 420].includes(this.networkId);
 	}
 
+	get hasWalletAddress() {
+		return !!this.walletAddress;
+	}
+
 	public async setProvider(provider: ethers.providers.Provider) {
 		this.context.provider = provider;
 		this.multicallProvider.init(provider);
--- a/sdk/context.ts
+++ b/sdk/context.ts
@@ -6,13 +6,13 @@ import * as sdkErrors from './common/errors';
 import { ContractName, getContractsByNetwork, ContractMap } from './contracts';
 
 export interface IContext {
-	provider?: ethers.providers.Provider;
+	provider: ethers.providers.Provider;
+	networkId: NetworkId;
 	signer?: ethers.Signer;
-	networkId?: NetworkId;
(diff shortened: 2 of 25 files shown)


sdk/services/synths.ts
```diff
@@ -30,8 +30,8 @@ export default class SynthsService {
 			currencyKeys,
 			synthsBalances,
 			synthsUSDBalances,
-		]: SynthBalancesTuple = await this.sdk.contracts.SynthUtil.connect(
-			this.sdk.provider
+		]: SynthBalancesTuple = await this.sdk.context.contracts.SynthUtil.connect(
```
avclarke, line 33: Is this connected on context so wouldn't need to connect here?

containers/Connector/Connector.tsx
```diff
@@ -56,7 +56,7 @@ const useConnector = () => {
 	);
 
 	useEffect(() => {
-		sdk.setProvider(provider).then(handleNetworkChange);
+		sdk.context.setProvider(provider).then(handleNetworkChange);
```
avclarke, line 59: What do you think about keeping the api as `sdk.setProvider`, and the sdk handling the routing there, as context feels more an internal thing to the sdk?

sdk/context.ts
```diff
@@ -0,0 +1,97 @@
...
+		if (!this.context.provider) {
+			throw new Error(sdkErrors.NO_PROVIDER);
+		}
+
+		return this.context.provider;
+	}
+
+	get signer() {
+		if (!this.context.signer) {
+			throw new Error(sdkErrors.NO_SIGNER);
```
avclarke, line 52: Not sure about the pattern of throwing errors inside getters. Makes sense why it's been added so we don't have to keep throwing the same error inside functions but I think this can be a bit misleading with expecting a safe reference here.

sdk/context.ts
```diff
@@ -0,0 +1,97 @@
+import { NetworkId } from '@synthetixio/contracts-interface';
+import { Provider as EthCallProvider } from 'ethcall';
+import { ethers } from 'ethers';
+
+import * as sdkErrors from './common/errors';
+import { ContractName, getContractsByNetwork, ContractMap } from './contracts';
+
+export interface IContext {
+	provider?: ethers.providers.Provider;
```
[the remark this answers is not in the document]
koredefashokun: Technically we can have cases where clients may want to lazily set the provider using the `setProvider` method after creating an instance. However, since we're only considering Kwenta's use case right now, we don't really need to worry about that.

sdk/context.ts
```diff
@@ -0,0 +1,97 @@
...
+	signer?: ethers.Signer;
+	networkId?: NetworkId;
```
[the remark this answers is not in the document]
koredefashokun: Same as above.

sdk/services/synths.ts
koredefashokun: Good catch. I missed this one from when I made the changes across the SDK initially.

containers/Connector/Connector.tsx
koredefashokun: True. Re-added the sdk setProvider methods.

sdk/context.ts
koredefashokun: It's a little ugly, but it saves us from having to check for the existence of the values every time we want to use them. It also ensures type safety when we access the values.

sdk/services/exchange.ts
```diff
@@ -225,16 +221,16 @@ export default class ExchangeService {
 	}
 
 	public async getFeeReclaimPeriod(currencyKey: string) {
-		if (!this.sdk.walletAddress) {
+		if (!this.sdk.context.walletAddress) {
```
avclarke, line 224: Would no longer need this check if we keep the getter returning an error

avclarke (approved):

platschi (approved):

PR #1592, opened 2022-11-10
Hotfix: Closing SOL Positions

Tburm:
Hotfix position closing for the issue where SOL market size was reduced.

## Description
Change closing short SOL positions to do two `modifyPosition` transactions instead. One to open a 1 wei long position, then close.


LeifuChen (approved):
Looks good to me

avclarke (approved):

PR #1608, opened 2022-11-15
hotfix: mobile-spot-exchange-ETH/sETH-pair.

zhyd1997:
## Description

- [x] find the bug
```txt
checkAllowance:  Error: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] 
(method="allowance(address,address)", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, 
code=CALL_EXCEPTION, version=abi/5.7.0)
```
- [x] i18n submission disabled reason

<div align="center">
before -VS- after
</div>

<div align="center">
Screen Shot 2022-11-15 at 16 21 50
Screen Shot 2022-11-15 at 16 21 22
</div>

## Motivation and Context

Users can not swap ETH for sETH.

### DIFF
--- a/sdk/services/exchange.ts
+++ b/sdk/services/exchange.ts
@@ -663,12 +663,17 @@ export default class ExchangeService {
 		]);
 
 		if (!!quoteCurrencyContract) {
-			const allowance = (await quoteCurrencyContract.allowance(
-				this.sdk.context.walletAddress,
-				approveAddress
-			)) as ethers.BigNumber;
+			try {
+				const allowance = (await quoteCurrencyContract.allowance(
+					this.sdk.context.walletAddress,
+					approveAddress
+				)) as ethers.BigNumber;
 
-			return wei(ethers.utils.formatEther(allowance));
+				return wei(ethers.utils.formatEther(allowance));
+			} catch (e) {
+				// eslint-disable-next-line no-console
+				console.error('checkAllowance: ', e);
+			}
 		}
 	}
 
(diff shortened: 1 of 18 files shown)


sdk/services/exchange.ts
```diff
@@ -663,12 +663,17 @@ export default class ExchangeService {
...
+			try {
+				const allowance = (await quoteCurrencyContract.allowance(
+					this.sdk.context.walletAddress,
+					approveAddress
+				)) as ethers.BigNumber;
+
+				return wei(ethers.utils.formatEther(allowance));
+			} catch (e) {
+				// eslint-disable-next-line no-console
+				console.error('checkAllowance: ', e);
```
LeifuChen, line 675: Could you please use our own `logError` function here? It's easy to manage these error messages together in the future. https://github.com/Kwenta/kwenta/blob/main/utils/logError.ts#L1

state/exchange/reducer.ts
```diff
@@ -225,6 +225,11 @@ const exchangeSlice = createSlice({
 			state.allowance = action.payload.allowance;
 			state.oneInchQuoteError = false;
 		});
+		builder.addCase(resetCurrencyKeys.rejected, (state, action) => {
+			// handle any errors here.
+			// eslint-disable-next-line no-console
+			console.error(action.error.message);
```
LeifuChen, line 231: Same. Please use `logError` here.

LeifuChen (changes requested):
Thanks for the quick fix. Just request a minor change to use `logError`. It's the same thing right now but we have the plan to improve our log analysis later.

platschi (approved):

PR #1620, opened 2022-11-16
Share exchangeRates between exchange and futures

koredefashokun:
## Description
This PR makes the futures store functional. It defines a number of actions and selectors that make the application use the SDK exchangeRates as a single source of truth for rate-related data. As a result, the `useExchangeRatesQuery` hook has been deleted.

## Related issue
N/A

## Motivation and Context
This PR aims to improve querying performance.


sections/futures/UserInfo/UserInfo.tsx
```diff
@@ -218,7 +215,7 @@ const UserInfo: React.FC = () => {
 			)}
 			{showShareModal && (
 				<ShareModal
-					position={position}
+					position={position as any}
```
avclarke, line 218: We should try and avoid this sort of casting

avclarke (approved):

PR #1624, opened 2022-11-17
Refactor futures to use redux and sdk (part 1)

avclarke:
## Description
Refactor the futures page to use redux and the sdk to manage state


state/home/actions.ts
```diff
@@ -0,0 +1,17 @@
...
+import { ThunkConfig } from 'state/types';
+import { serializeMarkets } from 'utils/futures';
+
+export const fetchOptimismMarkets = createAsyncThunk<
+	{ markets: FuturesMarketSerialized[] },
+	providers.Provider,
+	ThunkConfig
+>('home/fetchOptimismMarkets', async (mainnetL2Provider, { extra: { sdk } }) => {
+	// For the home page we always fetch OP mainnet markets
+	const markets = await sdk.futures.getMarkets({ provider: mainnetL2Provider, networkId: 10 });
```
koredefashokun, line 14: Instead of passing a provider to this method, what do you think about creating an L2 provider instance on the SDK itself, since we might probably need it in other cases in the future.

state/futures/actions.ts
```diff
@@ -1 +1,38 @@
...
+
+import { Period } from 'sdk/constants/period';
+import { FuturesMarketSerialized } from 'sdk/types/futures';
+import { AppThunk } from 'state/store';
+import { ThunkConfig } from 'state/types';
+import { FuturesMarketAsset, serializeMarkets } from 'utils/futures';
+
+import { FundingRateSerialized } from './types';
+
+export const setMarketAsset = (asset: FuturesMarketAsset): AppThunk => (dispatch) => {
```
koredefashokun, line 11: Since this does not have any async requirements, I was wondering if there was a specific reason we don't just use the action returned from `createSlice` directly.

koredefashokun (commented):
Everything looks pretty good. Just a couple of small things I noticed and a few questions.

state/home/selector.ts
```diff
@@ -0,0 +1,5 @@
+import { RootState } from 'state/store';
+import { unserializeMarkets } from 'utils/futures';
+
+export const selectOptimismMarkets = (state: RootState) =>
```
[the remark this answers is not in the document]
avclarke: Good to know and yea will fix the name

state/home/actions.ts
avclarke: Yea might make sense, was thinking it could be a bit inefficient to keep two instances of the sdk around with one of them only being used on rare occasions. Seems this is the only place it's used right now so could revisit when there are more use cases?

state/futures/selectors.ts
```diff
@@ -1,47 +1,72 @@
...
+export const selectMarketKeys = (state: RootState) =>
+	state.futures.markets.map(({ asset }) => {
+		return MarketKeyByAsset[asset];
+	});
+
+export const selectMarketAssets = (state: RootState) =>
+	state.futures.markets.map(({ asset }) => asset);
+
+export const selectAverageFundingRates = (state: RootState) =>
+	unserializeFundingRates(state.futures.fundingRates);
```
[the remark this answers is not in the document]
avclarke: Was just reading here https://redux.js.org/usage/deriving-data-selectors that createSelector should always include transformation logic, so will update ones like these with unserialzing logic etc but the regular `(state: RootState) => state.futures.marketsQueryStatus;` types I'll leave as is. Maybe that's what you meant anyway.

state/futures/actions.ts
avclarke: Yea good spot, think it was in anticipation and then never added anything, will update.

state/futures/selectors.ts
```diff
@@ -1,47 +1,76 @@
...
+	(futures, marginType) => futures[accountType(marginType)].marketAsset
+);
+
+export const selectMarketRate = createSelector(
+	selectMarketKey,
+	selectExchangeRates,
+	(marketKey, exchangeRates) => newGetExchangeRatesForCurrencies(exchangeRates, marketKey, 'sUSD')
+);
+
+export const selectMarkets = (state: RootState) => unserializeMarkets(state.futures.markets);
```
koredefashokun, line 34: Looks like you might have skipped this selector by accident. Adding this suggestion, to help make the change a one-click one:

```suggestion
export const selectMarkets = createSelector(
  (state: RootState) => state.futures.markets,
  unserializeMarkets
);
```

state/futures/selectors.ts
```diff
@@ -1,47 +1,76 @@
...
+	(markets) => markets.map(({ asset }) => MarketKeyByAsset[asset])
+);
+
+export const selectMarketAssets = createSelector(
+	(state: RootState) => state.futures.markets,
+	(markets) => markets.map(({ asset }) => asset)
+);
+
+export const selectAverageFundingRates = (state: RootState) =>
+	unserializeFundingRates(state.futures.fundingRates);
```
koredefashokun, line 49: Leaving a suggestion for this selector as well:

```suggestion
export const selectAverageFundingRates = createSelector(
  (state: RootState) => state.futures.fundingRates,
  unserializeFundingRates
)
```

state/home/actions.ts
koredefashokun: Yeah. Fair point. It might be a premature optimization. Maybe we can make a note in the SDK readme for the future though.

koredefashokun (approved):

PR #1641, opened 2022-11-19
fix: disabled claim rewards button if the reward is not greater that 0.

zhyd1997:
## Description

Prevent additional gas fees cost.

## How Has This Been Tested?

Go to `/dashboard/earn/` page and see if the `Claim Rewards` button is disabled.

Screen Shot 2022-11-19 at 14 14 02

### DIFF
--- a/sections/earn/StakeGrid.tsx
+++ b/sections/earn/StakeGrid.tsx
@@ -45,7 +45,14 @@ const StakeGrid = () => {
 		<GridContainer>
 			<GridData title="Your Yield / Day" value={yieldPerDay} hasKwentaLogo />
 			<GridData title="Your Rewards" value={truncateNumbers(earnedRewards, 4)} hasKwentaLogo>
-				<Button fullWidth variant="flat" size="sm" style={{ marginTop: 10 }} onClick={handleClaim}>
+				<Button
+					fullWidth
+					variant="flat"
+					size="sm"
+					style={{ marginTop: 10 }}
+					disabled={!toWei(earnedRewards).gt(0)}
+					onClick={handleClaim}
+				>
 					Claim Rewards
 				</Button>
 			</GridData>


sections/earn/StakeGrid.tsx
```diff
@@ -45,7 +45,14 @@ const StakeGrid = () => {
 		<GridContainer>
 			<GridData title="Your Yield / Day" value={yieldPerDay} hasKwentaLogo />
 			<GridData title="Your Rewards" value={truncateNumbers(earnedRewards, 4)} hasKwentaLogo>
-				<Button fullWidth variant="flat" size="sm" style={{ marginTop: 10 }} onClick={handleClaim}>
+				<Button
+					fullWidth
+					variant="flat"
+					size="sm"
+					style={{ marginTop: 10 }}
+					disabled={!toWei(earnedRewards).gt(0)}
```
zhyd1997, line 53: bug fixes

LeifuChen: This is a good improvement in UX. Thanks for the initiative. I just want to emphasize that it's not a bug because the staking rewards contract supports the `zero reward claim` scenario.

zhyd1997: I can not find the contract so that I don't know what the scenario is, claim zero reward means users can get nothing but cost additional gas fee?

zhyd1997: Found it:

https://optimistic.etherscan.io/address/0x6077987e8e06c062094c33177Eb12c4A65f90B65#code

in `StakeRewards.sol` file, `getReward` function.

LeifuChen (approved):
Looks good. Thanks for the initiative.

platschi (approved):
Thank you, tested on mainnet and works great and as expected, good catch!

PR #1645, opened 2022-11-19
Update optimism ExchangeRates address

avclarke:
Fixes a number of perps not having rates and preventing from trading.


LeifuChen (approved):
It works now. Thanks for the quick fix.

platschi (dismissed):

platschi (approved):

Issue #1545, opened 2022-10-31
GnosisSafe wallet auto connect broken


platschi:
### Subject of the issue
From dHedge:
> GnosisSafe wallet auto connect connects to MetaMask instead of Gnosis Safe when opening Kwenta not on the main page, but on e.g. https://kwenta.io/market/?accountType=isolated_margin&asset=sETH

### Deliverables
- Make sure GnosisSafe wallet auto connect also works when visiting other pages than the main page

LeifuChen:
@platschi we can close this ticket.

PR #1648, opened 2022-11-20
Update the wagmi and rainbowkit version

LeifuChen:
## Description
Update both the `wagmi` and `rainbowkit `versions to the latest version.

Note:
1. The `watch-mode` bug in `wagmi` has been fixed in v0.8.5 (ref: https://github.com/wagmi-dev/wagmi/pull/1282). This makes our dapp more smoothly.
2. The old `safe-appd-wagmi` connector is not incompatible with the latest `wagmi` version.
We fixed this by copying the code directly into our code repo. Here is the reference: https://github.com/safe-global/safe-apps-sdk/issues/395
3. #1545 has been fixed after updating the latest rainbowkit version.

## Related issue
#1545 

## Motivation and Context
1. To make the development more productive by increasing the local dev performance
2. To prepare for adding more custom wallets by updating the latest version of rainbowkit.


hooks/useStakingData.ts
```diff
@@ -212,28 +218,27 @@ const useStakingData = () => {
...
+	const { data: vestingSchedules } = useContractRead({
 		...rewardEscrowContract,
 		functionName: 'getVestingSchedules',
 		args: [walletAddress ?? undefined, 0, 1000],
-		watch: false,
+		scopeKey: 'staking',
+		watch: true,
 		enabled: !!walletAddress,
-		select: (data) => data.filter((d) => d.escrowAmount.gt(0)),
+		select: (data) => (data as VestingEntry[]).filter((d) => d.escrowAmount > 0),
```
avclarke, line 228: Is this definitely a number? Seems like it would be a BigNumber

hooks/useStakingData.ts
```diff
@@ -50,54 +50,59 @@ type VestingEntry = {
 	entryID: number;
 };
 
+type VestingClaimable = {
+	quantity: number;
```
avclarke, line 54: Seems like values which should be strings or BigNumbers?

avclarke (commented):
Looks good, just a couple of questions on the types.

hooks/useStakingData.ts
LeifuChen: Yes. This should be `BigNumber`. The change has been made.

hooks/useStakingData.ts
LeifuChen: Yes. This should be `BigNumber`. The change has been made.

platschi (approved):

Issue #1630, opened 2022-11-18
Dashboard: Show liquid token holdings held in connected wallet on Spot Balance


platschi:
### Subject of the issue
Currently, the Kwenta Dashboard only shows the synth spot holdings of the connected wallet. It would be good to show also other major token holdings (stables, ETH, OP, etc.). Additionally, we would like to add the Kwenta token to the list.

### Current table example

![Screenshot 2022-11-18 at 10 13 19](https://user-images.githubusercontent.com/548702/202713306-757fb6dc-caa9-4ea5-823f-2e3f8f10ef32.png)

### Notes from devs
> We currently only show synth rates on the spot balance table. It requires some extra work by adding a new price feed (like the exchange page) to add Kwenta token into it.

### Expected behaviour
- Add new price feed to spot balance table component
- Fetch current USD price from Coingecko API
- Rename table header "Oracle Price" to just "Price"
- Show liquid token holdings to the table, if balance >0
- Add icons to repository if necessary (see https://github.com/kwenta/brand-kit) 
- Make sure USD value of liquid tokens is added up to the total "Portfolio value" on Dashboard 
- Make sure Total Spot balance value includes token holdings in the "Spot Balance" tab (43.78 USD in the example screenshot above)

### Bounty
500 $OP

PR #1652, opened 2022-11-21
feat: added Kwenta to spot balances table.

zhyd1997:
## Description

"**Spot Balance**" update:

- [x] show major token holdings (stables, ETH, OP, etc) including liquid $KWENTA (if balances > 0).
    > only on L2 like exchange page.
- [x] the price of the tokens above is fetched from CoinGecko.
- [x] Rename table header "Oracle Price" to just "Price".
- [x] add $KWENTA token icon.
- [x] add token holdings' USD balance to "Portfolio value" on Dashboard.
- [x] add token holdings' USD balance to "Spot Balance" tab on Dashboard.

## Related issue

Closes: #1630 

## How Has This Been Tested?

- switch to L2 mainnet and go to `/dashboard` to check "Portfolio value" and "Spot Balance" if includes token holdings.

- click "Spot Balance" tab to check if the table below includes major tokens (including $KWENTA if its balance > 0, make sure that its icon is shown too).

- works both on Mobile and PC, and sorting functionality is not broken on the spot balances table.

**PC**
https://user-images.githubusercontent.com/31362988/203453115-472b58b4-af1a-4f94-8063-e110f6a7344b.mov
**Mobile**
Screen Shot 2022-11-23 at 09 28 14


platschi:
Screenshot 2022-11-21 at 12 16 29

Awesome draft PR, this looks great already. A few small details I noticed:

1) `sUSD` is twice in the list; probably because it is an option on 1inch as well as among the classic synthetic assets we support. We probably want to make sure not to show them twice. Not sure if the same applies to the other synthetic assets (sBTC, sEUR, sLINK and so on).

2) Naming label is off, most assets just have the name "Synthetic" below their symbol. I believe this is because a few months ago, we were manipulating the string because the contracts were returning something like `Synthetic Synthetic US Dollar`. This affects both the Spot Balance table as well as the Exchange component, so we need to make sure to get these right.

3) As discussed in DM, we can probably show `Amount` values below 0.1 as <0.01. Alternatively, we could also think of a checkbox so users can decide if they want to hide Amounts <2 US$ or similar. Let's discuss this on Discord!

sections/dashboard/Overview/Overview.tsx
```diff
@@ -39,9 +48,117 @@ const Overview: FC = () => {
...
 	const [activeMarketsTab, setActiveMarketsTab] = useState<MarketsTab>(MarketsTab.FUTURES);
 
+	const { network, walletAddress, synthsMap } = Connector.useContainer();
+
+	const { tokenBalances } = useAppSelector(({ balances, exchange }) => ({
+		balancesMap: balances.balancesMap,
+		tokenList: exchange.tokenList,
+		tokenBalances: balances.tokenBalances,
+		balancesStatus: balances.status,
+	}));
```
koredefashokun, line 58: I think we can change this to:

```suggestion
	const tokenBalances = useAppSelector(({ balances }) => balances.tokenBalances);
```

sections/dashboard/Overview/Overview.tsx
```diff
@@ -39,9 +48,117 @@ const Overview: FC = () => {
...
+	// Only available on Optimism mainnet
+	const oneInchEnabled = network.id === 10;
+
+	const isL2 = useIsL2();
+
+	const [exchangeTokens, setExchangeTokens] = useState<any>([]);
+	const [kwentaBalance, setKwentaBalance] = useState(zeroBN);
+
+	useContractRead({
+		address: KWENTA_TOKEN_ADDRESS['10'],
```
avclarke, line 63: There is also a Goerli address here so we get by network `const tokenAddress = KWENTA_TOKEN_ADDRESS[network.id] ?? KWENTA_TOKEN_ADDRESS['10']`

And then also replace all the other places we reference KWENTA_TOKEN_ADDRESS['10']

queries/coingecko/useCoinGeckoTokenPricesQuery.ts
```diff
@@ -21,7 +21,7 @@ const useCoinGeckoTokenPricesQuery = (
 			const response = await axios.get<PriceResponse>(
 				`${CG_BASE_API_URL}/simple/token_price/${platform}?contract_addresses=${tokenAddresses
 					.join(',')
-					.replace(ETH_ADDRESS, ETH_COINGECKO_ADDRESS)}&vs_currencies=usd`
+					.replace(ETH_ADDRESS, ETH_COINGECKO_ADDRESS)}&vs_currencies=usd&include_24hr_change=true`
```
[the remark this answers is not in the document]
zhyd1997: Good catch 👍

sections/dashboard/Overview/Overview.tsx
zhyd1997: 1inch token list is only enabled on L2 mainnet and Coingecko is limited to query the Kwenta price at correct network (optimism-ethereum).

queries/coingecko/useCoinGeckoTokenPricesQuery.ts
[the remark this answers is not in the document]
zhyd1997: I found it before, but the fn in sdk needs 2 parameters?

sections/dashboard/Overview/Overview.tsx
avclarke: Ok no problem then

queries/coingecko/useCoinGeckoTokenPricesQuery.ts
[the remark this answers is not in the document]
zhyd1997: Done fb84d74.

sections/dashboard/Overview/Overview.tsx
```diff
@@ -39,9 +48,116 @@ const Overview: FC = () => {
...
+					description: name,
+					balance,
+					usdBalance,
+					price,
+					priceChange,
+				};
+			});
+
+			setExchangeTokens(_exchangeTokens);
+		})();
```
avclarke, line 151: Let's name the function instead of using an anon function and we should wrap this all in a try catch in case we get a failure, will cause the app to crash otherwise. Eventually we'll want to move this into redux but we can tackle that in a separate PR.

sections/dashboard/Overview/Overview.tsx
```diff
@@ -39,9 +48,116 @@ const Overview: FC = () => {
...
+		(token: any) => token.token.address.toLowerCase() === KWENTA_TOKEN_ADDRESS['10'].toLowerCase()
+	);
+
+	useEffect(() => {
+		(async () => {
+			const kwentaTokenObj = {
+				Kwenta: {
+					balance: kwentaBalance,
+					token: {
+						address: KWENTA_TOKEN_ADDRESS['10'].toLowerCase(),
```
avclarke, line 87: As we reference this a lot it would be nice to make it a const e.g. `const kwentaAddress = KWENTA_TOKEN_ADDRESS['10'].toLowerCase()`

avclarke (changes requested):
Looking good, just a couple of small things.

koredefashokun (approved):

platschi (approved):

PR #1656, opened 2022-11-21
Update Aggregations

Tburm:
Update queries to use new `FuturesAggregateStat` entities

## Description
- Update hourly volume to use the new entity
- Update volume and trades stats to use the "all markets" entity instead of adding all individual markets

Note: The `FuturesHourlyStat` entity can be deprecated on the frontend and subgraph when this is merged.

### DIFF
--- a/queries/futures/useGetStatsVolumes.ts
+++ b/queries/futures/useGetStatsVolumes.ts
@@ -8,6 +8,7 @@ import { minTimestampState } from 'store/stats';
 import { weiFromWei } from 'utils/formatters/number';
 import logError from 'utils/logError';
 
+import { AGGREGATE_ASSET_KEY } from './constants';
 import { getFuturesAggregateStats } from './subgraph';
 import { getFuturesEndpoint } from './utils';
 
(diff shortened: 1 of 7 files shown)


queries/futures/useGetStatsVolumes.ts
```diff
@@ -30,14 +24,16 @@ export const useGetStatsVolumes = () => {
...
+			const response = await getFuturesAggregateStats(
 				futuresEndpoint,
 				{
 					first: 999999,
 					orderBy: 'timestamp',
 					orderDirection: 'asc',
 					where: {
+						period: `${PERIOD_IN_SECONDS.ONE_DAY}`,
 						timestamp_gt: minTimestamp,
+						asset: '0x',
```
avclarke, line 36: Just wondering what this 0x asset is here?

avclarke (dismissed):
Looks good, just a question about the 0x asset, not sure if it's worth a comment to explain that?

queries/futures/useGetStatsVolumes.ts
Tburm: When I create aggregates I create one for each asset, then create this `0x` one which is the total of all assets

avclarke: Ok yea I see, might be worth creating a constant to help document, like `const AGGREGATE_ASSET = '0x'`? Other than that all good

avclarke (approved):

platschi (approved):

PR #1662, opened 2022-11-22
Fixed the missing period when selecting a new option

LeifuChen:
## Description
The `period` will be reset to `undefined` randomly when the user uses the epoch selector. So the card may show `Epoch All` under the `trading rewards` tab because `All` is the default value of `period`. This PR fixed this issue.

## How Has This Been Tested?
1. Click on the epoch selector to browse the options
2. The `trading rewards` card always shows the previous `period`.


platschi (approved):
Works like a charm now

Issue #1663, opened 2022-11-23
Can't drag to select input contents within Profit Calculator


0xwalde:
### Subject of the issue
When the Profit Calculator modal is open, text inside the inputs can't be selected by dragging the mouse across it as you'd expect since this is interpreted as a modal drag event instead (see video below). Instead, the entire modal moves regardless of whether you're actually dragging the body of the modal or the input content.

https://user-images.githubusercontent.com/109358247/203457926-7b03de62-8d65-4b29-a5ed-043e70476a98.mov

### Steps to reproduce
- Open the Profit Calculator
- Try to select the entry price by clicking and dragging
- The modal moves instead

0xwalde:
@platschi a small issue I noticed when trying to enter stuff into the Profit Calculator modal. Obviously not world-shattering, but a bit annoying and unexpected for desktop users.

PR #1664, opened 2022-11-23
Prevent profit calculator modal from moving when you drag inside an input

0xwalde:
The profit calculator modal current behaves a little weirdly since any click-and-drag event inside it results in the whole modal moving, even if the user is trying to select text inside an input on the modal. This is the default behavior of the [react-rnd](https://github.com/bokuweb/react-rnd) library used to provide the dragging implementation for the modal.

This PR fixes the weird behavior by specifying a [cancel-string](https://github.com/bokuweb/react-rnd#cancel-string) to disable the dragging movement on input elements within the modal.

## Related issue
https://github.com/Kwenta/kwenta/issues/1663

## How Has This Been Tested?
Tested on Optimism Goerli

## Video:
https://user-images.githubusercontent.com/109358247/203458855-2522d3b4-bdf0-451b-ae69-d8d103b826af.mov


platschi (approved):
Great find, thank you!

PR #1670, opened 2022-11-23
Fix position card avg entry preview calculation so it matches position chart

0xwalde:
The `modifiedAverage` functions in `PositionCard.tsx` and `PositionChart.tsx` are supposed to be calculating the same thing, but they get different results. It looks like the calculation in `PositionCard.tsx` is incorrect because it is adding the `size` property of the potential trade to get the new total size when it should be using `sizeDelta` instead.

In addition, the data being fed into the `PositionCard` version of the calculation can be positive or negative depending on the side of the trade, which isn't being handled quite right either.

## Related issue
https://github.com/Kwenta/kwenta/issues/1669

## How Has This Been Tested?
Tested on Goerli Optimism in the 4 following scenarios:
- Adding to an existing long position
- Subtracting from an existing long position
- Adding to an existing short position
- Subtracting from an existing short position

## Video:
https://user-images.githubusercontent.com/109358247/203590579-ff0d0a62-d234-498b-baf8-9cde501de13f.mov

https://user-images.githubusercontent.com/109358247/203591227-ad627d60-060f-413d-bab2-f0fcce907e28.mov


0xwalde:
Ah yeah, thanks for the feedback Troy! I was mainly concerned with making the card preview match the chart preview, but you're right that they're both wrong in several of those scenarios.

I pushed up another commit with some new calculation logic that should handle all of the cases you pointed out-- let me know if that seems to behave better. Thanks again!

Tburm (approved):

Issue #1658, opened 2022-11-22
Staking UI Polish


avclarke:
There's been some feedback that there is room for improvement in the staking UI / UX, let's review the designs and frontend, check they are in sync and frontend implemented fully and then see if there some areas for improvement.

platschi:
### Deliverables

In general, follow the Figma design as close as possible. For design considerations/examples, discuss with @asherism.

- [ ] Staking modal: Balance values should be a visible, clickable link (different shade for links, underscore), visible element when hovering over balance(?)
- [ ] Sort tabs by usage:  Staking, Escrow, Trading Rewards, Redemption
- [ ] Make top boxes clickable to bring you to the respective tab. So if you click "Escrow", "Staked Escrow" or "Vestable", it should switch to Staking UI "Escrow" tab, "Claimable" to the "Staking tab", and so on.
- [ ] Question: Why vesting table layout different from other table designs on the website? Streamline table design with other tables (e.g. see leaderboard as reference). Discuss with Yashar if necessary
- [ ] Fix background of Kwenta dApp dark theme to Figma design
- [ ] Bump contrast on text on dark theme (see Figma how it should be, partly solved with fixing background)

On Dashboard (left navigation):

- [ ] Remove gradient style
- [ ] Update button styles to latest design specifications
- [ ] Remove Governance from the list
- [ ] Update header colors etc. to current design

Issue #1584, opened 2022-11-09
Update header variants


platschi:
### Subject of the issue
> add all the new header variants from Figma (for mobile and desktop) with stories

### Deliverables
- add new header variants (for mobile and desktop) with stories to Kwenta
- tbd

PR #1673, opened 2022-11-24
Small fixes of the staking UI

LeifuChen:
## Description
Note: A new PR will be created to add the following feature
- Add `estimated rewards` back by showing the cross-margin-only fee and changed the ratio from `0.2` to `0.05` 

The following changes have been made.

### Trading Rewards Tab
- [x] Set sUSD values on the trading rewards tab to two decimals (leave $kwenta rewards at 4 decimals!)

### Staking Tab
- [x] Balance values should be a visible, clickable link (different shade for links, underscore), a visible element when hovering over balance(?)

### Staking Tabs
- [x] Sort tabs by usage:  Staking, Escrow, Trading Rewards, Redemption

### Staking Dashboard
- [x] Make top boxes clickable to bring you to the respective tab. So if you click "Escrow", "Staked Escrow" or "Vestable", it should switch to Staking UI "Escrow" tab, "Claimable" to the "Staking tab", and so on.

### General
- [x] Fix the background of Kwenta dApp dark theme to Figma design
- [x] Bump contrast on text on dark theme (see Figma how it should be, partly solved with fixing background)
- [x] Update Governance (move to Community)
- [x] Update header colors etc. to the current design (tracked in #1584)

## Related issue
#1658  #1584


sections/dashboard/Stake/TradingRewardsTab.tsx
```diff
@@ -101,25 +109,23 @@ const TradingRewardsTab: React.FC<TradingRewardProps> = ({
...
-	const SpotFeeQuery = useGetSpotFeeForAccount(walletAddress!, start, end);
-	const spotFeePaid = useMemo(() => {
-		const t = SpotFeeQuery.data?.synthExchanges ?? [];
+	const futuresFeeQuery = useGetFuturesFeeForAccount(walletAddress!, start, end);
+	const futuresFeePaid = useMemo(() => {
+		const t = futuresFeeQuery.data ?? [];
 
 		return t
-			.map((trade: any) => Number(trade.feesInUSD))
+			.map((trade: any) => Number(trade.feesPaid) / 1e18)
```
avclarke, line 117: could we use the correct type here instead of `any` and probably better to keep values in wei or BigNumber for arithmetic of these number sizes / decimals.

queries/staking/utils.ts
```diff
@@ -30,3 +31,13 @@ export function getStakingApy(totalStakedBalance: number, weekCounter: number) {
...
 }
+
+export function cobbDouglas(feesPaid: number, stakedBalance: number) {
+	return feesPaid ** 0.7 * stakedBalance ** 0.3;
+}
+
+export function getTradingRewards(weekCounter: number) {
+	const supplyRate = 1 - DECAY_RATE;
+	const initialWeeklySupply = INITIAL_WEEKLY_SUPPLY;
+	return initialWeeklySupply * supplyRate ** weekCounter * TRADING_REWARDS_RATIO;
```
avclarke, line 42: Should probably also do this math in BigNumber

LeifuChen: Sure. Just curious if there are any benefits to doing this math in BigNumber as most of them are constant numbers and `weekCounter `is a small integer.

sections/dashboard/Stake/TradingRewardsTab.tsx
LeifuChen: Yes. I will go through all the similar cases and use the correct type. Thanks for pointing it out.

queries/staking/utils.ts
avclarke: Initial weekly supply is a large decimal and straight up floating point arithmetic is prone to errors. It's only for display so not a huge issue but it's good practice to use big number / wei when working with decimal maths. Can read a bit more here and you might of noticed funky stuff when doing things with decimals in js number before. https://floating-point-gui.de/

queries/staking/utils.ts
```diff
@@ -23,10 +62,19 @@ export function getEpochDetails(networkId: number, epoch: number) {
...
+	const startWeeklySupply = wei(INITIAL_WEEKLY_SUPPLY).mul(SUPPLY_RATE.pow(weekCounter));
+	const yearlyRewards = startWeeklySupply.mul(wei(1).sub(SUPPLY_RATE.pow(52))).div(wei(DECAY_RATE));
+	return wei(totalStakedBalance).gt(0)
+		? yearlyRewards.mul(wei(STAKING_REWARDS_RATIO)).div(wei(totalStakedBalance))
+		: zeroBN;
+}
+
+export function cobbDouglas(feesPaid: number, stakedBalance: number) {
+	// pow(n), n must be a integer
+	return feesPaid ** 0.7 * stakedBalance ** 0.3;
```
avclarke, line 75: Same here would be good to update to Wei so we remain consistent with that

avclarke (commented):
Looks good, just one other place to switch to wei for math.

queries/staking/utils.ts
LeifuChen: Thanks for catching it. This function is supposed to be removed as we will read the score directly from the ipfs later.

Still curious how to re-write this to Wei. Is this the right way to do it `return wei(Math.pow(feesPaid, 0.7)).mul(wei(Math.pow(stakedBalance, 0.3)))`?

LeifuChen: The function has been removed in the latest commit.

platschi (approved):

PR #1674, opened 2022-11-24
Move Staking UI to Redux + SDK

koredefashokun:
## Description
This PR moves the staking UI away from wagmi hooks and to using Redux with new SDK methods.

## Related issue
N/A

## Motivation and Context
N/A

## How Has This Been Tested?
N/A


sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,153 @@
...
+			!KwentaToken ||
+			!SupplySchedule ||
+			!vKwentaToken ||
+			!MultipleMerkleDistributor ||
+			!veKwentaToken
+		) {
+			throw new Error(sdkErrors.UNSUPPORTED_NETWORK);
+		}
+
+		const RewardEscrowContract = new EthCallContract(RewardEscrow.address, []);
```
avclarke, line 124: Do you think it's worth also initialising EthCall contracts along with the typed contracts? Also did you get a chance to see if that fork was any good for combining them?

koredefashokun: I think it's probably worth a shot. Maybe we only do that for the contracts we use with multicall though?

sections/dashboard/Stake/InputCards/EscrowInputCard.tsx
```diff
@@ -1,112 +1,115 @@
...
+		setActiveTab(tabIndex);
+	}, []);
 
-	const submitEscrow = useCallback(async () => {
-		if (kwentaTokenApproval) {
-			const approveTxn = await kwentaApprove?.();
-			monitorTransaction({
-				txHash: approveTxn?.hash ?? '',
-			});
+	const isDisabled = useMemo(() => {
```
avclarke, line 74: Would be good to also disable these actions buttons while a transaction is pending and show a loading status. Can use the pattern of storing transaction state in staking state.

sections/dashboard/Stake/EscrowTable.tsx
```diff
@@ -237,15 +227,15 @@ const EscrowTable = () => {
 							{t('dashboard.stake.tabs.stake-table.kwenta-token')}
 						</div>
 					</div>
-					<VestButton disabled={!vest} onClick={() => setOpenConfirmModal(true)}>
+					<VestButton disabled={false} onClick={openConfirmModal}>
```
avclarke, line 230: can just omit instead of false

sections/dashboard/Stake/InputCards/EscrowInputCard.tsx
```diff
@@ -1,112 +1,115 @@
...
 
-	const amountBN = useMemo(
-		() => (amount === '' ? zeroBN.toString(0, true) : wei(amount).toString(0, true)),
-		[amount]
-	);
+	const escrowedKwentaBalance = useAppSelector(selectEscrowedKwentaBalance);
+	const stakedEscrowedKwentaBalance = useAppSelector(selectStakedEscrowedKwentaBalance);
+	const isKwentaTokenApproved = useAppSelector(selectIsKwentaTokenApproved);
+
+	const amountBN = useMemo(() => toWei(amount).toString(0, true), [amount]);
```
avclarke, line 32: named amountBN but it's returning a string. If it's for passing the calls as BigNumber wei value I think a better option could be  `wei(amount).toBN()` and we make those params BigNumber where it's more clear the expected type is a wei denomination.

sections/dashboard/Stake/InputCards/EscrowInputCard.tsx
```diff
@@ -1,112 +1,115 @@
...
+		unstakeEnabled,
+	]);
+
+	const balance = useMemo(() => {
+		const b = activeTab === 0 ? unstakedEscrowedKwentaBalance : stakedEscrowedKwentaBalance;
+		return truncateNumbers(b, DEFAULT_CRYPTO_DECIMALS);
+	}, [activeTab, unstakedEscrowedKwentaBalance, stakedEscrowedKwentaBalance]);
+
+	const handleChange = useCallback((_: any, newValue: string) => {
+		if (newValue !== '' && newValue.indexOf('.') === -1) {
```
avclarke, line 107: Just wondering why we need this extra logic here but it's not something we use on the other number value inputs?

sections/dashboard/Stake/InputCards/StakeInputCard.tsx
```diff
@@ -3,99 +3,103 @@ import _ from 'lodash';
...
+	const kwentaBalance = useAppSelector(selectKwentaBalance);
+	const stakedKwentaBalance = useAppSelector(selectStakedKwentaBalance);
 
 	const [amount, setAmount] = useState('');
 	const [activeTab, setActiveTab] = useState(0);
-	const amountBN = useMemo(
-		() => (amount === '' ? zeroBN.toString(0, true) : wei(amount).toString(0, true)),
-		[amount]
-	);
+	const amountBN = useMemo(() => toWei(amount).toString(0, true), [amount]);
```
avclarke, line 32: Again here maybe we want `wei(amount).toBN()`

state/staking/actions.ts
```diff
@@ -0,0 +1,198 @@
...
+export const fetchEscrowData = createAsyncThunk<any, void, ThunkConfig>(
+	'staking/fetchEscrowData',
+	async (_, { extra: { sdk } }) => {
+		const { totalVestable, escrowData } = await sdk.kwentaToken.getEscrowData();
+
+		return { totalVestable, escrowData };
+	}
+);
+
+export const vestEscrowedRewards = createAsyncThunk<any, number[], ThunkConfig>(
```
avclarke, line 83: Be good to replace the `any` return types with void if that's all it is, try and avoid `any` where possible.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,469 @@
...
+			this.sdk.context.walletAddress,
+			0,
+			1000
+		);
+
+		const vestingSchedules = schedules.filter((schedule) => schedule.escrowAmount.gt(0));
+
+		const calls = [];
+
+		for (const schedule of vestingSchedules) {
```
avclarke, line 235: I think we talked about using this for awaiting promises inside but if we're just pushing calls we can go with a simpler `.map`

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,469 @@
...
+			[]
+		);
+
+		return hash;
+	}
+
+	// TODO: Delete `approveLPToken` method.
+	// Also, consider creating a generic service called 'util' or something,
+	// that has handy methods for approving, redeeming etc.
+	// Signature: approveToken(contract: ContractName, spender?: ContractName)
```
avclarke, line 294: Nice idea, we'll prob use this same logic in various places.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,469 @@
...
+		const vestingEntries: {
+			quantity: ethers.BigNumber;
+			fee: ethers.BigNumber;
+		}[] = await this.sdk.context.multicallProvider.all(calls);
+
+		const escrowData = [];
+
+		let totalVestable = 0;
+
+		for (let i = 0; i < vestingSchedules.length; i++) {
```
avclarke, line 253: `reduce` would be a cleaner method here, maybe just copied code though.

avclarke (commented):
Nice work! Just a few suggestions

sections/dashboard/Stake/InputCards/EscrowInputCard.tsx
LeifuChen: It's the same component of the currency input card (exchange page). This is a quick fix to the following issue.
image

sdk/services/kwentaToken.ts
koredefashokun: I decided to use a for loop for the performance gains in this scenario.

sdk/services/kwentaToken.ts
koredefashokun: Yeah, I know. I personally prefer using for loops outside components because they're quite a bit faster, and we're not really working in a FP context here. It might be something interesting to discuss though, since we should probably be setting standard for these things anyway. I just figured we could take a perf win wherever we can find one.

sections/dashboard/Stake/EscrowTable.tsx
koredefashokun: Yeah. This was a placeholder anyway. Updated. That said, I think omitting sets it to `true` though. I believe it's done that way since most booleans default to false.

avclarke: Ah I just meant to remove the disabled prop rather than set to false but yea all good now anyway.

sdk/services/kwentaToken.ts
avclarke: I think if we were iterating over thousands of items or was a specific area we were looking to optimise it might be something to consider but where there would really be no noticeable gain I think we should always go with the cleaner code option and remain consistent with our other iteration patterns throughout the codebase.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,469 @@
...
+			throw new Error(sdkErrors.UNSUPPORTED_NETWORK);
+		}
+
+		const { hash } = await this.sdk.transactions.createContractTxn(
+			this.sdk.context.contracts.StakingRewards,
+			action,
+			[wei(amount).toBN()]
+		);
+
+		return hash;
```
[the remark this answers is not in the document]
koredefashokun: Done in the latest commit

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,469 @@
...
+
+		const { contract, spender } = map[token];
+
+		if (!contract || !spender) {
+			throw new Error(sdkErrors.UNSUPPORTED_NETWORK);
+		}
+
+		const { hash } = await this.sdk.transactions.createContractTxn(contract, 'approve', [
+			spender.address,
+			ethers.constants.MaxUint256,
```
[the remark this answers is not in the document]
koredefashokun: Added `amount` param to `approveToken` method with `ethers.constants.MaxUint256` as the default.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,469 @@
...
+			throw new Error(sdkErrors.UNSUPPORTED_NETWORK);
+		}
+
+		const { hash } = await this.sdk.transactions.createContractTxn(RewardEscrow, 'vest', [ids]);
+
+		return hash;
+	}
+
+	// TODO: Rename this and refactor.
+	public async performStakeAction(
```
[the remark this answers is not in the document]
koredefashokun: True. Added as well.

sections/dashboard/Stake/InputCards/StakeInputCard.tsx
koredefashokun: Done.

sections/dashboard/Stake/InputCards/EscrowInputCard.tsx
koredefashokun: Done

state/staking/actions.ts
koredefashokun: Fixed.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,464 @@
...
+		const MultipleMerkleDistributorContract = new EthCallContract(
+			MultipleMerkleDistributor.address,
+			MultipleMerkleDistributorABI
+		);
+
+		const fileNames = periods
+			.slice(0, -1)
+			.map(
+				(i) =>
+					`trading-rewards-snapshots/${
```
avclarke, line 394: noticed this is done in two places, I think it would be nice to extract into a helper, would also clean things up a bit.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,464 @@
...
+
+		const claimed: boolean[] = await this.sdk.context.multicallProvider.all(
+			rewards.map((reward) => MultipleMerkleDistributorContract.isClaimed(reward[0], reward[4]))
+		);
+
+		const { totalRewards, claimableRewards } = rewards.reduce(
+			(acc, next, i) => {
+				if (!claimed[i]) {
+					acc.claimableRewards.push(next);
+					acc.totalRewards += Number(next[2]) / 1e18;
```
avclarke, line 422: Should probably use Wei here so we remain consistent with how we treat big numbers and just to keep on the safe side of floating point issues.

state/staking/types.ts
```diff
@@ -0,0 +1,31 @@
...
+	claimableBalance: string;
+	totalStakedBalance: string;
+	stakedEscrowedKwentaBalance: string;
+	stakedKwentaBalance: string;
+	epochPeriod: number;
+	weekCounter: number;
+	kwentaAllowance: string;
+	vKwentaAllowance: string;
+	veKwentaAllowance: string;
+	totalVestable: number;
```
avclarke, line 19: I think we should avoid using number anywhere for amounts as we use Wei everywhere else throughout the app, so best to remain consistent.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,464 @@
...
+			proof: string[];
+		};
+	};
+};
+
+export type EscrowData = {
+	id: number;
+	date: string;
+	time: string;
+	vestable: number;
```
avclarke, line 44: Would be good to avoid using number and always stick to Wei for values so we remain consistent throughout.

avclarke (commented):
Just a couple of clean up bits I noticed, I think we should avoid using number anywhere for amounts so we remain consistent throughout with the use of Wei.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,464 @@
...
+		const { RewardEscrow } = this.sdk.context.contracts;
+
+		if (!RewardEscrow) {
+			throw new Error(sdkErrors.UNSUPPORTED_NETWORK);
+		}
+
+		return this.sdk.transactions.createContractTxn(RewardEscrow, 'vest', [ids]);
+	}
+
+	public stakeKwenta(amount: string | ethers.BigNumber) {
```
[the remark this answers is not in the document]
koredefashokun: Yeah it's not necessary. It still returns a promise. `async` is only necessary when there's an await in the function body.

sdk/services/kwentaToken.ts
koredefashokun: Actually just realized we only use one of those two instances, so I deleted the other one (they were doing the exact same thing).

sdk/services/kwentaToken.ts
koredefashokun: Fair enough.

state/staking/types.ts
koredefashokun: Will fix

sdk/services/kwentaToken.ts
koredefashokun: Yeah. Will replace with wei.

sdk/services/kwentaToken.ts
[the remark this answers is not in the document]
avclarke: ok yea just more a readability thing then, would be good to either add it here or remove from the others for consistency.

sdk/services/kwentaToken.ts
```diff
@@ -0,0 +1,428 @@
...
+		);
+
+		const vestingSchedules = schedules.filter((schedule: any) => schedule.escrowAmount.gt(0));
+
+		const calls = vestingSchedules.map((schedule: any) =>
+			RewardEscrow.getVestingEntryClaimable(this.sdk.context.walletAddress, schedule.entryID)
+		);
+
+		const vestingEntries: {
+			quantity: ethers.BigNumber;
```
avclarke, line 207: Would be nice to import BigNumber directly so we don't have to do ethers.BigNumber everywhere

koredefashokun: Sure. Change pushed.

sdk/services/kwentaToken.ts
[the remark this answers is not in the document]
koredefashokun: The thing is, those other instances require the `async` keyword since we use `await` in their body. However, here, we don't use `await` in the body, so we can omit both the `async` and `await` and still return a promise.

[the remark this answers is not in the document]
avclarke: Ah sorry I just mean specifically these functions all calling  `this.performStakeAction`, they all have async apart from this one, so just a clean up thing for consistency:
Screenshot 2022-12-05 at 07 48 25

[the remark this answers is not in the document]
koredefashokun: Ah. Yeah. Good catch. For some reason, I didn't notice that those were still using `async`.

avclarke (approved):
Nice work!

avclarke (dismissed):

platschi (changes requested):
Feedback see Discord

PR #1690, opened 2022-11-29
Add `marketKey` to queries

Tburm:
Adds new `marketKey` field to several queries, includes ones where we summarize volume and count trades. This is necessary to transition to Perps V2 since some new markets will contain the same `asset` value (`sETH`) but have a different market key (`pETH` instead of `sETH`)


platschi:
On landing page:

Screenshot 2022-11-28 at 21 43 46

Also it  seems `24 Vol` data is not fetched on the vercel deploy, but maybe that's expected?

Tburm:
@platschi This is expected, I meant to leave as a draft until tomorrow morning when the subgraph syncs. It looks good on testnet, I can just verify on mainnet when that's complete.

platschi (approved):

Issue #1682, opened 2022-11-25
Markets: Implement suggestedDecimals


platschi:
### Subject of the issue
From Discord:
> why is it when you open a matic position, it lists it at $x.xx instead of like the chart as $x.xxxx, sucks to not know where your entry is.

Currently, only `EUR, DOGE` and `DEBT`-PERPs prices in USD show 4 decimals instead of two. Other markets, such as `OP`, show only two decimals, which is counterproductive when trading low values. This should be streamlined throughout the dApp depending on the current USD price.

### Solution
We want to use `suggestedDecimals` in `utils/number` for all price values on the Markets UI.

### Deliverables
Make sure all values shown for all PERP markets follow `suggestedDecimals` on the UI:
- MarketDetails
- PositionCard
- TradeHistory panel
- Trades tab
- Markets dropdown
- Dashboard: Current open position
- Dashboard: Markets list
- History: Futures History, Price column
- Other places where this might be visible

![Screenshot 2022-11-28 at 21 23 51](https://user-images.githubusercontent.com/548702/204408646-d5790230-da73-49de-ba22-3fffb48e424f.png)

![Screenshot 2022-11-28 at 21 28 02](https://user-images.githubusercontent.com/548702/204408764-082bbd3d-26ab-4939-b85b-eae101fc389d.png)

![Screenshot 2022-11-28 at 21 30 08](https://user-images.githubusercontent.com/548702/204409073-f05faba3-a4f9-415c-9a2f-9dc278c0afb0.png)

### Bounty
250 $OP

PR #1692, opened 2022-11-29
fix: used `suggestedDecimals` in all possible places.

zhyd1997:
## Related issue

Resolves: #1682

### DIFF
--- a/utils/formatters/number.ts
+++ b/utils/formatters/number.ts
@@ -208,9 +208,10 @@ export const weiFromWei = (weiAmount: WeiSource) => {
 export const suggestedDecimals = (value: WeiSource) => {
 	value = wei(value).toNumber();
 	if (value >= 10000) return 0;
-	if (value >= 10 || value === 0) return 2;
-	if (value >= 0.001) return 4;
-	return 5;
+	if (value >= 4 || value === 0) return 2;
+	if (value >= 0.1) return 4;
+	if (value >= 0.01) return 5;
+	return 6;
 };
 
 export const floorNumber = (num: WeiSource, decimals?: number) => {
--- a/utils/formatters/number.ts
+++ b/utils/formatters/number.ts
@@ -4,11 +4,7 @@ import { BigNumber, ethers, utils } from 'ethers';
 import { parseUnits } from 'ethers/lib/utils';
 
 import { CurrencyKey } from 'constants/currency';
-import {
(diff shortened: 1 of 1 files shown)


platschi:
Thanks! I suggest we adjust `suggestedDecimals` to something like this:
```ts
export const suggestedDecimals = (value: WeiSource) => {
	value = wei(value).toNumber();
	if (value >= 10000) return 0;
	if (value >= 10) return 2;
	if (value >= 0.001) return 4;
	return 5;
};
```

@avclarke what do you think? Any implications with this changes for the Exchange?

With the current solution, however, `$0.00` placeholders are now set to 5 decimals by default all over the place. 

Screenshot 2022-11-29 at 11 43 56

We should catch this, maybe by adding something like `	if (value === 0) return 2;` this way dust amounts should still show with up to 5 decimals on the Exchange.

Screenshot 2022-11-29 at 11 48 36

zhyd1997:
@platschi 👍🏿

platschi:
Update suggestion from @Marko-roy :

```ts
export const suggestedDecimals = (value: WeiSource) => {
    value = wei(value).toNumber();
    if (value >= 10000) return 0;
    if (value >= 4) return 2;
    if (value >= 0.1) return 4;
    if (value >= 0.01) return 5;
    return 6;
};
```

platschi (approved):
Looks good on UI

utils/formatters/number.ts
```diff
@@ -215,10 +209,10 @@ export const weiFromWei = (weiAmount: WeiSource) => {
 export const suggestedDecimals = (value: WeiSource) => {
 	value = wei(value).toNumber();
 	if (value >= 10000) return 0;
-	if (value >= 1) return 2;
-	if (value >= 0.01) return 3;
-	if (value >= 0.001) return 4;
-	return 5;
+	if (value >= 4 || value === 0) return 2;
```
avclarke, line 212: Bigger than 4 seems like a strange rule, is this mean to be 10 or something?

zhyd1997: Although seems strange, that's correct.

platschi: Yeah it's correct, I double checked with Marko.

avclarke: I'll ask Marko what his reasoning his

platschi (approved):

Issue #1467, opened 2022-10-07
Rainbowkit: Add injected wallet connector & set default provider through wallet


platschi:
1) Brave Browser does not show up nor works anymore correctly with Kwenta.
2) We should try to push the provider back on the wallet if possible.

## Deliverables
- Add injected wallet connector to support different injected wallets.
- Bump rainbowkit, wagmi and dependencies to latest versions*
- Make sure Brave Browser support is working
- Investigate options and try to use wallet provider as default instead of Kwenta BlastAPI (BYOP better for decentralization)

e.g. rainbowkit 0.7.0 introduced tree shaking of individual wallets, see [rainbowkit/releases](https://github.com/rainbow-me/rainbowkit/releases)

LeifuChen:
@platschi shall we close this ticket?

PR #1696, opened 2022-11-29
Update the rainbowkit and fix the breaking changes

LeifuChen:
## Description
1. Brave Browser support is working in rainbowkit v8.0
2. Add injected wallet connector as a fallback wallet options

## Related issue
#1467


platschi (approved):
Works great on Brave. This update also seems to have fixed the disconnect wallet issue when switching between networks.

avclarke (approved):
Looks good

PR #1712, opened 2022-12-02
Futures market key update

Tburm:
Improve the futures refactor by using the `marketKey` as an index instead of the `marketAsset`. 

## Description
`marketAsset -> marketKey` is a "one-to-many" since we will have a V1 and V2 market with the same asset (`sETH`). Start migrating the UI to use the `marketKey` as the unique index for markets and positions.


avclarke (approved):
Looks good!

platschi (approved):
LGTM

PR #1722, opened 2022-12-05
Added atomic price feed

LeifuChen:
## Description
Atomic Swaps on Kwenta Exchange use the chainlink price. 
1. We should use the `value` from `effectiveAtomicValueAndRates` instead.
2. Show the `baseFeeRate` correctly on the page

## Related issue
#1051 

## Motivation and Context
Make the price feed more accurate for atomic swaps

## How Has This Been Tested?
Compare the price between the UI and contracts.
1. `ExchangeRates` on Ethereum
- https://etherscan.io/address/0xb4dc5ced63C2918c89E491D19BF1C0e92845de7C#readContract
3. Type the following currency pair into two functions ( `effectiveAtomicValueAndRates` is for atomic swap):
- `sourceKey`: 0x7343484600000000000000000000000000000000000000000000000000000000
- `destinationKey`: 0x7355534400000000000000000000000000000000000000000000000000000000
- `sourceAmount`: 1e18

image


sdk/contracts/abis/main.ts
```diff
@@ -12,6 +12,7 @@ export const SystemStatusABI = [
 
 export const ExchangeRatesABI = [
 	'function ratesForCurrencies(bytes32[] currencyKeys) external view returns (uint[])',
+	'function effectiveAtomicValueAndRates(bytes32 sourceCurrencyKey, uint256 amount, bytes32 destinationCurrencyKey) external view returns (uint value, uint systemValue, uint systemSourceRate, uint systemDestiationRate)',
```
avclarke, line 15: I think we can remove this, seems we're only using the other abi

state/exchange/actions.ts
```diff
@@ -184,6 +184,9 @@ export const resetCurrencyKeys = createAsyncThunk<any, void, ThunkConfig>(
 					sdk.exchange.getPriceRate(baseCurrencyKey, txProvider, coinGeckoPrices),
 				]);
 
+				// eslint-disable-next-line no-console
+				console.log(`reset`, Number(rate));
```
avclarke, line 188: Can we remove this log?

LeifuChen: Thanks for the catch.

PR #1741, opened 2022-12-12
Add Market Settings

Tburm:
Update the `FuturesMarket` entity to hold many more settings coming from the `PerpsV2MarketSettings` contract. We already call a single parameter, this change fetches them all

## Description
* Add settings to futures market data fetching
* Use delay times for executable/stale delayed orders
* Use market skew to adjust price by the premium/discount
* Update the futures rate to use this adjusted price instead of the oracle price

### DIFF
--- a/sections/futures/UserInfo/OpenOrdersTable.tsx
+++ b/sections/futures/UserInfo/OpenOrdersTable.tsx
@@ -15,7 +15,7 @@ import useNetworkSwitcher from 'hooks/useNetworkSwitcher';
 import { PositionSide } from 'queries/futures/types';
 import { DelayedOrder } from 'sdk/types/futures';
 import { cancelDelayedOrder, executeDelayedOrder } from 'state/futures/actions';
-import { selectMarketAsset, selectOpenOrders } from 'state/futures/selectors';
+import { selectMarketAsset, selectMarkets, selectOpenOrders } from 'state/futures/selectors';
 import { useAppDispatch, useAppSelector } from 'state/hooks';
 import { formatTimer } from 'utils/formatters/date';
 import { formatCurrency, formatDollars } from 'utils/formatters/number';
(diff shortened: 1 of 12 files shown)


sections/futures/UserInfo/OpenOrdersTable.tsx
```diff
@@ -47,36 +48,45 @@ const OpenOrdersTable: React.FC = () => {
...
-					dispatch(executeDelayedOrder(o.marketAddress));
-				},
-			}))
+			.map((o) => {
+				const market = futuresMarkets.find((m) => m.marketKey === o.marketKey);
+				const order = {
+					...o,
+					sizeTxt: formatCurrency(o.asset, o.size.abs(), {
+						currencyKey: getDisplayAsset(o.asset) ?? '',
+						minDecimals: o.size.abs().lt(0.01) ? 4 : 2,
```
avclarke, line 57: Could use `suggestedDecimals` helper here

sections/futures/UserInfo/OpenOrdersTable.tsx
```diff
@@ -47,36 +48,45 @@ const OpenOrdersTable: React.FC = () => {
...
+					}),
+					timeToExecution: countdownTimers ? countdownTimers[o.marketKey]?.timeToExecution : null,
+					timePastExecution: countdownTimers
```
avclarke, line 60: There's a lot of `countdownTimers[o.marketKey]` and null checks around here making this mapping feel a bit overloaded, could do `const timer = countdownTimers?.[o.marketKey]` and have the fields accept undefined as well as null, then would be cleaner to map like:
```
timePastExecution: timer?.timePastExecution,
timeToExecution: timer?.timeToExecution,
show: !!timer
```
... etc

avclarke (approved):
Looks good! Just a couple of cleanup suggestions but not blockers

PR #1742, opened 2022-12-12
Format Errors

Tburm:
Make a map of common errors to map to friendlier messages.

- [x] Move confirm order error into modal


components/Error/Error.tsx
```diff
@@ -21,7 +21,7 @@ type ErrorProps = {
 };
 
 export const FRIENDLY_I18N_MESSAGES: Record<string, string> = {
-	'Insufficient margin': 'futures.market.trade.margin.modal.deposit.min-margin-error',
+	'Insufficient margin': 'futures.market.trade.margin.modal.deposit.insufficient-margin',
```
avclarke, line 24: As this is under `modal.deposit` was it maybe initially used for when trying to deposit margin? As the new error wouldn't fit in that case.

Tburm: You're right, but in this case I can't actually get the original error to display anymore. I think it's cross margin specific since it does this check when you preview a position, and your position size could be too small.

I wonder how we handle this, since there are multiple scenarios where you could get the "Insufficient margin" error, but context specific.

PR #1745, opened 2022-12-14
Updated the 1inch API to v5 and optimized the display of price impact

LeifuChen:
## Description
1. Updated the 1inch API to v5
2. Move the price impact from `Currency Card` to `Summary Items`

## Related issue
#1736 

## How Has This Been Tested?
1. Open the exchange page
2. Select the `Quote Currency` as  `USDC` and `Base Currency` as `sUSD`
3. Check the display of the price impact 
4. Confirma the txn

image
image
image


LeifuChen:
> @LeifuChen I get a slippage amount when I quote a swap from `ETH -> sUSD` but I don't get one when I quote from `sUSD -> ETH` Any idea why?

This is because we only show the negative price impact.

Issue #1744, opened 2022-12-14
Refactor bug list


platschi:
### Subject of the issue
This ticket collects bugs/issues found after refactor implementations on `dev.kwenta.io`. More details/visuals can be found in the State Management forum thread on Discord:

- [x] Adjusting leverage fails in sending transaction. Also unclear error message on adjusting leverage modal when trying to send a tx
- [x] Closing position does not update transaction status/refetch position data, but UI is stuck on Close Position modal screen; same with deposit/withdraw modal
- [x] Dynamic fee value shown does not refresh after oracle price changes/updates [perps v1]
- [x] Setting leverage to a value other than 1x on a given market does not remember its leverage value setting when switching between markets.

E.g. if trader sets ETH-PERP to 5x, then switches to BTC-PERP, then back to ETH-PERP, the ETH-PERP leverage is reset to 1x. Would expect to remember 5x (same as prod currently)

PR #1749, opened 2022-12-14
Fix edit leverage modal and dynamic fee updates

avclarke:
Fixes 1, 3 & 4 here: https://github.com/Kwenta/kwenta/issues/1744

Also fixes a bug where the trading view chart wasn't clearing polling intervals, causing a memory leak and excessive queries


platschi:
Screenshot 2022-12-14 at 19 17 30

Only market is shows, I assume some tiny dynamic fee % still applying?

platschi (approved):

PR #1750, opened 2022-12-14
Disable order types

Tburm:
Disable all order types except delayed offchain orders. Remove the order type selection from the trade panel.

We can merge this after testing delayed offchain orders with keepers.


constants/futures.ts
```diff
@@ -3,8 +3,8 @@ import { wei } from '@synthetixio/wei';
 import { FuturesOrderType } from 'queries/futures/types';
 
 export const ISOLATED_MARGIN_ORDER_TYPES: FuturesOrderType[] = [
-	'delayed',
 	'delayed offchain',
```
avclarke, line 6: Not related to this PR but I think it could be good to maintain a consistent pattern for types with camel case format. It's not a great pattern to use spaces for things like types and keys and would be more consistent with the rest of our types.

avclarke (approved):

constants/futures.ts
Tburm: I noticed this pattern was carried over since we had `next price` and `stop market` orders already in here. I'm leaving a to-do and clean up separately.

Issue #1753, opened 2022-12-15
Exchange: Sort token list by holdings, then alphabetical


platschi:
### Subject of the issue
Currently, the list of tokens on Kwenta Exchange is sorted by holdings, and any tokens with 0 holdings are listed in a random order, making it difficult to quickly find the token one prefers.

![Image](https://user-images.githubusercontent.com/548702/207901056-5588de91-959f-44b1-a1e3-d07bef2a3565.png)

### Deliverables

- [x] Make sure token list is sorted by holdings first, then alphabetically (A-Z).

### Bounty
150 $OP

PR #1754, opened 2022-12-15
Sort exchange tokens by balance, then lexicographically

0xwalde:
Adds secondary sorting clauses to the token and synth lists in the `SelectCurrencyModal` based on token/synth name. This way the user's holdings will still be at the top, but anything with a 0 balance will have a stable alphabetical ordering below it for more predictability.

## Related issue
https://github.com/Kwenta/kwenta/issues/1753

## How Has This Been Tested?
Tested locally against optimism mainnet (since the list of tokens for exchange is very limited on Goerli)

kwenta-1753


platschi (approved):
Works fine on the deploy, list looks much better! Thank you.

LeifuChen (approved):
Looks good to me. Thanks for the quick implementation.

PR #1755, opened 2022-12-15
Component refactor: Part 2

koredefashokun:
## Description
This PR deals with deleting unused components, merging similar ones, and creating stories for components.

## Related issue
N/A

## Motivation and Context
This will help make the UI faster, and improve development velocity.

## How Has This Been Tested?
N/A


components/Text/index.ts
```diff
@@ -4,4 +4,4 @@ import { default as Data } from './Data';
 import { default as Heading } from './Heading';
 import { default as PositionType } from './PositionType';
 
-export default { Body, Data, Heading, BigText, PositionType };
+export { Body, Data, Heading, BigText, PositionType };
```
avclarke, line 7: What do you think of going with a few foundational text components of:
```
Heading
Body
Number
```
and potentially `Label`

And then all other sub styles becomes variants on those components?

koredefashokun: I think that `Number` can just be something like `<Body mono />` as well. I'm reluctant to remove BigText for now, even though it's technically equivalent to `Heading` with the added color and logo, because I don't want us to have the logo svg thing within the Heading component. Outside of that, I think the current state is pretty much equivalent to this. PositionType should eventually be taken out of the `Text` folder, but I still can't find the right place for it.

avclarke: Ok maybe we can rename `BigText` to `LogoText` or something and change any non logo use of BigText to a Heading? PositionType if it's specific to Futures we could move to under futures components?

koredefashokun: Good idea. I just pushed some code that does the first part. But since `PositionType` isn't specific to futures (we use it on the dashboard), I don't think we can move it under the futures components.

avclarke: If it's used in a futures related section of the dashboard I think it makes sense to add under futures still. If it's completely unrelated then maybe components/shared folder?

avclarke (commented):
Haven't had chance to try the StoryBook part yet but code looks great, just a couple of other suggestions.

components/InfoBox/InfoBox.tsx
```diff
@@ -57,6 +62,45 @@ const InfoBoxContainer = styled.div`
...
 	width: 100%;
 
+	.compact-box {
+		color: ${(props) => props.theme.colors.selectedTheme.rewardTitle};
+		font-size: 13px;
+		padding-left: 8px;
+		cursor: pointer;
+		margin-top: 16px;
+
+		.reward-copy {
```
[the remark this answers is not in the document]
koredefashokun: You're definitely right. Just noticed a bunch of these to be honest. Any style that does not directly affect the component should not be included.

sections/futures/FeeInfoBox/FeeInfoBox.tsx
```diff
@@ -76,10 +105,49 @@ const FeeInfoBox: React.FC = () => {
...
+								) : (
+									<div className="badge badge-red">
+										{t('dashboard.stake.tabs.trading-rewards.not-eligible')}
+										<NotEligibleIcon />
+									</div>
+								)}
+							</p>
+						</div>
+						<div>
+							<Paragraph className="reward-copy">
```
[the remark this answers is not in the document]
koredefashokun: Done. One of the next major things I'll be tackling is the use of `common.ts`. Many of those should either be base-level components or just deleted.

pages/_app.tsx
```diff
@@ -42,6 +44,18 @@ type AppPropsWithLayout = AppProps & {
 	Component: NextPageWithLayout;
 };
 
+process.env.GIT_HASH_ID!.toString();
+Sentry.init({
+	dsn:
+		'https://d48644bc80d04977a26132b346417210@o4504363236851712.ingest.sentry.io/4504363261362177',
+	maxBreadcrumbs: 50,
+	debug: process.env.NODE_ENV !== 'production',
+	release: 'kwenta-legacy@' + process.env.GIT_HASH_ID!.toString(),
```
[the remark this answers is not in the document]
koredefashokun: Merged the latest from perps-v2-dev

.github/workflows/audit_build_verify.yml
```diff
@@ -0,0 +1,128 @@
+name: Kwenta CI
```
avclarke, line 1: Not sure how or why this file was deleted

koredefashokun: I think it's one of the merged changes from the `perps-v2-dev` branch.

avclarke (approved):

.github/workflows/audit_build_verify.yml
platschi: It's currently broken, hence I removed it when cleaning up #1803 ; just noticed once I'm back I need to set up all the other checks for the new `perps-v2-dev` branch everywhere; or feel free to do so beforehand.

PR #1779, opened 2022-12-19
Position History

Tburm:
Add position history from the subgraph to the Redux state.

## Description
- Create state objects and selectors for position history
- Replace all instances of recoil position history with redux
- Delete the hook which is replaced by the SDK function

Note - There is one place that still calls a position history inside the component. This component is the `TraderHistory` that is displayed on the leaderboard, since it queries the relevant selected trader. Migration would require migrating leaderboard data to redux.


avclarke (approved):
Looks good!

PR #1800, opened 2022-12-21
Enable the estimated rewards for the current epoch

LeifuChen:
## Description
After applying the fix of #1789, the estimated rewards for the current epoch are disabled. 

So in this PR, we fix the following issues:
1. Enable the estimated rewards for the current epoch
- Solution: same as the change made in the `reward-estimator.ts` script, we reduce the distribution epoch by 1 to align the epoch dates and reward supply with the original design
2. Make sure the `claimMultiple` function works properly by skipping epoch 5.
- Solution: Starting from `epoch-5.json`, the index of the files is not the same as the epoch number. We add the index number by 1 before passing it (as the epoch number) to the parameters of the function call

## Related issue
#1789 

## Motivation and Context
Fixing the issue by minimizing the code change

## How Has This Been Tested?
1. Switch to the current epoch on the trading rewards tab. The estimated rewards can be seen
2. Switch to the optimism goerli. Then check the epoch number of the `epoch-4.json` and the `epoch-5.json`
- epoch-4.json: epoch number is 4
- epoch-5.json: epoch number is 6

image

### DIFF
--- a/sdk/services/kwentaToken.ts
+++ b/sdk/services/kwentaToken.ts
@@ -375,9 +375,11 @@ export default class KwentaTokenService {
 						this.sdk.context.networkId === 420 ? `goerli-` : ''
 					}epoch-${i}.json`
 			);
+
 		const responses: EpochData[] = await Promise.all(
-			fileNames.map(async (fileName, period) => {
+			fileNames.map(async (fileName, index) => {
 				const response = await client.get(fileName);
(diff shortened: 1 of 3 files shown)


JChiaramonte7 (approved):
LGTM but will let @platschi test

platschi (approved):

PR #1819, opened 2022-12-24
Fix limit orders not displaying

avclarke:
Also includes a fix for excessive keeper deposits


Tburm (approved):
This solution works. I think the better long term solution is implemented in perps v2, where we never use the `asset` value and always use the `marketKey`. But this will fix the issue on dev and we can move forward with the other pattern.

There will be no collisions here until we use our production subgraph on perps v2

Issue #1657, opened 2022-11-22
Review onboarding flow and integrate socket


avclarke:
When new users visit Kwenta we want to make it as simple as possible for them to begin using the app, this usually includes allowing them to bridge assets from other chains.

- Review onboarding process and evaluate a good place to allow for bridging assets (cross margin account creation, deposit to isolated margin, exchange page)
- Style the Socket widget where an example can be seen on [synthetix/bridge](https://staking.synthetix.io/bridge)

PR #1824, opened 2022-12-28
Integrated the socket plugin

LeifuChen:
## Description

1. Integrate the Socket widget into our onboarding flow (when user has less than 50 sUSD to deposit)
2. Style the Socket widget following our current theme design

## Related issue
#1657 

## Motivation and Context

Make the onboarding process more smoothly

### Test Case 1 - No wallet connected
Click on the `Deposit Margin` and the normal `Deposit Margin` modal pop-up.

### Test Case 2 - User with more than 50 sUSD
Click on the `Deposit Margin` and the normal `Deposit Margin` modal pop-up.

### Test Case 3 - User with less than 50 sUSD
1. Click on the `Deposit Margin` and the `Deposit` modal pop-up with the `Bridge & Swap` button 
2. Click on the `Bridge & Swap`, and the user can see the Socket widget

### Light theme
image
image
### Dark theme
image
image


sections/futures/Trade/TransferIsolatedMarginModal.tsx
```diff
@@ -26,10 +31,15 @@ type Props = {
...
 
+const SocketBridge = dynamic(() => import('../../../components/SocketBridge'), {
+	ssr: false,
+});
+
 const PLACEHOLDER = '$0.00';
 
 const TransferIsolatedMarginModal: React.FC<Props> = ({ onDismiss, defaultTab }) => {
 	const { t } = useTranslation();
+	const { walletAddress } = Connector.useContainer();
```
avclarke, line 42: I think we should make it good practice to fetch this from redux state (selectWallet) rather than connector, we'll probably remove it from connector eventually to make sure we only have a single data source. This would help with things like if we wanted to allow users to view the app from a selected wallet etc.

sections/futures/Trade/TransferIsolatedMarginModal.tsx
```diff
@@ -67,6 +88,8 @@ const TransferIsolatedMarginModal: React.FC<Props> = ({ onDismiss, defaultTab })
 		[amount, accessibleMargin]
 	);
 
+	const onChangeSocket = useCallback(() => setOpenSocket(!openSocket), [openSocket]);
```
avclarke, line 91: Maybe `onChangeShowSocket` ?

sections/futures/Trade/TransferIsolatedMarginModal.tsx
```diff
@@ -51,6 +62,16 @@ const TransferIsolatedMarginModal: React.FC<Props> = ({ onDismiss, defaultTab })
 		position?.accessibleMargin,
 	]);
 
+	const isLowBalance = useMemo(() => susdBalance.gt(zeroBN) && susdBalance.lt(minDeposit), [
+		susdBalance,
+		minDeposit,
+	]);
+
+	const isNoBalance = useMemo(() => susdBalance.eq(zeroBN), [susdBalance]);
```
avclarke, line 70: I think it might be nice to merge these two into a single memo `balanceStatus` with `'low_balance' | 'no_balance' | 'high_balance'`

avclarke (commented):
Looks good, just a couple of minor suggestions.

sections/futures/Trade/TransferIsolatedMarginModal.tsx
LeifuChen: Yes.Thanks for pointing it out.

sections/futures/Trade/TransferIsolatedMarginModal.tsx
LeifuChen: Better name. I will make the change.

sections/futures/Trade/TransferIsolatedMarginModal.tsx
LeifuChen: Agree. Will do it.

translations/en.json
```diff
@@ -781,13 +781,18 @@
...
 							"button": "Withdraw Margin"
 						},
 						"title": "Edit margin",
 						"actions": {
 							"withdraw": "Withdraw",
 							"deposit": "Deposit"
+						},
+						"bridge": {
+							"title": "Bridge",
+							"low-balance": "You sUSD balance is low. You can bridge more assets from another network below.",
```
avclarke, line 794: Typo `You` -> `Your`

LeifuChen: Thanks. I will correct it.

asherism (dismissed):

sections/futures/Trade/TradePanelHeader.tsx
```diff
@@ -31,19 +26,9 @@ export default function TradePanelHeader({ accountType, onManageBalance }: Props
 	const dispatch = useAppDispatch();
 	const theme = useTheme();
 
-	const { openConnectModal } = useConnectModal();
-	const wallet = useAppSelector(selectWallet);
 	const position = useAppSelector(selectPosition);
 	const positionStatus = useAppSelector(selectPositionStatus);
 	const balance = position ? position.remainingMargin : null;
-	const openModal = useAppSelector(selectOpenModal);
-	if (!wallet) {
```
avclarke: We can leave this as it was now that TransferIsolatedMarginModal is not below it any more.

LeifuChen: Reverted. Thanks.

pages/market.tsx
```diff
@@ -60,7 +66,17 @@ const Market: MarketComponent = () => {
 						<LeftSidebar />
 						<MarketInfo />
 						<StyledRightSideContent>
+							<TradePanelHeader
+								onManageBalance={() => dispatch(setOpenModal('futures_isolated_transfer'))}
+								accountType={'isolated_margin'}
```
avclarke, line 71: Can we move this back, as it is specific to IsolatedMargin?

avclarke (changes requested):
Looks good, I think we should just undo the TradePanelHeader change

pages/market.tsx
LeifuChen: Sure. I will revert this change. Let's minimize the change when introducing new features. My initial thought was to show the `deposit margin` when the user connected to our market page from a non-optimism chain. So they can start onboarding immediately by opening the `deposit margin` and bridging the token to optimism.

avclarke: Yea I see what you mean but I think this would need some more thought around correct UX. So best to leave how it was for now and we can create a design ticket if we think this approach makes sense.

LeifuChen: Reverted. Thanks.

components/Card/CardHeader.tsx
```diff
@@ -8,6 +8,7 @@ export type CardHeaderProps = {
 	className?: string;
 	lowercase?: boolean;
 	noBorder?: boolean;
+	onClick?: () => void;
```
koredefashokun, line 11: Do we use this?

components/SocketBridge/SocketBridge.tsx
```diff
@@ -0,0 +1,113 @@
...
+	const theme = useTheme();
+	const onBridgeSuccess = useCallback(() => {
+		dispatch(fetchBalances());
+	}, [dispatch]);
+
+	const background = hexToRGB(theme.colors.selectedTheme.input.secondary.background);
+	const text = hexToRGB(theme.colors.selectedTheme.button.text.primary);
+	const primaryButtonBg = hexToRGB(theme.colors.selectedTheme.socket.accent);
+
+	const customize: SocketCustomizationProps = {
```
koredefashokun, line 38: Maybe we can make a `utils/socket.ts` file, with a function that takes the theme object and returns the customization object.

sections/futures/Trade/TransferIsolatedMarginModal.tsx
```diff
@@ -51,6 +62,27 @@ const TransferIsolatedMarginModal: React.FC<Props> = ({ onDismiss, defaultTab })
 		position?.accessibleMargin,
 	]);
 
+	const balanceStatus: BalanceStatus = useMemo(
```
koredefashokun, line 65: We might be able to convert this into a Redux selector.

sections/shared/Layout/AppLayout/Header/NetworksSwitcher.tsx
```diff
@@ -66,11 +97,7 @@ const NetworksSwitcher: FC = () => {
 	}: ReactSelectOptionProps) => (
 		<ExternalLink href={link} onClick={onClick}>
 			<LabelContainer noPadding={!!prefixIcon}>
-				{prefixIcon === 'Optimism' && (
-					<PrefixIcon>
-						<OptimismIcon width={20} height={14} />
-					</PrefixIcon>
-				)}
+				{!!prefixIcon && activeChain && networkIcon(activeChain.name)}
```
koredefashokun, line 100: Instead of duplicating the `<PrefixIcon>` component in the `networkIcon` function above, we can use it here once like:

```
{!!prefixIcon && activeChain && (
  <PrefixIcon>
    {networkIcon(activeChain.name)}
  </PrefixIcon>
)}
```

sections/shared/Layout/AppLayout/Header/NetworksSwitcher.tsx
```diff
@@ -81,9 +108,7 @@ const NetworksSwitcher: FC = () => {
 	return !isL2 ? (
 		<Container onClick={openChainModal}>
 			<StyledButton noOutline size="sm">
-				<PrefixIcon>
-					<EthereumIcon width={20} height={14} />
-				</PrefixIcon>
+				{activeChain && networkIcon(activeChain.name)}
```
koredefashokun, line 111: If you decide to do the above, then we would have to change it here too.

koredefashokun (commented):
Looks good! Left a few small comments. That said, I can approve this immediately if we need to get it merged today. The comments are only to highlight things that would be nice-to-haves in the code.

components/Card/CardHeader.tsx
LeifuChen: Yes. We add an expand/collapse header to the deposit/withdraw modal.
https://github.com/Kwenta/kwenta/pull/1824/files#diff-b664ab95209131e3939e86b84aae29750a4edb86ac17568e4053c549ccda565cR148

sections/shared/Layout/AppLayout/Header/NetworksSwitcher.tsx
LeifuChen: Thanks. It's a better solution. Will make the change.

sections/shared/Layout/AppLayout/Header/NetworksSwitcher.tsx
LeifuChen: Thanks. Will make the change.

components/SocketBridge/SocketBridge.tsx
LeifuChen: Good idea. I can also move all the constants to `utils/socket.ts`.

koredefashokun (approved):

PR #1826, opened 2023-01-03
Responsive Price

Tburm:
Correct some price displays and make them responsive to offchain price updates.

## Description
* Update position selectors to recalculate unrealized pnl from the skew adjusted price
* Update header price to display skew adjusted price

### DIFF
--- a/state/futures/selectors.ts
+++ b/state/futures/selectors.ts
@@ -136,12 +136,35 @@ export const selectFuturesAccount = createSelector(
 );
 
 export const selectCrossMarginPositions = createSelector(
+	selectPrices,
 	(state: RootState) => state.futures,
-	(futures) => {
+	(prices, futures) => {
 		return futures.crossMargin.account && futures.crossMargin.positions[futures.crossMargin.account]
 			? futures.crossMargin.positions[futures.crossMargin.account].map(
 					// TODO: Maybe change to explicit serializing functions to avoid casting
-					(p) => deserializeWeiObject(p, futuresPositionKeys) as FuturesPosition
+					(p) => {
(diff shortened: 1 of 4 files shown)


state/futures/selectors.ts
```diff
@@ -136,26 +136,72 @@ export const selectFuturesAccount = createSelector(
...
+	selectPrices,
 	(state: RootState) => state.futures,
-	(wallet, futures) => {
+	(wallet, prices, futures) => {
 		if (!wallet) return [];
 		return futures.isolatedMargin.positions[wallet]
 			? futures.isolatedMargin.positions[wallet].map(
 					// TODO: Maybe change to explicit serializing functions to avoid casting
-					(p) => deserializeWeiObject(p, futuresPositionKeys) as FuturesPosition
+					(p) => {
```
avclarke, line 182: Could we extract this to a separate function as quite a bit of repeated code across these selectors?

avclarke (changes requested):
Looks good, just think it would be good to extract that duplicated selector logic

state/futures/selectors.ts
Tburm: Good point - I pulled the function into `utils`, logic for updating it is still in the selector.

Tburm: It still duplicates a bit of code, but if we ever need to add some specific logic to update cross margin positions we may need them separate.

avclarke: Not sure I follow why the whole function can't be pulled there as they seem identical?

Tburm: You're right... I was thinking about a future state where we may want to handle these position differently mapping the exact same function might not be ideal. But I don't have a specific reason, so I agree it's best to move the whole thing.

avclarke (approved):

PR #1827, opened 2023-01-03
`TransferPanelHeader` Display

Tburm:
Update the logic for displaying the options for the `TradePanelHeader`. Only display the yellow "Deposit Margin" button if we have checked the balance to avoid flashing before contracts are loaded.


avclarke (approved):
Nice fix, looks good!

PR #1832, opened 2023-01-05
Price update color

Tburm:
Add a responsive color to the price on each price update. 

## Description
* [x] Show color on `MarketDetails` for offchain price
* [x] Show color on `PositionCard` for offchain price
* [x] Show colors on `FuturesMarketsTable` for offchain price


sections/dashboard/FuturesMarketsTable/FuturesMarketsTable.tsx
```diff
@@ -436,6 +437,15 @@ const StyledText = styled.div`
 	font-family: ${(props) => props.theme.fonts.bold};
 `;
 
+const ColoredPrice = styled.div<{ color: string }>`
+	color: ${(props) =>
+		props.color === 'red'
+			? props.theme.colors.selectedTheme.red
+			: props.color === 'green'
+			? props.theme.colors.selectedTheme.red
```
avclarke, line 445: is `props.color === 'green' ? props.theme.colors.selectedTheme.red` meant to be `props.theme.colors.selectedTheme.red` 

And if we had the prop `color: PriceColor ` we could just do `props.theme.colors.selectedTheme[props.color]`

state/futures/selectors.ts
```diff
@@ -110,6 +110,17 @@ export const selectMarketPrice = createSelector(
 	}
 );
 
+export const selectMarketPriceColor = createSelector(
+	selectMarketInfo,
+	(state: RootState) => state.prices,
+	(marketInfo, { offChainPriceColors }) => {
+		if (!marketInfo || !offChainPriceColors[marketInfo.asset]) return 'white';
+		const color = offChainPriceColors[marketInfo.asset]?.color;
+		const expiresAt = offChainPriceColors[marketInfo.asset]?.expiresAt;
```
avclarke, line 119: I don't think this works as expected, as if I understand correctly seems you want the color to only flash for 1 second and then expire back to white but there is no event to trigger this selector again, it will only be called on the next prices update and watching the color change intervals I'm noticing it vary quite a bit.

If we really want it to expire back to white after 1 second then we'll need some other update to trigger the change, maybe in `getPriceColors` we add a setTimeout trigger and fire another action to `resetPriceColors` or something?

sections/dashboard/FuturesMarketsTable/FuturesMarketsTable.tsx
```diff
@@ -436,6 +437,15 @@ const StyledText = styled.div`
 	font-family: ${(props) => props.theme.fonts.bold};
 `;
 
+const ColoredPrice = styled.div<{ color: string }>`
```
avclarke, line 440: Should we make this a dedicated component and move to components folder, then we could use it throughout for price display. Also if we wanted to add some subtle effect like fade back to white it would be contained in one component then.

sections/futures/PositionCard/PositionCard.tsx
```diff
@@ -290,7 +292,7 @@ const PositionCard: React.FC<PositionCardProps> = () => {
 				<DataCol>
 					<InfoRow>
 						<StyledSubtitle>{data.marketShortName}</StyledSubtitle>
-						<StyledValue>{data.marketPrice}</StyledValue>
+						<StyledValue className={marketPriceColor}>{data.marketPrice}</StyledValue>
```
avclarke, line 295: If we create a shared `PriceColor` component like mentioned above could we just use that here instead?

avclarke (commented):
Nice work, it's a great addition! Just left a couple of suggestions.

state/futures/selectors.ts
Tburm: Yeah you're right - I actually had written a version with a reset before the `expiresAt` version. It also wasn't working as I expected at the time, but it may have been before I set up the selectors. I'll try that.

Tburm: I found that with a reset based on a timeout sometimes we get a new price update just before the reset. This causes the opposite problem, where the color flashes for <1 second.

Maybe we need to combine the two solutions, we need some way to trigger the selector.

Tburm: A reset which only changes any non-expired colors is probably the way. That way if we get a price update before the timeout, the reset won't change the color.

sections/dashboard/FuturesMarketsTable/FuturesMarketsTable.tsx
Tburm: This type doesn't exactly work because we pass it through the data in the table. I opted to just use `'white` | 'red' | 'green'` here even though we create this type elsewhere.

state/prices/selectors.ts
```diff
@@ -4,22 +4,47 @@ import { wei } from '@synthetixio/wei';
...
+				merged[key] = {
+					offChain: wei(price),
+				};
+			}
+		});
+		return merged;
+	}
+);
+
+export const selectPriceColors = createSelector(
```
avclarke, line 33: I'm not sure we need this extra selector, we could just have the `ColoredPrice` label accept a `PriceInfo` object and it decide the color based on change?

state/prices/types.ts
```diff
@@ -1,13 +1,32 @@
...
+	price: T;
+	change: PriceChange;
+};
+
+export type PriceColor = {
+	offChain?: PriceColorOptions;
+	onChain?: PriceColorOptions;
+};
+
+export type PriceColors = Record<string, PriceColor>;
```
avclarke, line 19: Related to the previous comment I don't think we need this type and the selector it's used for.

sdk/constants/prices.ts
```diff
@@ -27,4 +27,4 @@ export const PYTH_IDS = {
 };
 
 // Allow to be set from config so users can customise
-export const PRICE_UPDATE_THROTTLE = 5000;
+export const PRICE_UPDATE_THROTTLE = 500;
```
avclarke, line 30: What do you think about making this 1000?

avclarke (commented):
Much simpler now, just wondering if we can also remove the prices to color mapping selector, seems unnecessary?

state/prices/selectors.ts
Tburm: This works, but I had to make a function to export for `MarketDetails` to convert `PricesInfo` into a color string since we don't use that component.

avclarke (approved):
Looks good!

PR #1836, opened 2023-01-06
Fixed the wrong reward badge after the market page refresh

LeifuChen:
## Description
The 'Eligible' text switches to 'Not Eligible' after refreshing the market page even though the user has staked Kwenta.

## How Has This Been Tested?
Case 1: Disconnected the wallet
1. Show `Not Eligible` badge
2. Same text after page refresh

Case 2: Connected the wallet with staked Kwenta
1. Show `Eligible` badge
2. Same text after page refresh

Case 3: Connected the wallet with no staked Kwenta: 
1. Showing `Not Eligible` badge
2. Same text after page refresh

### DIFF
--- a/hooks/useFuturesData.ts
+++ b/hooks/useFuturesData.ts
@@ -91,7 +91,6 @@ const useFuturesData = () => {
 	const { tradeFee: crossMarginTradeFee, stopOrderFee, limitOrderFee } = useAppSelector(
 		selectCrossMarginSettings
 	);
-
 	const isAdvancedOrder = useAppSelector(selectIsAdvancedOrder);
 	const marketAssetRate = useAppSelector(selectMarketAssetRate);
 	const orderPrice = useAppSelector(selectCrossMarginOrderPrice);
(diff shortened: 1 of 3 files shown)


hooks/useFuturesData.ts
```diff
@@ -293,6 +295,7 @@ const useFuturesData = () => {
 
 	useEffect(() => {
 		resetTradeState();
+		dispatch(fetchStakingData());
```
avclarke, line 298: I think the best place for this is to be added to `usePollMarketFuturesData` in `state/futures/hooks` you can probably add a line like:

`useFetchAction(fetchStakingData, { changeKeys: [networkId, wallet] });`

avclarke (commented):
Looks good, just think we can move the fetch to the place that handles all the futures data fetches.

hooks/useFuturesData.ts
LeifuChen: Sure. I will make the changes.

avclarke (approved):

PR #1837, opened 2023-01-06
Alternative leverage input fix

avclarke:
Also fixes an error of fetching max fee


Tburm (approved):
This works as expected and much simpler than my version 👍🏼 

Will still need to resubmit one bug I fixed on my PR where errors codes changed.

avclarke:
> This works as expected and much simpler than my version 👍🏼
> 
> Will still need to resubmit one bug I fixed on my PR where errors codes changed.

Ok sounds good, will just merge this to yours and then we can merge yours with the error code changes

PR #1852, opened 2023-01-09
Perps V1: Switch to testnet subgraph

Tburm:
Update the rates subgraph to use the testnet subgraph. This is a temporary fix while the mainnet subgraph lags behind blocks.

Description
* Update rates subgraph to OP Goerli
* Update the candles query to use the display asset name. This solves an issue where the sLINK candles return the ETH price


avclarke (approved):
Looks good!

PR #1864, opened 2023-01-10
Perps V2: Fix price feeds

Tburm:
Point perps v2 at the OP Goerli price feed to avoid the lagging price feeds. Additionally updates some other subgraph-related configurations.

## Description
* Update lagged prices to match V1 implementation
* Change rates subgraph to point at testnet
* Update futures subgraph to point at the correct perps version
* Update an async function which was throwing errors


avclarke (approved):
Looks good!

Issue #1873, opened 2023-01-11
Fix broken WBTC logo on Exchange


platschi:
### Subject of the issue
The WBTC fetched from our synthswap/ 1inch exchange integration does not work (file not found). We should upload it to our repository and fetch it from there in order to not have a failing icon on the frontend, similar to what we do with the Kwenta token icon.

![Screenshot 2023-01-11 at 17 51 17](https://user-images.githubusercontent.com/548702/211915084-a9a0fe0b-c97f-4fab-a719-e2edea8bf357.png)

![Screenshot 2023-01-11 at 17 53 10](https://user-images.githubusercontent.com/548702/211915507-38f433e6-7b41-413c-8ed9-9d1b20cc2cfc.png)

### Deliverables
- Fix WBTC logo on Exchange

### Bounty
80 $OP

PR #1875, opened 2023-01-11
Add self-hosted WBTC logo and use it in currency icons

0xwalde:
Adds the WBTC logo to static assets and puts in an override in `utils/icons.ts` to reference that instead of the image url from the 1inch token list. 

## Related issue
https://github.com/Kwenta/kwenta/issues/1873

## How Has This Been Tested?
Tested locally against Optimism Mainnet

Kwenta-1873


LeifuChen (approved):
Looks good. Thanks.

PR #1882, opened 2023-01-13
Use native, right-clickable anchors for sidebar links

0xwalde:
Replaces the button-based `NavButton.tsx` component with a link-based `NavLink.tsx` component for links in the sidebar of the `DashboardLayout.tsx`. This should keep the exact same look-and-feel and UX while enabling native browser behavior when interacting with these links.

## Motivation and Context
HTML `buttons` don't allow some built-in browser actions such as middle-clicking, right-clicking, and command/control-clicking to open the page they direct to in a new tab. Since these are effectively just links (they aren't doing anything special besides navigation), they should probably be rendered as native anchor tags to support these actions.

## How Has This Been Tested?
Test locally against Optimism mainnet and goerli.

Kwenta-sidebar-links


avclarke (approved):
Looks good!

platschi (approved):

Issue #1859, opened 2023-01-10
Rainbowkit: Avoid using default alchemy rpc fallback


platschi:
### Subject of the issue
Users report that they get rate limited on Kwenta through the rainbowkit default public alchemy rpc. Leifu reported that this is the default rpc for rainbowkit.

However, the rpc should first use infura instead of alchemy if BlastAPI might fail. There might be some misconfiguration or other issue at hand here overriding Kwenta settings.

### Deliverables
- Find the cause of the issue
- Fix the fallback provider to be in order of: Blast, Infura, public
- Display an error if a fallback provider is used which might result in degraded performance / rate limits

PR #1884, opened 2023-01-13
Initializes contracts with provider instead of signer and improves orders query

avclarke:
## Description
Fixes an issue where users own rpc was being used to make requests, resulting in rate limiting and other problems.

## Related issue
#1859


platschi (approved):
Tested with a couple of mainnet transactions, no rpc errors to be seen anywhere.

Issue #1596, opened 2022-11-11
Exchange: Locale settings break input fields


platschi:
### Subject of the issue
International users report that their locale settings break the input fields of the exchange, making it unusable.

Comments:

> Im from europe and it helped to change windows regional setting of numbers where you set you decimal from "," to "." and reboot computer

### Screenshots

![spot1](https://user-images.githubusercontent.com/548702/201344183-aa194144-07ed-4588-bea5-ede88e9378f5.png)

### Deliverables
- Investigate issue and research possible solutions
- Implement best solution

platschi:
Some users also report that the ETH/sETH swap is off for them, not sure this is related:

![image](https://user-images.githubusercontent.com/548702/201661291-630c3ea5-3557-4ec4-8c87-9350ed37767f.png)

PR #1842, opened 2023-01-06
Component refactor (Part 3)

koredefashokun:
## Description
This PR continues previous efforts to refactor existing components in the codebase. In particular, It removes duplicated components, and extracts a number of them to base components. It also includes very minor performance improvements (moving props around and cleaning up certain hook instances). In the future, performance will be tackled in a more comprehensive manner.

## Related issue
N/A

## Motivation and Context
This is part of a series of changes that aim to improve Kwenta's developer and user experiences.

## How Has This Been Tested?
N/A

PR #1889, opened 2023-01-16
Fix input locale issue

koredefashokun:
## Description
This PR attempts to fix the locale-related issues with inputs across the Kwenta interface.

## Related issue
#1596 

## Motivation and Context
This ensures that users in European regions do not have issues entering values in the UI.

## How Has This Been Tested?
N/A

### DIFF
--- a/components/Input/NumericInput.tsx
+++ b/components/Input/NumericInput.tsx
@@ -44,6 +44,7 @@ const NumericInput: FC<NumericInputProps> = memo(
 				defaultValue={defaultValue}
 				disabled={disabled}
 				id={id}
+				lang="en-US"
 			/>
 		);
 	}


avclarke (approved):

platschi (changes requested):
Unfortunately, this did not change the input. Noticing different behavior between Firefox and Chrome, so browsers might forcefully overwrite the set locale. 

E.g. Firefox displays `5.8995`, while Chrome displays `5,8995` on the deploy for this PR.

Might need to investigate if there's a way to avoid that?

LeifuChen:
> Unfortunately, this did not change the input. Noticing different behavior between Firefox and Chrome, so browsers might forcefully overwrite the set locale.
> 
> E.g. Firefox displays `5.8995`, while Chrome displays `5,8995` on the deploy for this PR (tested on Exchange)
> 
> Might need to investigate if there's a way to avoid that?

Chrome and Firefox define the locale differently.
- Chrome: by the browser’s language.
- Firefox: by the “lang” attribute (as @koredefashokun did).

We may try sth similar as https://github.com/yairEO/react-number-input

platschi:
This has been fixed with #1842 by using `text`

Issue #1453, opened 2022-10-05
Improve feedback system


asherism:
Create a toaster component to improve error handling

Ref:
![Image](https://user-images.githubusercontent.com/11524787/194080458-93894b38-04d5-4f02-9c54-fc29beaeda6e.png)

platschi:
See #1298 for more info

asherism:
<!-- Uploading "image.png"... —>

https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=6300%3A31499

asherism:
https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=6300%3A31499


![Image](https://user-images.githubusercontent.com/11524787/200010584-e423d56f-8641-4f3d-9aba-a6e8b621edde.png)

Issue #1298, opened 2022-08-18
Improve overall app error handling and display


avclarke:
### Feature Proposal
Improved error feedback 

### Rationale
Error displays are currently inconsistent, can contain long unfriendly error messages and not efficient to implement.

### Implementation
Implement an error notification component, which would overlay the whole app (similar to the tx notifier) and receive notification events from any component. Could be good to use React portals for this.

The error notifier would also handle parsing messages from smart contract failures and will take a more generic fallback message in the case it receives an error it cannot parse. Example function signature on the error handler:

```
notifyError(error: Error, fallback: string)
```

### Error message mapping

There are a number of errors that can be returned during transactions that are currently just shown to the users as generic failed, leaving them unaware of what actually went wrong and difficult for us to debug. To improve the feedback we should identify specific errors we are aware of and feedback more helpful and user friendly messaging. 

**Generic erros**
```INSUFFICIENT_FINDS ``` - Not enough ETH to cover gas for the tx

**Exchange**
```Surpassed volume limit``` - User cannot exchange more than 100k sUSD when doing atomic swaps

**Futures**
```Max market size exceeded``` - OI cap reached

platschi:
Related and includes a list of common errors: #491

platschi:
Design ready in #1453

PR #1897, opened 2023-01-18
Cross margin accounts and other futures / state related refactor

avclarke:
**Refactor futures state logic**
This moves the remaining futures recoil state over to redux along with migrating the related queries to the sdk.

**Account separation**
Cross margin account specific data is now nested by account and network, this means as users switch networks and wallets the display data will change immediately to match the network / account context.

**Implement Redux persist**
Redux persist allows us to persist state between sessions by storing in local storage and hydrating state when the app is loaded.

_State migrations - Now that state is persisted it's important when any existing state changes are made a [migration](https://github.com/rt2zz/redux-persist#migrations) is included to avoid errors occurring._ 

**Implement Error Notifications**
There is now a function called `notifyError` which takes a message as parameter and will display an error notification to the user. This uses a similar toast component to our transaction components. This relates to issues #1298 #1453 but there needs to be further work to fully implement the error notifications throughout the app.

closes: #1856


sections/futures/TradeCrossMargin/EditCrossMarginLeverageModal.tsx
```diff
@@ -49,9 +48,8 @@ type DepositMarginModalProps = {
 
 export default function EditLeverageModal({ onDismiss, editMode }: DepositMarginModalProps) {
```
Tburm, line 49: Having trouble with this component when I don't have a position open.

With no position, I click "edit" to open the modal. When I click confirm, it throws an error using the notifier. I can't click on screen to dismiss the modal, and clicking the X button doesn't dismiss the modal.

sections/futures/TradeCrossMargin/EditCrossMarginLeverageModal.tsx
```diff
@@ -122,56 +119,8 @@ export default function EditLeverageModal({ onDismiss, editMode }: DepositMargin
 	);
 
 	const onConfirm = useCallback(async () => {
-		setError(null);
-		if (editMode === 'existing_position' && position?.position) {
```
Tburm, line 126: losing this logic means it tries to submit a transaction when in "new position" mode.

sections/futures/TradeCrossMargin/EditCrossMarginLeverageModal.tsx
```diff
@@ -122,56 +119,8 @@ export default function EditLeverageModal({ onDismiss, editMode }: DepositMargin
...
-						},
-					});
-				}
-			} catch (err) {
-				setSubmitting(false);
-				setError(t('common.transaction.transaction-failed'));
-				logError(err);
-			}
-			resetTradeState();
-		} else {
```
Tburm, line 155: I think we need to preserve this logic

sections/futures/PositionCard/PositionCard.tsx
```diff
@@ -89,10 +88,8 @@ const PositionCard: React.FC<PositionCardProps> = () => {
 			: undefined;
 
 	const thisPositionHistory = useMemo(() => {
-		return positionHistory[futuresAccountType].find(
-			({ asset, isOpen }) => isOpen && asset === marketAsset
-		);
-	}, [positionHistory, marketAsset, futuresAccountType]);
+		return positionHistory.find(({ asset, isOpen }) => isOpen && asset === marketAsset);
+	}, [positionHistory, marketAsset]);
```
Tburm, line 92: We reuse this logic in a few different components, maybe should just have a `currentPositionHistory` selector that takes care of this logic in state.

sections/futures/Trade/ManagePosition.tsx
```diff
@@ -8,15 +8,14 @@ import Loader from 'components/Loader';
 import { useFuturesContext } from 'contexts/FuturesContext';
 import { previewErrorI18n } from 'queries/futures/constants';
 import { setOpenModal } from 'state/app/reducer';
-import { selectOpenModal } from 'state/app/selectors';
```
Tburm, line 11: Comment is for this component. We display the `orderError` here, should we still display transaction errors even though they would display in the notifier:
image

Not related to changes you made, but an improvement to displaying the unformatted errors.

sections/futures/TradeCrossMargin/EditCrossMarginLeverageModal.tsx
Tburm: Specifically when opening this way:
image

When opened from an open position it works as expected.

queries/futures/types.ts
```diff
@@ -25,37 +25,6 @@ export type MarginTransfer = {
...
-	side: PositionSide;
-	pnl: Wei;
-	pnlWithFeesPaid: Wei;
-	totalVolume: Wei;
-	trades: number;
-};
-
 export enum PositionSide {
 	LONG = 'long',
 	SHORT = 'short',
```
Tburm, line 30: Looked like we were still importing this one in some places, might be worth removing this as well and always importing from SDK.

sections/futures/PositionCard/PositionCard.tsx
avclarke: yea good call

sections/futures/Trade/ManagePosition.tsx
avclarke: Yea makes sense, I've removed the transaction error for now but eventually when we do some more clean up around errors we probably want to push all these order / preview errors through the notifier.

state/store.ts
```diff
@@ -46,4 +73,8 @@ export type AppThunk<ReturnType = void> = ThunkAction<
 	AnyAction
 >;
 
+export const persistor = persistStore(store);
+
+// persistor.purge();
```
koredefashokun, line 78: I think you forgot this comment.

state/types.ts
```diff
@@ -11,7 +11,7 @@ export enum FetchStatus {
 
 export type QueryStatus = {
 	status: FetchStatus;
-	error: string | null | undefined;
+	error?: string | null | undefined;
```
koredefashokun, line 14: We can omit the explicit `undefined` since this is optional.

koredefashokun (commented):
Changes look very good! I left a few comments around small things, but I'm pretty excited about the changes (especially about getting rid of the `RefetchContext`).

sdk/utils/futures.ts
```diff
@@ -86,7 +84,7 @@ export const marketsForNetwork = (networkId: number) => {
 		case 420:
 			return TESTNET_MARKETS;
 		default:
-			logError('Futures is not supported on this network.');
+			logError(new Error('Futures is not supported on this network.'));
```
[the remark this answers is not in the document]
koredefashokun: By the way, I noticed why this is this way, so you can ignore.

sdk/queries/futures.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { getFuturesEndpoint } from 'sdk/utils/futures';
+
+export const queryAccountsFromSubgraph = async (
+	networkId: NetworkId,
+	walletAddress: string | null
+): Promise<string[]> => {
+	if (!walletAddress) return [];
+
+	const futuresEndpoint = getFuturesEndpoint(networkId);
```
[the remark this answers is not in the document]
avclarke: Yea good call, there's already a getter on the futures sdk service `futuresGqlEndpoint` so should probably just use that instead.

components/Error/ErrorNotifier.tsx
```diff
@@ -0,0 +1,66 @@
...
+	if (!message) return '';
+	if (message.includes('insufficient funds for intrinsic transaction cost'))
+		return 'Insufficient ETH balance for gas cost';
+	return message;
+};
+
+const StyledToastContainer = styled(ToastContainer)`
+	.Toastify__toast-container {
+		background-color: ${(props) => props.theme.colors.navy};
+		border: 1px solid ${(props) => props.theme.colors.green};
```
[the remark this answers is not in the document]
avclarke: Seems these styles aren't actually relevant, will just remove, good spot though was some copy paste

sdk/queries/futures.ts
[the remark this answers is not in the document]
avclarke: Thinking about this some more I'm going to leave for now as the next step will be to move all these queries over to the sdk anyway.

platschi (approved):

Tburm (commented):
All looks good except one issue. I opened this modal to adjust leverage on an open cross margin position:
image

It doesn't look like there is a way to close the modal. The X doesn't work, clicking off the modal doesn't work.

platschi (approved):

PR #1900, opened 2023-01-20
Query funding rates from market page

avclarke:
Fixes funding rate not displaying on futures


LeifuChen (approved):
Thanks for the quick fix.

Issue #1878, opened 2023-01-12
Perps V2 Outstanding Issues


Tburm:
## Priority

- [ ] Leifu: #1974
- [x] Troy: Add responsive colors on price updates (#1832)
- [x] Adam: #1839 (PR: #1973)

## Refactor:

- [x] #1958

## Low priority
- [ ] Define PnL calculations and show PnL in Trade History
- [ ] #1935 
- [x] Improve markets dropdown for long list of markets

## Done
- [x] Enable offchain order execution (#1914)
- [x] Add button loading state for order execution/cancellation (#1914)
- [x] Fix trade panel "Deposit margin" display logic (#1919)
- [x] Remove hard-coded frontend fix for funding [link](https://github.com/Kwenta/kwenta/blob/e63cc50326f33553abcd03a7e73c72f3eb8c6caa/sections/futures/MarketDetails/useGetMarketData.ts#L52) (#1918)
- [x] #1904 
- [x] Update delayed order query to use multicall [link](https://github.com/Kwenta/kwenta/blob/4cd3fa77285ee6c2482f7772be1d4741a711cbdd/state/futures/actions.ts#L319) (#1925)
- [x] Update "oracle price" to "mark price" on dashboard (#1923)
- [x] Clean up order type naming (`delayed offchain` -> `delayedOffchain`) (#1922)
- [x] Update 1H Funding Rate tooltip text to explain new velocity-based model (#1939)
- [x] #1930 (#1940)
- [x] Reenable margin preview (#1931)
- [x] #1932 
- [x] Decide if we want to kick out (broken) Spot Markets table
- [x] Landing Page: ETH market data incorrect (100% price change, no volume shown) [fixed in #1926)

PR #1914, opened 2023-01-26
Order Execution Logic

Tburm:
Add logic to execute orders using Pyth from the frontend if a keeper fails to execute the order.

## Description
* Add Pyth contract interface
* Add logic for executing offchain orders
* Remove "Commit deposit" from orders table
* Add "Actions" column with logic to display Cancel/Execute buttons
* Add a configurable delay before the execute button is displayed (currently 30 seconds after executable)

## Related issue
- #1878 

## Motivation and Context
Intermittent keepers create a poor experience for traders, who lose deposits when keepers don't pick up their orders. Creating a fallback is necessary to give traders the ability to execute when keepers fail.

## How Has This Been Tested?
This is difficult to test due to keepers being live. I tested by reducing the delay to 0 seconds and racing the keeper to execute. These transactions were confirmed.


Tburm:
@platschi just a heads up that some weird linting changes got included in this PR when I used `npm run lint:fix` as well as some updates to contract interfaces when generating new types.

sections/futures/UserInfo/OpenOrdersTable.tsx
```diff
@@ -84,7 +102,14 @@ const OpenOrdersTable: React.FC = () => {
 						);
 					},
 					onExecute: () => {
-						dispatch(executeDelayedOrder(o.marketAddress));
+						if (!o.marketKey) return null;
```
avclarke, line 105: Not really related to this PR but is there a reason for marketKey to be nullable? Seems we should never have an order with no market

sections/futures/UserInfo/OpenOrdersTable.tsx
```diff
@@ -117,6 +140,10 @@ const OpenOrdersTable: React.FC = () => {
 				return acc;
 			}, {} as CountdownTimers);
 			setCountdownTimers(newCountdownTimers);
+		};
+
+		const timer = setInterval(() => {
```
avclarke, line 145: We should prob clear this interval on unmount, something like:
```
useEffect(() => {
  const id = setInterval(updateTimers, 1000);
  return () => {
    clearInterval(id);
  };
}, []);
```

avclarke (changes requested):
Looks good, just a couple of minor comments.

sections/futures/UserInfo/OpenOrdersTable.tsx
Tburm: Fixed this with a comment above. It's because we don't want to pass the entire market to the SDK query, so I added a few nullable fields that get hydrated when the orders come back. I decided to just make them required. It would only exclude orders where we don't have markets, which isn't actually possible.

sections/futures/UserInfo/OpenOrdersTable.tsx
avclarke: Yep sorry not sure how I missed that, all good 😅

sdk/services/prices.ts
```diff
@@ -139,6 +139,15 @@ export default class PricesService {
 		return this.formatOffChainPrices(pythPrices ?? []);
 	}
 
+	public async getPriceUpdateData(marketKey: FuturesMarketKey) {
```
avclarke, line 142: I think this function name could do with being a bit more specific as a public sdk call, maybe `getPythPriceUpdateData`?

avclarke (commented):
Looks good, only suggestion is we update the function name for fetching pyth price data just so it's not confusing for other users when the sdk becomes public.

sdk/services/prices.ts
Tburm: Good catch, updated this

avclarke (approved):
Looks good

platschi (approved):

PR #1917, opened 2023-01-26
Fix OP Goerli chain config

Tburm:
Limits the check for active chains to the chain ID. Currently this comparison will return false for OP Goerli, resulting in a mainnet provider being connected.


platschi (approved):
Works as expected on both OP Goerli and OP mainnet

PR #1919, opened 2023-01-26
Deposit button logic

Tburm:
There is some flickering with the "Deposit Margin" button which can be improved. The current logic only displays the yellow `Deposit Margin` button when the position query has succeeded. This causes flickering when the query is in progress.

## Description:
* Update futures position query to return values where margin is zero. This will help distinguish when the query has _not been run_ versus when it has _succeeded_
* Update logic to ONLY display `Deposit Margin` button when we have confirmed the user has $0 margin

Note that the "startup" behavior for a connected wallet with no margin can still look strange: Display "Connect wallet" -> Display $0 -> Display "Deposit Margin"

### DIFF
--- a/sections/futures/Trade/TradePanelHeader.tsx
+++ b/sections/futures/Trade/TradePanelHeader.tsx
@@ -10,9 +10,8 @@ import { NumberDiv } from 'components/Text/NumberLabel';
 import { EXTERNAL_LINKS } from 'constants/links';
 import { FuturesAccountType } from 'queries/futures/subgraph';
 import { setOpenModal } from 'state/app/reducer';
-import { selectPosition, selectPositionStatus } from 'state/futures/selectors';
+import { selectPosition } from 'state/futures/selectors';
 import { useAppDispatch, useAppSelector } from 'state/hooks';
-import { FetchStatus } from 'state/types';
 import { selectWallet } from 'state/wallet/selectors';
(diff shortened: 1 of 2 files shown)


Tburm:
@avclarke Do you see any problems with storing positions when the user has 0 margin? Right now we filter them out, instead this will store values where `position` is null and `remainingMargin` is a wei value of zero.

sections/futures/Trade/TradePanelHeader.tsx
```diff
@@ -41,7 +39,7 @@ export default function TradePanelHeader({ accountType, onManageBalance }: Props
 		);
 	}
 
-	if (!balance && positionStatus.status === FetchStatus.Success) {
+	if (!!balance && balance.eq(0)) {
```
avclarke, line 42: I'd done some similar changes in the refactor PR which covers some extra cases for cross margin but yea this looks good and when I merge the refactor PR I'll make sure to bring back the case for cross margin.

avclarke:
> @avclarke Do you see any problems with storing positions when the user has 0 margin? Right now we filter them out, instead this will store values where `position` is null and `remainingMargin` is a wei value of zero.

I don't see any issues storing positions with zero margin, these changes make sense.

avclarke (approved):

platschi (approved):
Already smoother, thanks!

PR #1923, opened 2023-01-27
Add mark prices to state

Tburm:
Add a selector with skew adjusted prices for all markets. Update the dashboard to display these prices, along with some other type improvements.

## Description
* Add a `selectMarkPrices` selector containing a map of market keys to skew-adjusted prices
* Update `FuturesMarketTable` to use mark prices
* Update copy to "Mark price"
* Improve `FuturesMarket` type by making `openInterest` a required field (it's always available)


platschi (approved):
Small but great improvement on the Dashboard!

Issue #1858, opened 2023-01-10
Earn UI: Show APY


platschi:
### Subject of the issue
> https://kwenta.eth.limo/dashboard/earn/ should show APY for LP staking. Should be easy enough to calculate given amount of KWENTA allocated per period and total amount of LP staked. Thanks.

### Your environment
* Network used (Mainnet? L2?)
* Which browser and its version

### Expected behaviour
Show APY for LP staking

### Actual behaviour
Shows nothing

PR #1924, opened 2023-01-27
Added APY to Earn UI

LeifuChen:
## Description
Show APY for LP staking.

## Related issue
#1858 

## Motivation and Context
To attract more LP to increase the liquidity.

image
image
image


constants/currency.ts
```diff
@@ -79,7 +79,10 @@ export const sUSD_EXCHANGE_RATE = new Wei(1);
 export const SYNTH_DECIMALS = 18;
 
 export const ETH_ADDRESS = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee'; // For 1inch API
-export const ETH_COINGECKO_ADDRESS = '0x4200000000000000000000000000000000000006'; // For coingecko API
+
+// For coingecko API
+export const ETH_COINGECKO_ADDRESS = '0x4200000000000000000000000000000000000006';
+export const KWENTA_COINGECKO_ADDRESS = '0x920cf626a271321c151d027030d5d08af699456b';
```
avclarke, line 85: Not sure 'COINGECKO' is needed, as it's just the contract address and nothing specific to coingecko I think `KWENTA_ADDRESS` makes more sense.  Not sure the background behind the ETH coingecko address above.

sections/earn/StakeGrid.tsx
```diff
@@ -18,14 +27,66 @@ const TimeRemainingData = () => {
...
+				: zeroBN;
+
+			setTokenPrices({
+				kwentaPrice: kwentaPrice,
+				wethPrice: wethPrice,
+			});
+		};
+
+		(async () => {
+			await tokens();
```
avclarke, line 65: I think it would be good to move this to redux state and actions. We could add a 
```
coinGecko: {
  weth: string,
  kwenta: string
}
```
to `PricesState` which would hold these. And then add an action `fetchEarnTokenPrices`  which will handle making the request, along with reducer for setting them. 

We'll probably want to add a `usePollAction('fetchStakingTokenPrices', fetchStakingTokenPrices, 30000)` in this file so we keep fetching the prices regularly

sections/earn/StakeGrid.tsx
```diff
@@ -18,14 +27,66 @@ const TimeRemainingData = () => {
...
+		})();
+	}, []);
+
+	const tvl = useMemo(
```
avclarke, line 69: I think we can move these memo'd values to `state/earn/selectors` as we'll aim to keep all data logic separate to UI components, which should help with performance as well.

avclarke (changes requested):
Looks good but left some comments where we should move some state to redux.

constants/currency.ts
LeifuChen: Yes, you are right. There is no need to add `COINGECKO` for KWENTA address.
1inch API and Coingecko API use two ETH addresses in the token list.

sections/earn/StakeGrid.tsx
LeifuChen: Thanks. Will make the changes.

sections/earn/StakeGrid.tsx
LeifuChen: Thanks. Will make the changes.

sections/earn/StakeGrid.tsx
LeifuChen: Thanks for the guidance. I moved all the calculation logic of APR (including LP token price, LP token amount, etc.) into `state/earn`.

avclarke (approved):
Nice changes, looks good! Just a minor suggestion for plural naming on one function.

sdk/services/kwentaToken.ts
```diff
@@ -100,6 +106,21 @@ export default class KwentaTokenService {
 			totalSupply: wei(totalSupply),
 			lpTokenBalance: wei(lpTokenBalance),
 			allowance: wei(allowance),
+			wethAmount: wei(wethAmount),
+			kwentaAmount: wei(kwentaAmount),
+			lpTotalSupply: wei(lpTotalSupply),
+		};
+	}
+
+	public async getEarnTokenPrice() {
```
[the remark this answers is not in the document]
LeifuChen: Thanks. The function has been renamed.

avclarke (approved):

platschi (approved):

PR #1931, opened 2023-01-30
Margin Preview Fix

Tburm:
Fix the margin preview displayed for a potential order. There were some bugs in the current version that didn't take our leverage soft-cap into account, as well as disabled functionality for previews.

## Description
* Take leverage cap into account for available margin and margin usage
* Add back trade preview for available margin, margin usage, and buying power
* Fix bug in buying power preview when trades are on the opposite side of positions.

## Related Issues
- #1878 

image

### DIFF
--- a/sections/futures/MarketInfoBox/MarketInfoBox.tsx
+++ b/sections/futures/MarketInfoBox/MarketInfoBox.tsx
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 
 import InfoBox from 'components/InfoBox';
 import PreviewArrow from 'components/PreviewArrow';
-import { FuturesPotentialTradeDetails } from 'sdk/types/futures';
+import { PositionSide } from 'sdk/types/futures';
 import {
 	selectDelayedOrderFee,
 	selectMarketInfo,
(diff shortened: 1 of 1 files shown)


sections/futures/MarketInfoBox/MarketInfoBox.tsx
```diff
@@ -25,78 +25,88 @@ const MarketInfoBox: React.FC = () => {
...
 
-	const previewAvailableMargin = React.useMemo(() => {
-		const potentialAvailableMargin = getPotentialAvailableMargin(
-			potentialTrade,
-			marketInfo?.maxLeverage
-		);
+	// adjust accessible margin due to frontend soft cap on leverage
+	const availableMargin = useMemo(() => {
+		if (!position?.position || !marketInfo) return zeroBN;
+		return getAvailableMargin(position.position.notionalValue, totalMargin, marketInfo.maxLeverage);
```
avclarke, line 53: Should this also be using the selector `maxLeverage` with our local cap rather than the market leverage?

Tburm: No, because the `getAvailableMargin` function uses the actual static max leverage for the calculation. It's something like `total margin - (notional size / max leverage)`.

If we use the selector, it changes depending on the trade size, so this function can't be reused for both calculating "available margin" and "preview available margin".

Also - The reason why I'm using a function to calculate "available margin" here instead of using `accessibleMargin` is because the contract returns a value valid for 100x leverage. Using this function will account for our settings override. Same was true of "buying power" since it was calculated from "available margin"

avclarke: ok makes sense, I was mostly wondering if `marketInfo.maxLeverage` is 100x but I guess it's not.

avclarke (approved):

PR #1941, opened 2023-02-01
Remove futures context

avclarke:
## Description
Continuing the state refactor, we can now remove FuturesContext for good, moving all existing logic there to redux or the market page directly.


state/futures/selectors.ts
```diff
@@ -774,3 +774,18 @@ export const selectHasRemainingMargin = createSelector(
 			: posMargin.gt(0);
 	}
 );
+
+export const selectMaxUsdInputAmount = createSelector(
```
Tburm, line 778: Not really a suggestion for this PR, but this would be a great place to also add a check on OI caps to avoid users trying to submit orders over the cap and getting an error.

Tburm (commented):
I can't see how it was caused by this PR, but I opened an order which filled and the order hasn't been removed from the table. I'm stuck with an expired order even though my position has been modified.

Tburm:
It also doesn't look like the first load directly on the "Futures" page pulls my data. On first load I see the "Deposit margin" button and no position. If I navigate to dashboard and back to futures, my positions are populated.

avclarke:
> It also doesn't look like the first load directly on the "Futures" page pulls my data. On first load I see the "Deposit margin" button and no position. If I navigate to dashboard and back to futures, my positions are populated.

I missed that the new futures data hook was living inside context, must not have noticed because of persist. Good catch, will update.

platschi:
This error popped up again with this PR; after deposit margin tx went through and balance updated correctly, but the tx notifier got stuck at "being sent to network":

Screenshot 2023-02-02 at 12 16 13

platschi:
> It also doesn't look like the first load directly on the "Futures" page pulls my data. On first load I see the "Deposit margin" button and no position. If I navigate to dashboard and back to futures, my positions are populated.

Seems to be resolved now :+1:

avclarke:
> This error popped up again with this PR; after deposit margin tx went through and balance updated correctly, but the tx notifier got stuck at "being sent to network":
> 
> Screenshot 2023-02-02 at 12 16 13

@platschi I can't recreate this, does it happen consistently for you?

platschi (approved):

state/futures/selectors.ts
avclarke: Good call but I think we need some UX direction, ideally displaying to the user what the available OI is rather than a confusing max value. Will create a ticket.

PR #1945, opened 2023-02-01
Fix number input crash and disable Sentry logs in dev

### LINKED ISSUE #1943: Exchange: Can enter non number characters in number input
### Subject of the issue
Since changing the number input type to "text" to deal with localisation issues there is now an issue where users can type any character and cause an error in trying to convert invalid value to Wei.

https://kwenta.sentry.io/issues/3897297109/?project=4504363261362177&query=is%3Aunresolved&referrer=issue-stream

### Expected behaviour
Non number characters should be prevented from typing in the number input

### Actual behaviour
Any characters can be typed

### Steps to reproduce
When in local dev mode type any character in the input and you'll get an error pop up. (This error is silent in prod)

avclarke:
## Description
Fixes an issue where typing a letter into the number input on exchange page causes the app to crash

closes #1943


LeifuChen:
Great job! The crash issue has been fixed.

koredefashokun (dismissed):

platschi (approved):

PR #1947, opened 2023-02-01
Fix spot tables

Tburm:
Update subgraph endpoints to re-enable spot history and market tables. Additionally, update the "previous day rates" query to fetch _all_ synths in our price object instead of limiting to futures markets.


platschi:
Deployment immediately crashes upon loading the landing page.

sdk/types/prices.ts
```diff
@@ -16,6 +16,11 @@ export type Prices<T = Wei> = Record<string, Price<T>>;
 
 export type PricesMap<T = Wei> = Partial<Record<AssetKey, T>>;
 
+export type SynthRate = {
```
avclarke, line 19: I was thinking more specifically `SynthPrice` rather than `rate`, as that's the naming convention we shifted to in the sdk, so would be good to maintain consistency there.

state/prices/reducer.ts
```diff
@@ -1,18 +1,25 @@
...
 
+import { fetchPreviousDayRates } from './actions';
 import { PricesState } from './types';
 
-const initialState: PricesState = {
+export const PRICES_INITIAL_STATE: PricesState = {
 	onChainPrices: {},
 	offChainPrices: {},
 	connectionError: null,
+	previousDayRates: [],
```
avclarke, line 14: I think now we've moved this to prices service it would be good to change the name from rates to prices to stay consistent.

avclarke (commented):
Still only a nitpick but I think now we've moved from futures into the prices context it would be nice to rename any occurrence of `rates` to `prices`.

avclarke (approved):

PR #1948, opened 2023-02-01
Optimize queries to reduce lag and remove synth redeemer

avclarke:
## Motivation and Context
There is extreme lag between switching networks and wallet which originates from `fetchRedeemableBalances` due it relying on log filtering. 

This now removes that query and the ability redeem old synths completely, users can still perform that action through the legacy app.


LeifuChen:
@platschi Compared with the dev, switching between Ethereum and Optimism is very smooth now. The user can click the network label immediately to open the network switcher. 
However, switching between Polygon and Arbitrum is a bit laggy. From the console, the issue is more likely on the RPC. So I want to double-check if we enabled the Arbitrum and Polygon network on Blast API and Infura.

LeifuChen (approved):

platschi (approved):

PR #1968, opened 2023-02-04
Fix market price

Tburm:
Update the price selector for futures markets to use `marketInfo` since the exchange prices were deprecated


LeifuChen (approved):
Thanks for the quick fix.

PR #1970, opened 2023-02-04
Fix market infinite loader on mobile

avclarke:
Fixes an issue on mobile where market page won't load because it is waiting on cross margin query.


LeifuChen (approved):
Verified on metamask mobile. Thanks for the quick test.

PR #1989, opened 2023-02-09
Fix number formatter error

avclarke:
## Description
Fixes a number formatter issues which was highlighted from a crash when selecting market details tab on mobile.


LeifuChen (approved):
Work like a charm. Thanks.

PR #1991, opened 2023-02-09
Show index and market price in market details

avclarke:
## Motivation and Context
Seeing one price in the chart (index price) and a different price in the markets details bar (skew adjusted price) is confusing and can look like a bug of price data being out of sync. These changes add market price and index price to the market details bar so it is more clear where the difference comes from.

Screenshot 2023-02-09 at 17 09 29


Tburm (commented):
Changes look good, but based on our discussion can we kick out `24H Volume` and move the funding rate into that position? Funding rate is more important on the trading page, and is currently hidden behind a horizontal scroll.

platschi (approved):

PR #2010, opened 2023-02-13
Number decimal formatting updates and fix market trade history

avclarke:
- Ensures market detail and trade history values are rendered with correct decimal formatting
- Updates the suggested decimal function to increase number of digits for some ranges
- Fixes a bug where trade history on the markets page was not loading when switching markets


Tburm (approved):
Approving, but agree with Ralf's comment to potentially trim the decimals in the tradingview chart. It currently shows 4 decimals for all assets, including BTC.

Tburm (approved):

platschi (approved):

PR #2013, opened 2023-02-13
Add sorting to mobile markets table

Tburm:
Add default sort by volume on mobile `FuturesMarketsTable`. Also removes a tab that only has one option:

Old:
image

New:
image


platschi:
Pushed a commit to fix the lint warning. Two notes on the sorting:

1) Would it be useful to also be able to sort by market name?
2) The Sort icon is kind of in between the labels, which does not make it immediately clear if I am sorting by 24 Volume or 24H Change.

Tburm:
@platschi - Since we stack the labels in a single column we would need a new solution that complies with `react-table`. I considered disabling the actual sorting on the button for this reason.

For now we could disable the column sorting and find a design that works. The main thing I wanted to accomplish was default sorting

platschi (approved):

PR #2025, opened 2023-02-15
Reset the status of submit/approve after page refresh

LeifuChen:
## Description
We got a few users' reports that the button on the exchange page is disabled with the label `submitting` or `approving`.
They may face the issue that the button status isn't reset, so they can't use the exchange page anymore.

## Related issue
Similar issue #1985 

## Motivation and Context
Make the dapp usable for the user again.

## How Has This Been Tested?
Asked the user to test the deployment in this PR and it works.

image


avclarke (approved):
Looks all good for a quick fix

platschi (approved):

PR #2026, opened 2023-02-15
Fix SDK Types

Tburm:
Adds strict typing to the result of some multicall queries.


avclarke (approved):
Nice changes, the multicall library not supporting types is a real pain. There's a fork of it that does which it might be worth trying if it's well maintained.

PR #2031, opened 2023-02-17
Pnl Update

Tburm:
Update the pnl displays across the app.

## Description
* Add `fundingRateVelocity` to the `FuturesMarket` entity
* Point subgraph links at a prototype subgraph with new pnl calculations


platschi:
- PnL is before Fees? 

I wonder if it makes a difference for users to see PnL and Fees next to it, or just the final value of Profit or Loss of a trade, including Fees already.

- Would be nice to color code positive and negative values red/green

sdk/constants/futures.ts
```diff
@@ -2,10 +2,10 @@ import { gql } from 'graphql-request';
 
 import { FuturesMarketAsset, FuturesMarketConfig, FuturesMarketKey } from 'sdk/types/futures';
 
-export const FUTURES_ENDPOINT_OP_MAINNET = `https://subgraph.satsuma-prod.com/${process.env.NEXT_PUBLIC_SATSUMA_API_KEY}/kwenta/optimism-perps/api`;
+export const FUTURES_ENDPOINT_OP_MAINNET = `https://api.thegraph.com/subgraphs/name/tburm/optimism-perps`;
```
avclarke, line 5: Would be nice to just remove these duplicated constants while we're making changes here

avclarke (dismissed):

sdk/constants/futures.ts
Tburm: I went through and deleted all the stuff that isn't used from this file

Tburm: Or transitioned into the SDK to avoid importing from this file.

Tburm:
@platschi - Pnl on the trade history does not include fees, as fees are right next to it.

Realized pnl on the futures page includes fees and any accrued funding.

Realized pnl on the leaderboard all includes fees and funding.

Tburm:
@platschi  @avclarke just note that we can't merge this right away. I will need to deploy the subgraph and coordinate with the release.

platschi (approved):

PR #2049, opened 2023-02-28
Feat/issue#2046

pldespaigne:
- [x] Make submit trade button the color of the trade side
- [x] Make the submit button in the confirmation view the color of the trade side
- [x] Update confirmation submit button labels to include side, e.g. "Submit Short Order"
- [x] Use the colored trade side label in the trade confirmation view
- [x] Update the confirmation view layout

| | |
|-|-|
| image | image |

| | |
|-|-|
| image | image |

image
image
image

### DIFF
--- a/sections/futures/Trade/DelayedOrderConfirmationModal.tsx
+++ b/sections/futures/Trade/DelayedOrderConfirmationModal.tsx
@@ -1,3 +1,4 @@
+import { wei } from '@synthetixio/wei';
 import { FC, useCallback, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import styled from 'styled-components';
@@ -6,10 +7,13 @@ import HelpIcon from 'assets/svg/app/question-mark.svg';
 import BaseModal from 'components/BaseModal';
 import Button from 'components/Button';
 import Error from 'components/ErrorView';
+import { InfoBoxContainer, InfoBoxRow } from 'components/InfoBox';
 import { FlexDivCentered } from 'components/layout/flex';
 import { ButtonLoader } from 'components/Loader/Loader';
 import { DesktopOnlyView, MobileOrTabletView } from 'components/Media';
+import Spacer from 'components/Spacer';
 import Tooltip from 'components/Tooltip/Tooltip';
+import { DEFAULT_FIAT_DECIMALS } from 'constants/defaults';
 import { PositionSide } from 'sdk/types/futures';
 import { getDisplayAsset, OrderNameByType } from 'sdk/utils/futures';
 import { setOpenModal } from 'state/app/reducer';
@@ -36,6 +40,7 @@ import {
 	formatDollars,
 	formatPercent,
 	formatNumber,
+	truncateNumbers,
 } from 'utils/formatters/number';
 
 import BaseDrawer from '../MobileTrade/drawers/BaseDrawer';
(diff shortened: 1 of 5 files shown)


sdk/types/futures.ts
```diff
@@ -244,7 +244,7 @@ export type FuturesOrderTypeDisplay =
 	| 'Market'
 	| 'Liquidation'
 	| 'Delayed'
-	| 'Delayed Offchain';
```
avclarke, line 247: We shouldn't change this underlying type but just make sure we map the naming for user display

sections/futures/Trade/DelayedOrderConfirmationModal.tsx
```diff
@@ -291,4 +365,52 @@ const Disclaimer = styled.div`
...
+	&.long {
+		color: ${(props) => props.theme.colors.selectedTheme.green};
+	}
+
+	&.short {
+		color: ${(props) => props.theme.colors.selectedTheme.red};
+	}
+`;
+
+const Spacer = styled.div`
```
avclarke, line 403: We have a `Spacer` component you can use

sections/futures/Trade/ManagePosition.tsx
```diff
@@ -212,6 +213,23 @@ const PlaceOrderButton = styled(Button)`
 	height: 55px;
 	text-align: center;
 	white-space: normal;
+	color: ${(props) => props.theme.colors.selectedTheme.black};
+
+	&.long {
+		background: ${(props) => props.theme.colors.selectedTheme.green};
+	}
+	&.long:hover {
```
avclarke, line 221: Would be nice if we could add these styles to the base button so we don't have to repeat them in the various places.

translations/en.json
```diff
@@ -845,6 +848,10 @@
 					"insufficient-margin": "Insufficient free margin",
 					"error": "Error generating preview"
 				},
+				"fees": {
+					"tooltip": "Fees are displayed as maker / taker. Maker fees apply to orders that reduce the market skew. Taker fees apply to orders that increase the market skew.",
```
avclarke, line 852: I think we can just use the existing translations rather than duplicating.

sections/futures/Trade/DelayedOrderConfirmationModal.tsx
```diff
@@ -291,4 +365,52 @@ const Disclaimer = styled.div`
...
 `;
 
+const OrderSummaryLine = styled.div`
+	display: flex;
+	flex-direction: row;
+	justify-content: space-between;
+	gap: 12px;
+`;
+
+const OrderSummaryCell = styled.div`
```
avclarke, line 375: I think we can use `InfoBoxContainer` and `InfoBoxRow` for these instead of creating new styled components.

sections/futures/Trade/DelayedOrderConfirmationModal.tsx
```diff
@@ -155,6 +164,11 @@ const DelayedOrderConfirmationModal: FC = () => {
 		]
 	);
 
+	const truncateLeverage = useMemo(
+		() => truncateNumbers(wei(Number(leverageInput) ?? 0), DEFAULT_FIAT_DECIMALS),
```
avclarke, line 168: We'll probably want to use the calculated leverage in `potentialTradeDetails` here

avclarke (changes requested):
Looking good, just a few minor things.

platschi:
continued in #2062

PR #2055, opened 2023-03-01
Added options to nav header

LeifuChen:
## Description
Add `Options` to Kwenta navigation header

## Motivation and Context
To provider the user a more seamless experience by switching options and perps.

## How Has This Been Tested?
1. Open the page and click on each item of the sub-menu

image
image

### DIFF
--- a/constants/links.ts
+++ b/constants/links.ts
@@ -16,6 +16,12 @@ export const EXTERNAL_LINKS = {
 		OneInchLink: (from: CurrencyKey, to: CurrencyKey) => `https://1inch.exchange/#/${from}/${to}`,
 		OptimismTokenBridge: 'https://gateway.optimism.io',
 	},
+	Options: {
+		Portfolio: 'https://kwenta-options.on.fleek.co/#/portfolio',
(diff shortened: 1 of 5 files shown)


constants/links.ts
```diff
@@ -16,6 +16,12 @@ export const EXTERNAL_LINKS = {
 		OneInchLink: (from: CurrencyKey, to: CurrencyKey) => `https://1inch.exchange/#/${from}/${to}`,
 		OptimismTokenBridge: 'https://gateway.optimism.io',
 	},
+	Options: {
+		Portfolio: 'https://kwenta-options.on.fleek.co/#/portfolio',
```
platschi, line 20: For the release, we should change the domain names to `https://options.kwenta.eth.limo/`

platschi (changes requested):
Looks great, just need to make sure we update the URLs when we release.

constants/links.ts
LeifuChen: updated. thanks.

platschi (approved):

PR #2066, opened 2023-03-03
Fix leverage input zero input state and reduce sentry errors

avclarke:
- Ensures the leverage input shows the placeholder before a user starts typing
- Reduces some of the errors being reported by Sentry


platschi (changes requested):
Looks good, just the badge on the Futures Market table could need some spacing between the market label and the badge itself.

platschi (approved):

PR #2070, opened 2023-03-06
Revert background color of side selector

avclarke:
Screenshot 2023-03-06 at 15 37 22

Was using a new theme background before which was inconsistent to other buttons


platschi (approved):
great catch

Issue #2074, opened 2023-03-06
Mobile: Hide open position table if no open positions


platschi:
### Subject of the issue
On the markets page, Kwenta shows a "other positions" table. On desktop, if there is no other position, it shows the table simply stating "You have no other open positions", which is fine.

On mobile, however, it states "Trade perpetual futures now!", and takes up valuable space. We should therefore hide this table **on mobile** if there is no other position open. If there is another position (other than the one displaying right now), we should show the other position(s).

### Your environment
* Network used (Mainnet? L2?)
* Which browser and its version

### Screenshots
This is how it should look like if a users has a position (e.g. EUR), but visits another market:
![Screenshot 2023-03-06 at 17 46 30](https://user-images.githubusercontent.com/548702/223226921-ef2e300b-1e8d-4136-9195-08704bc4aa55.png)

This is how it currently looks if visiting the EUR-PERP market with the open position. Notice the empty table with the placeholder text above the Positions/Orders/Trades/Transfers tab.

![Screenshot 2023-03-06 at 17 47 37](https://user-images.githubusercontent.com/548702/223227083-8fa4b7a5-8462-42f5-b8d5-c1aefd188f4f.png)

### Deliverables
- On the second screenshot, the "Trade perpetual futures now!" table for open positions should be hidden if no other open position can be fetched.
- Note this only applies to mobile and should not affect desktop

### Bounty
40 $OP

PR #2079, opened 2023-03-07
Hide mobile empty positions table on markets page

0xwalde:
Hides the positions table entirely on the markets page for mobile screen sizes when there are no positions to show there, since all that get rendered are a table header and the "Trade perpetual futures now!" call to action link. The table should behave exactly the same on the dashboard page as it did before, and should look the same on the markets page if the user has other open positions as well.

## Related issue
https://github.com/Kwenta/kwenta/issues/2074

## How Has This Been Tested?
Tested on Optimism Goerli with Chrome 110.0.5481.117

### With no positions open at all:
kwenta-2074-1

### With a single position open, viewing the open position:
kwenta-2074-2

### With a single position open, viewing a different market than the one with the open position
kwenta-2074-3


LeifuChen (approved):
Great work. Looks good to me.

platschi (approved):

PR #2083, opened 2023-03-07
Implement Pyth Tradingview API

Tburm:
Implement the TradingView price API provided by pyth

## Description
* Remove the subgraph queries to fetch historical candle data
* Implement the Pyth TV API


Tburm:
@platschi Can you test this version? I confirmed it's sending requests above 1H to Satsuma. Requests 1H and below will be fetched from the Pyth TradingView API endpoint.

I still noticed some gaps in data for 30M and 1H when scrolling farther back, but not bad enough to avoid merging the change.

avclarke (approved):
Looks good!

PR #2096, opened 2023-03-10
Fix: Pnl percentage

Tburm:
This adjusts the unrealized pnl percentage denominator to include deposits to match the realized pnl percentage.


Tburm:
Discussed with Marko the new calculation:
- Removed pnl percentage for realized profits
- For unrealized, we base the percentage on `initial margin + net transfers`. This means the percentage will adjust when margin is deposited or withdrawn, but the percentage will always match your actual margin amount. Meaning, it would reach -100% when your margin is truly fully exhausted.

platschi (approved):

PR #2110, opened 2023-03-21
cc reward harvesting

avclarke:
View and interact with CC vesting contract from staking page.

- Access like `kwenta.eth.limo/dashboard/staking/?ccAddress=<VESTING_CONTRACT_ADDR>`
- Click claim to harvest rewards
- And use stake escrow with any amount set to stake all of your available escrowed Kwenta


platschi:
Works perfectly, would just like to have the two GitHub code scanning issues addressed before merging (if relevant).

PR #2112, opened 2023-03-21
Remove dropdown autofocus

### LINKED ISSUE #2075: Mobile: Market Dropdown Zooms unsolicited
### Subject of the issue
When using the new markets drop down on mobile, it zooms in when clicking/selecting a new market. However after selection, it does not "zoom out" again. We should avoid the zoom as the data by default is well readable.

Note this might be browser specific, pending testing

### Deliverables
- Disable zoom on mobile for markets dropdown

Tburm:
Using the `autoFocus` property is causing mobile to zoom on selection of a new market. Add a meta tag to adjust this behavior

## Description
- Add a meta tag to the document to avoid unwanted zooming on mobile/iPhones

## Issues
closes #2075

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -84,6 +84,10 @@ export default class MyDocument extends Document {
 						type="font/woff2"
 						crossOrigin="anonymous"
 					/>
+					<meta
+						name="viewport"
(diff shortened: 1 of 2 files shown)


platschi (approved):
So much better!

PR #2133, opened 2023-03-29
Smart margin: Margin transfers

Tburm:
Update the margin transfer table to include smart margin account transfers.

image


avclarke (approved):
Looks good!

PR #2134, opened 2023-03-30
UI Refresh - Futures positions table

### LINKED ISSUE #2038: UI Refresh: Implement new positions table UI
## Description

Implement the new positions tables UI from the futures design refresh

Screenshot 2023-03-06 at 17 30 15

Show a count in the tab with number of open positions.

koredefashokun:
## Description
This PR implements the new futures positions table according to the Figma designs.

## Related issue
closes #2038 

## Motivation and Context
This PR is a part of the UI refresh.

## How Has This Been Tested?
N/A

N/A


components/Currency/CurrencyPrice/CurrencyPrice.tsx
```diff
@@ -4,54 +4,54 @@ import styled, { css } from 'styled-components';
...
 	change?: number;
 	conversionRate?: WeiSource;
 	formatOptions?: FormatCurrencyOptions;
 	truncate?: boolean;
-	side?: 'positive' | 'negative';
+	side?: 'positive' | 'negative' | 'preview';
+	colored?: boolean;
 };
 
 export const CurrencyPrice: FC<CurrencyPriceProps> = memo(
```
avclarke, line 24: Maybe one for another PR but feel this is something that could be refactored or re-organised at least to fit with our new component lib.

components/Tab/Tab.tsx
```diff
@@ -10,22 +10,34 @@ export const TabList: FC = ({ children, ...props }) => (
...
-const TabPanelContainer = styled.div`
+const TabPanelContainer = styled.div<{ $fullHeight?: boolean }>`
 	outline: none;
+	${(props) =>
+		props.$fullHeight &&
+		css`
+			flex-grow: 1;
+			& > div,
+			& > div > div {
+				height: 100%;
```
avclarke, line 40: Would be better if we could put child styling on the child div components themselves. Also as this seems to be only used in one place there's maybe no need to have some dynamic styling. I think we could just focus on styling just the container and then how the children fit in can be done on the tab components themselves.

sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
```diff
@@ -60,6 +59,15 @@ const LegacyLink = () => {
...
+// - Market
+// - Side
+// - Size
+// - Average Entry
+// - Market Margin
+// - Leverage
+// - Liquidation Price
+// - PnL
+
 const FuturesPositionsTable: FC<FuturesPositionTableProps> = ({
```
avclarke, line 71: I think we said about just using the one table across both dashboard and market page, so can remove this now and configure the other one to not include the action buttons when used on dashboard.

sections/futures/UserInfo/PositionsTable.tsx
```diff
@@ -0,0 +1,430 @@
...
+								const formatOptions = cellProps.row.original.position.notionalValue.gte(1e6)
+									? { truncate: true }
+									: {};
+
+								return (
+									<div>
+										<Currency.Price
+											price={cellProps.row.original.position.size}
+											currencyKey={cellProps.row.original.market.asset}
+											showCurrencyKey
```
avclarke, line 187: I know it's in the design but I think we can remove the currency key here as it's clear from the first col what asset it is and just causes problem with using up valuable width like so:
Screenshot 2023-04-01 at 09 14 47

cc @asherism

avclarke (changes requested):
It's looking great, really feels like everything comes together with the new positions table. Just a few comments and suggestions.

components/Currency/CurrencyPrice/CurrencyPrice.tsx
koredefashokun: True. I was trying to do that in this PR, but this component needs a major rework (especially with the new `NumericValue` stuff we have going now).

sections/futures/UserInfo/PositionsTable.tsx
```diff
@@ -0,0 +1,430 @@
...
+				const market = futuresMarkets.find((market) => market.asset === position.asset);
+				const description = getSynthDescription(position.asset, t);
+				const thisPositionHistory = positionHistory.find((ph) => {
+					return ph.isOpen && ph.asset === position.asset;
+				});
+
+				return {
+					market: market!,
+					position: position.position!,
+					description,
```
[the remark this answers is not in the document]
koredefashokun: Yeah it is. Since the value is not undefined after the application of the filter, this means we don't have to add any nullish checks when using the position and market values in the table.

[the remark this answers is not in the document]
avclarke: was meaning just `description`?

sections/futures/UserInfo/PositionsTable.tsx
```diff
@@ -0,0 +1,430 @@
...
+								);
+							},
+							width: 100,
+						},
+						{
+							Header: <TableHeader>Funding</TableHeader>,
+							accessor: 'funding',
+							Cell: (cellProps: CellProps<typeof data[number]>) => {
+								return (
+									<div>
```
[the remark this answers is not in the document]
koredefashokun: Removed in the latest commit.

sections/futures/UserInfo/PositionsTable.tsx
```diff
@@ -0,0 +1,430 @@
...
+								);
+							},
+							width: 90,
+						},
+						{
+							Header: <TableHeader>TP/SL</TableHeader>,
+							accessor: 'tp-sl',
+							Cell: (cellProps: CellProps<typeof data[number]>) => {
+								return (
+									<div style={{ display: 'flex', alignItems: 'center' }}>
```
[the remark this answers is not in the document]
koredefashokun: Changed in the latest commit.

sections/futures/UserInfo/PositionsTable.tsx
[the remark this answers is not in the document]
koredefashokun: Ah. Didn't realize. In that case, you're right. I'll take it out.

sections/futures/UserInfo/PositionsTable.tsx
```diff
@@ -0,0 +1,430 @@
...
+						{
+							Header: (
+								<TableHeader>{t('dashboard.overview.futures-positions-table.market')}</TableHeader>
+							),
+							accessor: 'market',
+							Cell: (cellProps: CellProps<typeof data[number]>) => {
+								return (
+									<StyledText>
+										{cellProps.row.original.market.marketName}
+										{/* <MarketBadge
```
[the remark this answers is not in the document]
koredefashokun: Yeah. Removed.

sections/futures/UserInfo/UserInfo.tsx
```diff
@@ -159,7 +158,15 @@ const UserInfo: React.FC = memo(() => {
 	}, [position]);
 
 	return (
-		<div style={{ flex: 1 }}>
+		<div
+			style={{
```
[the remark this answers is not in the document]
koredefashokun: Yeah. For sure.

sections/futures/UserInfo/PositionsTable.tsx
koredefashokun: Yeah, I agree. It should be easy for users to deduce the currency key, based on the market.

sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
avclarke: I would say don't worry about the order for now. But let's do this in a different PR anyway.

components/Tab/Tab.tsx
koredefashokun: @avclarke I think I found a solution. However, for it to work, I'll have to make the mobile changes first. We should probably add this to the list of things to handle in a separate PR, and I'll handle it alongside the mobile changes.

sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
```diff
@@ -277,16 +230,46 @@ const FuturesPositionsTable: FC<FuturesPositionTableProps> = ({
...
+										</PnlContainer>
+									);
+								},
+								width: 125,
+							},
+							{
+								Header: <TableHeader>Funding</TableHeader>,
+								accessor: 'funding',
+								Cell: () => {
+									return <div></div>;
```
avclarke, line 261: Are these left empty because we'll replace this with the other table in a separate PR?

sections/futures/UserInfo/PositionsTable.tsx
```diff
@@ -0,0 +1,409 @@
...
+				const market = futuresMarkets.find((market) => market.asset === position.asset);
+				const thisPositionHistory = positionHistory.find((ph) => {
+					return ph.isOpen && ph.asset === position.asset;
+				});
+
+				return {
+					market: market!,
+					position: position.position!,
+					avgEntryPrice: thisPositionHistory?.avgEntryPrice,
+					stopLoss: accountType === 'cross_margin' ? (position as any).stopLoss : undefined,
```
avclarke, line 90: Would be good if we could avoid `as any` here just so if we change something this wouldn't go unnoticed

koredefashokun: The issue here is that TS does not pick up the stopLoss and takeProfit values from the cross margin positions for some reason. I'll just add those to the isolated positions type as well then.

sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
koredefashokun: Something like that. That said. I can just copy it over instead of leaving it empty

PR #2141, opened 2023-03-31
Fix delay in position appearing after order executes

### LINKED ISSUE #2136: Bug: Position and order display delays
### Description

When a delayed order executes there can be a delay between the position displaying, causing some users to think it failed and submit a duplicate order.

### Proposed solution

On receiving new pending orders data, check if an order has dropped (been executed) and then refetch positions.

avclarke:
- Fixes an issue where when an order is executed there is a delay between the position appearing, causing some traders to submit duplicate trades. 
- Updates position modification actions to use either `desiredFillPrice` or `priceImpactDelta` depending on network.

Closes #2136


Tburm:
@avclarke  I think something else is off about the contract ABIs in this deployment, and possibly in the `cm-v2` deployment. I'm getting failed calls to get positions, and feel like it may be some contract ABI changes.

Tburm (approved):

PR #2152, opened 2023-04-03
SIP-2004 & SIP-2005 Release

Tburm:
Update ABIs, addresses, and contract logic to support contract upgrades from Synthetix

## Description
* [x] Update `PerpsV2MarketData` ABI
* [x] Update `PerpsV2MarketSettings` ABI
* [x] Update position modify logic
* [x] Update `PerpsV2MarketSettings` address
* [x] Update `PerpsV2MarketData` address


state/futures/actions.ts
```diff
@@ -992,11 +994,17 @@ export const modifyIsolatedPosition = createAsyncThunk<
...
 		const marketInfo = selectMarketInfo(getState());
-		const priceImpact = selectPriceImpactOrDesiredFill(getState());
+		const priceImpact = selectIsolatedPriceImpact(getState());
+		const desiredFill = selectDesiredTradeFillPrice(getState());
 		if (!marketInfo) throw new Error('Market info not found');
+		if (!account) throw new Error('Account not connected');
 
 		try {
+			const isFlagged = await sdk.futures.getIsFlagged(account, marketInfo.market);
+			const priceImpactOrFill = !!isFlagged ? desiredFill : priceImpact;
```
avclarke, line 1006: Not sure this will work will it, as isFlagged only returns when the account is flagged for liquidation, so this would return false when not reverting? We should prob make a special temporary function, `isFlaggedReverts` or something and return true on revert.

Tburm: You're right - I thought I committed `isFlagged == null` since that will work here and that's actually the version I tested.

Tburm: The function sends `null` if the transaction reverts in the sdk, since I added a `.catch()`

avclarke (approved):
Looks good, would just maybe add some todo / comment to remove the temp logic so we don't forget

sdk/contracts/types/PerpsV2Market.ts
```diff
@@ -466,6 +526,14 @@
...
     data: BytesLike
   ): Result;
+  decodeFunctionResult(
+    functionFragment: "submitCloseDelayedOrderWithTracking",
+    data: BytesLike
+  ): Result;
+  decodeFunctionResult(
+    functionFragment: "submitCloseOffchainDelayedOrderWithTracking",
+    data: BytesLike
+  ): Result;
```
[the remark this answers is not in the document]
platschi: Is this one an issue?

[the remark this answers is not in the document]
Tburm: I don't think we need to worry about it, it's in a file that is auto-generated from the contract ABIs.

PR #2155, opened 2023-04-04
UI Refresh - Mobile

koredefashokun:
## Description
This PR implements the new mobile designs.

## Related issue
N/A

## Motivation and Context
This is a part of the general UI refresh.

## How Has This Been Tested?
N/A


sections/futures/mobile.ts
```diff
@@ -26,13 +26,12 @@ export const SectionSeparator = styled.div`
...
 	height: 325px;
-
-	${(props) =>
-		props.noPadding &&
-		css`
-			padding: 15px 0 0;
-		`}
 `;
+
+// ${(props) =>
```
avclarke, line 33: Can we remove this?

sections/futures/MobileTrade/UserTabs/UserTabs.tsx
```diff
@@ -48,16 +48,23 @@ const UserTabs: React.FC = () => {
...
 `;
 
 const TabButtonsContainer = styled.div`
 	display: grid;
 	grid-template-columns: repeat(4, 1fr);
-	grid-column-gap: 15px;
 	margin-bottom: 15px;
 	overflow: auto;
+
+	> button {
```
avclarke, line 60: Noticed this is a repeat of above, could we break it out to a re-usable component, or add it to the TabButton itself?

sections/futures/MobileTrade/drawers/TradePanelDrawer.tsx
```diff
@@ -0,0 +1,63 @@
...
+		</StyledModal>
+	);
+};
+
+const StyledModal = styled(FullScreenModal)`
+	top: 0;
+	background-color: transparent;
+	z-index: ${zIndex.DRAWER};
+
+	& > div {
```
avclarke, line 30: Are these styled just being applied to direct div child which is Background, so we could add them to that instead? Prefer to avoid the nested css styles where possible

avclarke (commented):
Looks good, just a couple questions

sections/futures/MobileTrade/drawers/TradePanelDrawer.tsx
koredefashokun: It's actually being added to DialogOverlay, which is indirectly the first child because of the structure of the BaseModal component. We can't target it directly from here, and extending the component itself might have too many negative externalities. We probably should make a more dumb/barebones base modal component that isn't too opinionated because of all the new styles we're getting.

sections/futures/MobileTrade/UserTabs/UserTabs.tsx
koredefashokun: Done in latest commit.

sections/futures/mobile.ts
koredefashokun: Removed.

sections/futures/MobileTrade/drawers/TradePanelDrawer.tsx
avclarke: ok got it, another one for the refactor list

components/Button/TabButton.tsx
```diff
@@ -19,6 +19,7 @@ export type TabButtonProps = {
 	nofill?: boolean;
 	isRounded?: boolean;
 	onClick?: () => any;
+	flat?: boolean;
```
avclarke, line 22: Annoying we still have this prop style around but let's deal with it in a different component focused PR

avclarke (commented):
Code looks great, just the bits mentioned in Discord need addressing.

platschi:
Looks good so far. A few issues

- [ ] On first visit (Landing page -> click "Trade now"), the "Manage" button next to "Available margin does **not** show. It appears only if I visit the Dashboard, and from there click on the "Trader perpetual futures now" link or from the markets page.
- [ ] Empty state for Position and Trades tab missing
- [ ] Empty state text in Orders and Transfers tab could use some padding

IMG_7051C17CADAD-1

- [ ] Descriptive text for Size and Leverage use different shades of gray, make sure they are streamlined. This is the issue on both mobile and desktop. See screenshot:

IMG_4B7EF65B55A3-1

platschi:
- [ ] And last some issue with numbers "overlapping" sometimes. This seems to happen randomly. See screenshot where it happens to both the last digit on Index and Market price, as well as the last two digits of Funding:

IMG_9FC0BC525664-1

platschi:
Will merge, remaining smaller issues added to the minor mobile fix list at #2181

Issue #2153, opened 2023-04-03
OP Rewards: Implement Rewards page & badge


platschi:
### Subject of the issue
Integrate a new rewards page where users can claim their Kwenta trading rewards, upcoming OP trading rewards as well as OP. trading rewards from Synthetix.

Integration can be done against the `cm-v2` branch as part of the global UI refresh and SM launch.

### Design
See Figma: https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?node-id=8398-44078

### Rewards navigation component

- [x] Remove the sUSD wallet balance component in the navigation
- [x] Instead, add the Rewards component (see design)
- [x] On component, show:
    - [x] Kwenta icon + OP icon
    - [x] Total value of claimable rewards (OP+Kwenta) in sUSD
- [x] Add dropdown with three rows, showing claimable trading rewards Kwenta, claimable Kwenta $OP rewards as well as claimable $SNX rewards and data according to design
- [x] Network Switcher: Show only network icon and remove name of network in the visible navigation component (see figma design)
- [x] Move network switcher behind wallet address (see design)

### Rewards page

- [x] Add a new Rewards page to (left side) sidebar on Dashboard page
- [x] Remove link to Earn page from left sidebar
    - [x] Add a link button to the Earn page (`/dashboard/earn`) to the left of the 'Staking Docs' button on the Staking page
- [x] Add three "boxes" for each reward type according to design

### Todo

- [x] Hook up trading rewards scripts once ready
- [ ] Update Lore Ipsum texts (ping devDAO PM)
- tbd

PR #2179, opened 2023-04-09
Implement Rewards Page

LeifuChen:
## Related issue
#2153


sections/shared/Layout/AppLayout/Header/BalanceActions.tsx
```diff
@@ -1,250 +1,275 @@
...
+			isDisabled: claimDisabledSnxOp,
+		},
+	];
 
 	useEffect(() => {
-		setBalanceLabel(formatDollars(susdWalletBalance, { sign: '$' }));
-	}, [balanceLabel, susdWalletBalance]);
+		const tokenAddresses = [KWENTA_ADDRESS, OP_ADDRESS];
+		const initExchangeTokens = async () => {
+			const coinGeckoPrices = await sdk.exchange.batchGetCoingeckoPrices(tokenAddresses, true);
```
avclarke, line 126: async calls should be wrapped in a try catch, as this would result in an unhandled error. But also I think we can just add OP to `getEarnTokenPrices` and use the `fetchEarnTokenPrices` action. Ideally we shouldn't be making async requests inside components but keeping this sort of data fetching and storage in redux / sdk.

sections/shared/Layout/AppLayout/Header/BalanceActions.tsx
```diff
@@ -1,250 +1,275 @@
...
-		return null;
-	}
+			setRewardBalance(
+				toWei(kwentaPrice)
+					.mul(tradingRewards)
+					.add(toWei(opPrice).mul(kwentaOpRewards.add(snxOpRewards)))
+			);
+		};
+
+		(async () => {
```
avclarke, line 138: Let's try and always avoid anon functions but this relates to the comment above anyway so we can remove altogether.

pages/dashboard/rewards.tsx
```diff
@@ -0,0 +1,37 @@
...
+import RewardsTabs from 'sections/dashboard/Rewards/RewardsTabs';
+import { useAppDispatch, useAppSelector } from 'state/hooks';
+import { fetchClaimableRewardsAll, fetchStakingData } from 'state/staking/actions';
+
+type RewardsComponent = React.FC & { getLayout: (page: HTMLElement) => JSX.Element };
+
+const RewardsPage: RewardsComponent = () => {
+	const { t } = useTranslation();
+	const dispatch = useAppDispatch();
+	const walletAddress = useAppSelector(({ wallet }) => wallet.walletAddress);
```
avclarke, line 15: There is a selector `selectWalletAddress` you can use. It's better to try and avoid raw selectors in components so if we ever need to change something it can be done in a single place.

components/Pill.tsx
```diff
@@ -5,26 +5,56 @@ type PillProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
 	size?: 'small' | 'medium' | 'large';
 	color?: 'yellow' | 'gray' | 'red';
 	outline?: boolean;
+	fullWidth?: boolean;
+	isRounded?: boolean;
+	blackFont?: boolean;
```
avclarke, line 10: black font is a bit of weird name for a prop as we have light and dark theme

sdk/services/kwentaToken.ts
```diff
@@ -449,6 +449,89 @@ export default class KwentaTokenService {
...
+					? MultipleMerkleDistributorOp.isClaimed(reward[0], reward[4])
+					: MultipleMerkleDistributorPerpsV2.isClaimed(reward[0], reward[4])
+			)
+		);
+
+		const { totalRewards, claimableRewards } = rewards.reduce(
+			(acc, next, i) => {
+				if (!claimed[i]) {
+					acc.claimableRewards.push(next);
+					acc.totalRewards = acc.totalRewards.add(wei(formatEther(next[2])));
```
avclarke, line 524: I think wei already knows how to handle BigNumber as an initialiser so should be no need for formatEther

avclarke (commented):
Nice work, just wanted to give this a quick retroactive review as was merged before any code review. The main thing that should be addressed is around fetching token balances with in the component.

components/Pill.tsx
LeifuChen: Thanks. `black` means font weight here. It's not intuitive.  Do you have any other suggestion?
image

sdk/services/kwentaToken.ts
LeifuChen: Thanks. `formatEther` has been removed.

components/Pill.tsx
LeifuChen: I will just name it as `weight` same as the `Body` component.

pages/dashboard/rewards.tsx
LeifuChen: That's true. Updated. Thanks for the reminder.

sections/shared/Layout/AppLayout/Header/BalanceActions.tsx
LeifuChen: Thatnks. That's a better solution. Will change this.

PR #2189, opened 2023-04-13
Fix background z-index

koredefashokun:
## Description
This PR adds UI fixes to the beta release of smart margin.

## Related issue
N/A

## Motivation and Context
Part of UI refresh

## How Has This Been Tested?
N/A

### DIFF
--- a/sections/shared/Layout/HomeLayout/Background.tsx
+++ b/sections/shared/Layout/HomeLayout/Background.tsx
@@ -27,6 +27,7 @@ const Container = styled.div`
 	position: fixed;
 	top: 0px;
 	left: 0px;
+	z-index: -100;
 
 	> .line {
 		position: absolute;


sections/futures/FeeInfoBox/TradePanelFeeInfo.tsx
```diff
@@ -25,7 +25,7 @@ import TradeTotalFeesRow from './TradeTotalFeesRow';
 
 export const TradePanelFeeInfo = memo(() => {
 	return (
-		<FeeInfoBoxContainer>
+		<FeeInfoBoxContainer $hasBorder={false}>
```
avclarke, line 28: Since we separated out the row components I've just been avoiding wrapping them in the InfoBox to get this effect, not sure if you think that's a better solution as it's sort of just voiding the container anyway with this style.

koredefashokun: True. I'll update this

PR #2207, opened 2023-04-17
fix: landing page

platschi:
## Description
- [x] Make sure clicking on Markets (boxes) opens the market on smart margin, currently they link to isolated margin (market/?asset=XXX)
- [x] Remove commit hash from landing page (at bottom, before footer)
- [x] Update Footer "Perpetual Futures" link to point to smart margin


LeifuChen (approved):
Nice catch. It works perfectly.

PR #2216, opened 2023-04-18
SLTP fixes and improvements 

avclarke:
- Adds validation to sl / tp input fields 
- Fixes a bug where you can't type 0. something in edit sl / tp modal
- Fixes a bug where sl / tp was not being added to positions in selector

### DIFF
--- a/sections/futures/EditPositionModal/EditStopLossAndTakeProfitModal.tsx
+++ b/sections/futures/EditPositionModal/EditStopLossAndTakeProfitModal.tsx
@@ -15,7 +15,7 @@ import { ConditionalOrderTypeEnum, PositionSide } from 'sdk/types/futures';
 import { setShowPositionModal } from 'state/app/reducer';
 import { selectTransaction } from 'state/app/selectors';
 import { calculateKeeperDeposit, updateStopLossAndTakeProfit } from 'state/futures/actions';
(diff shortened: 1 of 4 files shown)


sections/futures/EditPositionModal/EditStopLossAndTakeProfitModal.tsx
```diff
@@ -161,7 +192,7 @@ export default function EditStopLossAndTakeProfitModal() {
 			<StyledSpacer marginTop={6} />
 			<EditStopLossAndTakeProfitInput
 				type={'take-profit'}
-				invalid={false}
+				invalid={tpInvalid}
```
LeifuChen, line 195: Shall we also apply `tpInvalid` and `slInvalid` checks to enable/disable the `Set TP/SL` button?

LeifuChen (commented):
Good work. Just leave one comment to check if we need to update the `SET SL & TP` button.

sections/futures/EditPositionModal/EditStopLossAndTakeProfitModal.tsx
avclarke: Yes good idea

LeifuChen (approved):
Great.

PR #2251, opened 2023-04-20
Adjust market page layout to support smaller screens

avclarke:
### Description

Improves the market page display on smaller screen size by:

- reducing the positions table height for smaller screens
- allowing position row to shrink more
- reducing trade panel width on smaller screens


platschi:
Screenshot 2023-04-20 at 12 18 25

We should check that the Edit pill won't overlap with the Avg. Entry price on bigger values

PR #2276, opened 2023-04-21
chore: eth deposit check

### LINKED ISSUE #2270: Chore: Conditional order ETH deposit warnings
### Description

When placing a conditional order and requiring an ETH deposit we should make sure there is helpful feedback when the user doesn't have enough ETH to cover the deposit.

avclarke:
### Description

Display warnings around ETH deposit when insufficient wallet balance

### Related issue

closes #2270

### Screenshots (if appropriate):

Screenshot 2023-04-21 at 16 55 19
Screenshot 2023-04-21 at 16 39 42

### DIFF
--- a/sdk/services/futures.ts
+++ b/sdk/services/futures.ts
@@ -428,15 +428,17 @@ export default class FuturesService {
 		if (!SUSD) throw new Error(UNSUPPORTED_NETWORK);
 
 		// TODO: EthCall
-		const [freeMargin, keeperEthBal, allowance] = await Promise.all([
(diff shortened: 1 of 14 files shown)


platschi (changes requested):
Using a wallet with sufficient margin, ETH and 0.01 ETH deposited to the smart margin wallet, when I try to open a limit order, in the Confirm Order modal, the button just turns the wheel endlessly, and does not let me to Submit the order:

Screenshot 2023-04-21 at 18 04 54

LeifuChen (approved):
Thanks for the quick fix.

platschi (approved):

PR #2278, opened 2023-04-21
Bug: Add realized funding to position

Tburm:
Fix a bug where realized funding from the offchain position history was not applied to the position.


avclarke (commented):
Looks good, only comment would be does it make sense to include in the selectors where we update pnl?

Tburm:
@avclarke Good call - I just worked it into the selector instead and this is much simpler.

platschi (approved):

PR #2282, opened 2023-04-22
Fix: Issue where users end up with a dust position after closing

### LINKED ISSUE #2274: Bug: Closing 100% position from isolated leaves some dust
### Subject of the issue

When closing a full position on isolated margin it is not totalling the full amount and leaving some dust in the position. 

### Expected behaviour

Should close 100% of the position

avclarke:
## Description

parseFloat in the strip zeros util included some rounding which resulted in dust positions being left open after closing. Changed to use a regex instead.

## Related issue
closes #2274


LeifuChen (approved):
Thanks for the quick fix.

PR #2292, opened 2023-04-25
Fix order id and increase fetch size

avclarke:
Fixes an issue users were having with cancelling conditional orders.


LeifuChen (approved):
thanks for the quick fix.

PR #2299, opened 2023-04-27
Fix smart margin on leaderboard

Tburm:
Add smart margin positions to the leaderboard by splitting the position history query.

## Description
* Add a new position history query to query by `account` (in addition to the `abstractAccount` query)
* Update the leaderboard query to use the complete position history

### DIFF
--- a/sections/leaderboard/TraderHistory.tsx
+++ b/sections/leaderboard/TraderHistory.tsx
@@ -189,13 +189,7 @@ const TraderHistory: FC<TraderHistoryProps> = memo(
 										accessor: 'pnl',
 										Cell: (cellProps: CellProps<typeof data[number]>) => (
 											<PnlContainer>
-												<Currency.Price
-													currencyKey="sUSD"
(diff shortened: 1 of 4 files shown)


sections/leaderboard/TraderHistory.tsx
```diff
@@ -189,9 +189,12 @@ const TraderHistory: FC<TraderHistoryProps> = memo(
 										accessor: 'pnl',
 										Cell: (cellProps: CellProps<typeof data[number]>) => (
 											<PnlContainer>
-												<ColorCodedPrice
+												<Currency.Price
+													currencyKey="sUSD"
 													price={cellProps.row.original.pnl}
-													$value={cellProps.row.original.pnl}
+													sign="$"
+													conversionRate={1}
```
koredefashokun, line 196: I think the `currencyKey`, `sign` and `conversionRate` props can be omitted, since they are set to their default values. Also, you can omit the `={true}` after the `colored` prop.

platschi (approved):

PR #2307, opened 2023-05-03
Fixed the issue that missing snx op file blocks the rewards claim

LeifuChen:
## Description
The user can't claim the trading rewards when missing one distribution file. 
The PR fixed the above issue. So the user can claim the other rewards if one of the distribution file is missing.

## How Has This Been Tested?
When snx op file is missing, the user can still claim the rewards.

image

### DIFF
--- a/sdk/services/kwentaToken.ts
+++ b/sdk/services/kwentaToken.ts
@@ -517,6 +517,7 @@ export default class KwentaTokenService {
 		);
 
 		const rewards = responses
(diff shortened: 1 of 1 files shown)


avclarke (approved):

platschi (approved):

platschi:
Approving so we show at least some data, however claim buttons on mobile devices are not working.

PR #2310, opened 2023-05-03
feat: add new markets

avclarke:
## Description

Updates the config for new markets listed in [sip-298](https://sips.synthetix.io/sips/sip-298/)


platschi (dismissed):
Looks good, only need to verify on UI once markets are live if they populate correctly with icons and market details.

Tburm (approved):

PR #2318, opened 2023-05-04
Fixed the market name on mobile trade history

LeifuChen:
## Description
The following issues have been fixed in this PR
1. The app crashes when the user clicks on the row of the trade history.
2. The timestamp of the trade is incorrect.

## How Has This Been Tested?
1. Open mobile trade and check the timestamp of the trade
2. Click on the row of the trade history


platschi:
What's up with the data when clicking on trades on mobile? Size, Price and Data are completely messed up (see your screenshot)

LeifuChen:
> What's up with the data when clicking on trades on mobile? Size, Price and Data are completely messed up (see your screenshot)

Good catch. Let me re-map the right field to the `TradeDrawer`.

platschi (approved):

Issue #2123, opened 2023-03-27
Deploy a Pyth price service


avclarke:
https://github.com/pyth-network/pyth-crosschain/tree/main/price_service/server

We should be running our own price service as to not rely on the Pyth public service.

Once deployed we should hook up the frontend to try first connecting to our own service and then falling back to the Pyth public feed.

platschi:
@avclarke done and available at `https://price.kwenta.io`. 

You can rename this ticket to 'Implement self-hosted pyth price server' :)

platschi:
Would be good to be able to set a default price server via `.env` variable; in this case it is easy to switch servers in case the Kwenta one is offline one day. For example SNX runs one as well, and so on.

Default fallback should be the public Pyth server in case the default price server is not responsive.

PR #2320, opened 2023-05-05
Added kwenta own-hosted pyth server

LeifuChen:
## Description
1. Use public Pyth server as a fallback provider
2. Add environment variable `NEXT_PUBLIC_DEFAULT_PRICE_SERVICE=KWENTA` to activate the kwenta's pyth server
3. Add retry count and maximum retries 
4. When we hit the maximum retries for one server, keep switching to the other one until any one of them works

## Related issue
#2123


sdk/services/prices.ts
```diff
@@ -54,6 +58,18 @@ export default class PricesService {
...
+				if (this.retryCount < this.maxRetries) {
+					this.retryCount++;
+					setTimeout(() => {
+						this.pyth = new EvmPriceServiceConnection(getPythNetworkUrl(params.networkId, true), {
+							logger: LOG_WS ? console : undefined,
+						});
+					}, Math.pow(2, this.retryCount) * 1000);
+				} else {
+					logError(new Error('Maximum retries exceeded'));
+					// TODO: Take action when maximum retries are exceeded, e.g. update the operational status
```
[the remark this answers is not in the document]
LeifuChen: Good idea. Updated.

avclarke (dismissed):
Changes look good

sdk/services/prices.ts
```diff
@@ -219,6 +201,58 @@ export default class PricesService {
...
+		this.sdk.context.events.emit('prices_connection_update', {
+			connected: true,
+		});
+
+		return this.pyth;
+	}
+
+	private setEventListeners() {
+		this.sdk.context.events.on('network_changed', (params) => {
+			if (this.pyth) {
```
avclarke, line 223: I think we can move this part to the start of `connectToPyth`

sdk/services/prices.ts
```diff
@@ -219,6 +201,58 @@ export default class PricesService {
...
+			logger: LOG_WS ? console : undefined,
+		});
+
+		this.pyth.onWsError = (error) => this.retryConnection(networkId, error);
+
+		this.sdk.context.events.emit('prices_connection_update', {
+			connected: true,
+		});
+
+		return this.pyth;
```
avclarke, line 218: As this is private and the return object never used I think we can remove this return line

sdk/utils/futures.ts
```diff
@@ -406,8 +408,9 @@ export const POTENTIAL_TRADE_STATUS_TO_MESSAGE: { [key: string]: string } = {
 	INSUFFICIENT_FREE_MARGIN: `You don't have enough sUSD for this trade`,
 };
 
-export const getPythNetworkUrl = (networkId: NetworkId) => {
-	return networkId === 420 ? 'https://xc-testnet.pyth.network' : 'https://xc-mainnet.pyth.network';
+export const getPythNetworkUrl = (networkId: NetworkId, currentServer: PriceServer = 'KWENTA') => {
```
avclarke, line 411: I think we could rename `currentServer` to just `server` as current sort of implies what's being used now

sdk/services/prices.ts
```diff
@@ -219,6 +201,58 @@ export default class PricesService {
...
+			}
+		});
+	}
+
+	private retryConnection(networkId: NetworkId, error: Error) {
+		if (this.pyth) {
+			this.pyth.closeWebSocket();
+		}
+
+		this.sdk.context.events.emit('prices_connection_update', {
```
avclarke, line 241: I think we should move this to inside `onWsError` directly

sdk/services/prices.ts
LeifuChen: Yes. That's correct. Updated

sdk/services/prices.ts
LeifuChen: Done.

sdk/services/prices.ts
```diff
@@ -28,40 +28,22 @@ import * as sdkErrors from '../common/errors';
 
 const DEBUG_WS = false;
 const LOG_WS = process.env.NODE_ENV !== 'production' && DEBUG_WS;
+const PRICE_SERVER = process.env.NEXT_PUBLIC_DEFAULT_PRICE_SERVICE === 'KWENTA' ? 'KWENTA' : 'PYTH';
```
avclarke, line 31: nitpick but would be good to call this DEFAULT_PRICE_SERVER just so it's distinct from the variable price server

sdk/services/prices.ts
```diff
@@ -219,6 +201,70 @@ export default class PricesService {
...
+		this.sdk.context.events.on('prices_connection_update', (params) => {
+			if (params.connected) {
+				this.subscribeToPythPriceUpdates();
+			} else {
+				this.retryConnection(this.sdk.context.networkId);
+			}
+		});
+
+		if (this.pyth) {
+			this.pyth.onWsError = (error: Error) => {
```
avclarke, line 245: I think we need to make this call where the pyth instance is initialized otherwise if we create a new instance this setEventListeners would not have been called again and we won't get this onError callback.

avclarke (changes requested):

sdk/services/prices.ts
LeifuChen: Done

sdk/services/prices.ts
LeifuChen: Yes. That's true. Moved this to the `connectToPyth`.

sdk/services/prices.ts
LeifuChen: Yes. That's true. Moved this to the connectToPyth.

LeifuChen: Thanks. Updated.

LeifuChen: Thanks for the reminder. I have reverted the changes and missed this one in the redo. Corrected.

avclarke (dismissed):
Changes look good

platschi (approved):

platschi (approved):

PR #2332, opened 2023-05-11
Add exact liquidation price

Tburm:


sdk/contracts/PerpsV2MarketInternalV2.ts
```diff
@@ -404,6 +418,48 @@ class FuturesMarketInternal {
...
+
+		const prices = Array.from(Array(1001).keys()).map((x) =>
+			wei(approxLiquidationPrice)
+				.mul(1 + (x - 500) / 1000)
+				.toBN()
+		);
+
+		// start with initial margin
+		const margins = await Promise.all(
+			prices.map(async (price) => await this._marginPlusProfitFunding(position, price))
```
avclarke, line 438: I think for this one we can remove the async from the function and then remove the awaits here

platschi (approved):

Issue #2331, opened 2023-05-11
Bug: Stake buttons get stuck after failed tx


avclarke:
### Subject of the issue
When staking, if the user cancels the tx or the tx fails the button does not reset to active to allow the user to try again, it gets stuck in a disabled state until the page reloads.

PR #2333, opened 2023-05-11
Reset the status when the transaction fails to execute

LeifuChen:
## Description
When staking, if the user cancels the transaction or if the transaction fails, the button does not revert to an active state. This means the user cannot try again, and the button remains in a disabled state until the page reloads. The PR will reset the status when the transaction fails to execute.

## Related issue
#2331 

## How Has This Been Tested?
1. Click on the `stake` button and the button will be disabled
2. Reject the txn when the metamask pop-up
3. The `stake` button is active again


platschi (approved):
Tested and works good

PR #2341, opened 2023-05-15
Fix min margin error

### LINKED ISSUE #2339: Bug: Preview error when less than $50 margin
### Subject of the issue
When setting less than $50 margin in the trade panel and not having a position open the preview generation will throw a division by zero error. This has started since updating the preview to correctly calculate liquidation.

### Solution

- [ ] Investigate where the division by zero happens in the preview and prevent this by throwing a min margin error 
- [ ] Add validation to the margin input to require min $50 margin before allowing a preview
- [ ] Ensure the submit button in the trade panel is disabled if a preview fails

avclarke:
## Description

Fixes an issue where attempting to open a position below min margin results in an error

closes #2339


platschi:
Good catch, thanks. Now successfully displays the error message informing user of the min 50$ requirement.

PR #2347, opened 2023-05-15
Update the position table

### LINKED ISSUE #2259: UX Bug: Close button is hidden on positions table
### Subject of the issue

On some smaller screens the close button at the end of the row gets hidden out of view, causing users to be confused about how to close a position.

A quick PR to alleviate it on some screens was opened [here](#2251 ) but requires some more UX thought.

LeifuChen:
## Description
Update the position table following the [new design](https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?type=design&node-id=9276-60611&t=pW7PNxmIGtBYqo5W-4)
image

## Related issue
closes #2259

image
image


components/Pill.tsx
```diff
@@ -45,6 +48,7 @@ const BasePill = styled.button<{
 	$fullWidth?: boolean;
 	$roundedCorner?: boolean;
 	$weight?: FontWeight;
+	$background?: PillColor;
```
avclarke, line 51: I think this background prop is confusing with the cross over of color prop. Really we should just have some color variants that derive style from `newTheme.pill[variant]`. Maybe we just need to add a new pill style there.

sections/futures/UserInfo/PositionsTable.tsx
```diff
@@ -267,12 +283,22 @@ const PositionsTable: FC<FuturesPositionTableProps> = () => {
 							accessor: 'pnl',
 							Cell: (cellProps: CellProps<any>) => {
 								return (
-									<PnlContainer>
-										<Currency.Price price={cellProps.row.original.position.pnl} colored />
-										<NumericValue value={cellProps.row.original.position.pnlPct} colored>
-											{formatPercent(cellProps.row.original.position.pnlPct)}
-										</NumericValue>
-									</PnlContainer>
+									<FlexDivRowCentered style={{ columnGap: '35px' }}>
```
avclarke, line 286: Can use the new columnGap prop

avclarke (commented):
Just a couple of minor things

sections/futures/UserInfo/PositionsTable.tsx
LeifuChen: Thanks for the remind. Updated.

components/Pill.tsx
LeifuChen: Good idea. Updated.

avclarke (approved):

platschi:
Also there's no hover state for the pencils, as we had before, so how do I know when I can click on it/they are clickable? 

Screenshot 2023-05-18 at 13 32 45

For the Size pencil, for bigger value positions it's also very close to the "Close" button, hard to actually click on it.

I am reluctant to merge this PR for now, would like some more design discussion on this.

LeifuChen:
> On smaller screen Close/Share buttons overlap with data.
> 
> Screenshot 2023-05-18 at 13 00 26

Good catch. I will fix this.

LeifuChen:
> Also there's no hover state for the pencils, as we had before, so how do I know when I can click on it/they are clickable?
> 
> Screenshot 2023-05-18 at 13 32 45
> 
> For the Size pencil, for bigger value positions it's also very close to the "Close" button, hard to actually click on it.
> 
> I am reluctant to merge this PR for now, would like some more design discussion on this.

1. Good catch. The original hover state is on the button background but we remove the button background and border in this PR. I will add the hover state to the pencil icon. 
2. I will try re-distribute the width of the columns to see if we can make it better.

avclarke:
Closing this as been replaced by #2413

PR #2350, opened 2023-05-16
Update the mobile view of market details

### LINKED ISSUE #2258: UI Refresh: Access full market details from mobile
### Description

In the latest mobile design the market details have been reduced to display only some data points. There should be a way to expand and show all the market info.

### Solution

Make the markets details bar expandable on mobile to show full info.

LeifuChen:
## Description
Show all the market info on the mobile following the [design](https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?type=design&node-id=9276-60858&t=pW7PNxmIGtBYqo5W-4)
image

## Related issue
closes  #2258 

image
image


sections/futures/MarketDetails/MarketDetails.tsx
```diff
@@ -190,24 +200,47 @@ const MarketSkew: React.FC<MarketDetailsProps> = memo(({ mobile }) => {
...
-			}
+			mobile={mobile}
+			dataKey={MarketDataKey[dataKey]}
+			value={mobile ? mobileValue : desktopValue}
 		/>
 	);
 });
 
+const OpenInterestLongDetail: React.FC<MarketDetailsProps> = (props) => (
+	<OpenInterestDetail isLong {...props} />
```
avclarke, line 238: Seems a bit overkill to have two separate components for a hardcoded prop, should we just pass the isLong prop at the point these are rendered instead, so we only use a OpenInterestDetail component?

sections/futures/MarketDetails/MarketDetails.tsx
```diff
@@ -35,22 +37,45 @@ type MarketDetailsProps = {
...
+	);
+
+	const MarketDetailsMobile = () => (
+		<MarketDetailsContainer mobile={mobile}>
+			<IndexPriceDetail mobile={mobile} />
+			<MarketSkew mobile={mobile} />
+			<HourlyFundingDetail mobile={mobile} />
+			<MarketPriceDetail mobile={mobile} />
+			<DailyChangeDetail mobile={mobile} />
+			<FlexDivRow style={{ columnGap: '25px' }}>
```
avclarke, line 67: Prefer to avoid inline styles, could we add this as a prop to FlexDivRow instead maybe?

avclarke (commented):
Looks good, just a couple of minor clean up things which could be good to include.

sections/futures/MarketDetails/MarketDetails.tsx
```diff
@@ -35,22 +37,45 @@ type MarketDetailsProps = {
...
 
+interface OpenInterestDetailProps extends MarketDetailsProps {
+	isLong: boolean;
+}
+
 const MarketDetails: React.FC<MarketDetailsProps> = ({ mobile }) => {
 	const dispatch = useAppDispatch();
 	const showHistory = useAppSelector(selectShowHistory);
+
+	const MarketDetailsDesktop = () => (
```
avclarke, line 48: Would be good to render these conditionally, as currently a waste rendering both but only using one currently

sections/futures/MarketDetails/MarketDetails.tsx
LeifuChen: Good idea. Thanks.

sections/futures/MarketDetails/MarketDetails.tsx
LeifuChen: Yes. That's a good idea. Updated.

avclarke (approved):
Changes look good!

platschi:
Screenshot 2023-05-17 at 11 36 33

What if we add a space between OI(L) and (S), as in `OI (L)` and `OI (S)` for better readability?

platschi (approved):

Issue #2144, opened 2023-04-02
UI Refresh: Implement funding chart


avclarke:
### Description

- Add a toggle in the market details bar to switch between price and funding chart
- Implement funding history chart similar to the portfolio chart
- Can construct the chart data from fundingRateUpdates query from the subgraph
- The chart should show the past weeks data with rates per hour

![image](https://github.com/Kwenta/kwenta/assets/4819006/e54cf039-e726-4c90-ad98-ebdfbbc29586)

Ref: [Figma Design](https://www.figma.com/file/YWU3PMsnqZaqr1gABgAFJK/UI-Refresh?type=design&node-id=8493-ref47034&t=ZnQij0qhBHA44JgA-4)

platschi:
See also #1161

koredefashokun:
@avclarke The Figma link above seems to not work correctly anymore. Would appreciate if you could send a new one or update it.

PR #2356, opened 2023-05-18
Funding rate chart

koredefashokun:
## Description
This PR adds a chart that displays the funding rate in the currently selected market over the past week.

## Related issue
#2144

## Motivation and Context
Post UI Refresh additions.

## How Has This Been Tested?
N/A


sdk/services/futures.ts
```diff
@@ -272,6 +272,32 @@ export default class FuturesService {
 		return positions;
 	}
 
+	public async getMarketFundingRates(marketAsset: FuturesMarketAsset) {
```
avclarke, line 275: Could we rename this to `getMarketFundingRatesHistory` just so it's more clear this is for historical data rather than current rates.

sdk/services/futures.ts
```diff
@@ -272,6 +272,32 @@ export default class FuturesService {
...
+		const periodLength = PERIOD_IN_SECONDS.TWO_WEEKS;
+		const minTimestamp = Math.floor(Date.now() / 1000) - periodLength;
+
+		const response = await request(
+			this.futuresGqlEndpoint,
+			gql`
+				query fundingRateUpdate($marketAsset: Bytes!, $minTimestamp: BigInt!) {
```
avclarke, line 282: Could be nice to move this into `queries` as the idea was to try and keep some of this more bloated query string code out of the service files.

sdk/services/futures.ts
```diff
@@ -272,6 +272,32 @@ export default class FuturesService {
...
+					fundingRatePeriods(
+						where: { asset: $marketAsset, timestamp_gt: $minTimestamp, period: "Hourly" }
```
avclarke, line 284: Same for period, let's pass as param instead of hard coding

sdk/services/futures.ts
```diff
@@ -272,6 +272,32 @@ export default class FuturesService {
 		return positions;
 	}
 
+	public async getMarketFundingRates(marketAsset: FuturesMarketAsset) {
+		const periodLength = PERIOD_IN_SECONDS.TWO_WEEKS;
+		const minTimestamp = Math.floor(Date.now() / 1000) - periodLength;
```
avclarke, line 277: Probably best we pass this in as a param just in anticipation of being flexible for others using this call once the sdk becomes public.

sections/futures/FundingChart.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const FundingChart = () => {
+	const theme = useTheme();
+	const dispatch = useAppDispatch();
+	const marketInfo = useAppSelector(selectMarketInfo);
+	const marketFundingRates = useAppSelector(({ futures }) => futures.marketFundingRates);
+
+	useEffect(() => {
+		if (marketInfo?.asset) {
+			dispatch(fetchFundingRates(marketInfo.asset));
```
avclarke, line 21: Might be good to use `usePollAction` here instead with an hourly interval

state/futures/reducer.ts
```diff
@@ -168,6 +171,7 @@ export const FUTURES_INITIAL_STATE: FuturesState = {
 		leverageInput: '0',
 	},
 	tradePanelDrawerOpen: false,
+	marketFundingRates: [],
```
avclarke, line 174: Would be nice to store funding rates by market so the data remains cached for speedy loading when switching around.

Also we'll probably want to bump the store version.

avclarke (changes requested):
Looking good, just a few suggestions.

sdk/services/futures.ts
koredefashokun: Good idea. Added in the latest commit.

sdk/services/futures.ts
koredefashokun: Moved in the latest commit.

sdk/services/futures.ts
koredefashokun: Done

sections/futures/FundingChart.tsx
koredefashokun: Added in the latest commit

state/futures/reducer.ts
koredefashokun: Pushed.

state/futures/reducer.ts
```diff
@@ -171,7 +171,7 @@ export const FUTURES_INITIAL_STATE: FuturesState = {
 		leverageInput: '0',
 	},
 	tradePanelDrawerOpen: false,
-	marketFundingRates: [],
+	marketFundingRates: {} as FuturesState['marketFundingRates'],
```
avclarke, line 174: You shouldn't need to do this as the whole object state is defined as `FuturesState`

state/futures/actions.ts
```diff
@@ -1943,12 +1943,14 @@ export const updateStopLossAndTakeProfit = createAsyncThunk<void, void, ThunkCon
 	}
 );
 
-export const fetchFundingRates = createAsyncThunk<any, FuturesMarketAsset, ThunkConfig>(
-	'futures/fetchFundingRates',
-	async (marketAsset, { extra: { sdk } }) => {
-		return sdk.futures.getMarketFundingRatesHistory(marketAsset);
-	}
-);
+export const fetchFundingRates = createAsyncThunk<
```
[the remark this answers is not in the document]
koredefashokun: Done.

state/futures/reducer.ts
avclarke: Oh yea I think I usually wrap the object type with Partial to deal with this `Partial<Record<...>>`

state/futures/types.ts
```diff
@@ -116,6 +116,7 @@ export type FuturesQueryStatuses = {
 	trades: QueryStatus;
 	selectedTraderPositionHistory: QueryStatus;
 	marginTransfers: QueryStatus;
+	marketFundingRates: QueryStatus;
```
avclarke, line 119: Same for query

state/futures/reducer.ts
koredefashokun: Switched to that.

state/futures/types.ts
```diff
@@ -242,6 +244,7 @@ export type FuturesState = {
 		>;
 	};
 	tradePanelDrawerOpen: boolean;
+	marketFundingRates: Record<FuturesMarketAsset, { timestamp: string; funding: string }[]>;
```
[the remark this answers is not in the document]
koredefashokun: Done in the latest commit.

state/futures/types.ts
koredefashokun: Done in the latest commit.

avclarke (approved):
Nice work, code looks good but will let @platschi give it a final test.

platschi (approved):

PR #2364, opened 2023-05-22
Truncate M values in Trading History

### LINKED ISSUE #2357: Bug: Truncate M values in Trading History
### Subject of the issue
See screenshot. We should truncate values over >1.000.000 to 1.00M in the trading history so that the trading history is readable.

Screenshot 2023-05-18 at 15 20 04

### Deliverables
- Investigate using `suggestedDecimals` for Amounts in the Trading History, verify with all other markets and if good, implement

LeifuChen:
## Description
We should truncate values over >1.000.000 to 1.00M in the trading history so that the trading history is readable.

## Related issue
closes #2357

## How Has This Been Tested?
1. Open the futures page and check the truncated value of trade history table

image

### DIFF
--- a/components/Currency/CurrencyPrice/CurrencyPrice.tsx
+++ b/components/Currency/CurrencyPrice/CurrencyPrice.tsx
@@ -16,7 +16,6 @@ type CurrencyPriceProps = {
 	change?: number;
 	conversionRate?: WeiSource;
 	formatOptions?: FormatCurrencyOptions;
-	truncate?: boolean;
 	colorType?: 'secondary' | 'positive' | 'negative' | 'preview';
 	colored?: boolean;
 };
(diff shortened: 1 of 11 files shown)


avclarke (dismissed):

avclarke (commented):
@LeifuChen the spec mentions 1M and above but the screenshots shows also thousands being truncated, do we need to make it only truncate 1M and above?

platschi:
> @LeifuChen the spec mentions 1M and above but the screenshots shows also thousands being truncated, do we need to make it only truncate 1M and above?

Good catch, I think we should only truncate 1M and upwards in the Trades History table, truncating thousands does not look very useful:

Screenshot 2023-05-23 at 09 09 33

platschi (approved):

PR #2365, opened 2023-05-22
Added locked margin to the trade balance

### LINKED ISSUE #2284: UX Bug: Display unavailable funds in closed markets
### Subject of the issue
Currently when a market is closed there is no way to see how much funds are available from that market, since we filtered out closed markets from idle margin.

### Proposed solution
We should find somewhere to display idle margin in closed markets.

LeifuChen:
## Description
Currently, when a market is closed, there is no way to see how much funds are available from that market since we filtered out closed markets from idle margins. In this PR, we add these missing margin (called `locked margin`) back to the balance section.

## Related issue
closes #2284 

### Desktop
image
### Mobile
image


sections/futures/Trade/TradeBalance.tsx
```diff
@@ -48,59 +54,116 @@ const TradeBalance: React.FC<TradeBalanceProps> = memo(({ isMobile = false }) =>
...
+							textTransform="none"
+							onClick={() => dispatch(setOpenModal('futures_isolated_transfer'))}
+						>
+							{t('header.balance.get-susd')}
+						</Button>
+					</DepositContainer>
+				) : (
+					<>
+						{isMobile ? (
+							<FlexDivCol style={{ rowGap: '2px' }}>
```
avclarke, line 83: There's a lot of duplicated code here just for col v row, I think it would be better to create styled components and pass `isMobile` to for determining the flex.

state/futures/selectors.ts
```diff
@@ -605,19 +605,21 @@ export const selectRemainingMarketMargin = createSelector(selectPosition, (posit
...
+				return market && market.isSuspended === isSuspended;
 			})
 			.filter((p) => !p.position?.size.abs().gt(0) && p.remainingMargin.gt(0))
 			.reduce((acc, p) => acc.add(p.remainingMargin), wei(0));
 		return idleInMarkets;
-	}
-);
+	});
+
+export const selectIdleMarginInMarkets = selectMarginInMarkets();
```
avclarke, line 620: Maybe we should rename this to `selectAvailableMarginInMarkets`

LeifuChen: Thanks. Updated.

sections/futures/Trade/TradeBalance.tsx
LeifuChen: Yes. Good idea. I will do that after merging #2361

avclarke (approved):

platschi (approved):

PR #2369, opened 2023-05-23
fix: icons on homepage

platschi:
## Description
Icons for new markets are missing on the landing page, this fixes it.

## Screenshots

Screenshot 2023-05-23 at 11 25 40

## Notes

Layout for PEPE is currently out of order due to 24H Change being messed up, this *should* fix itself once the value normalizes


avclarke (approved):
Looks good, thanks

PR #2378, opened 2023-05-23
v7.1.5

platschi:
## Description
PR to deploy `v7.1.5` which includes 8 new markets.

### DIFF
--- a/sections/futures/Trade/SmartMarginOnboardModal.tsx
+++ b/sections/futures/Trade/SmartMarginOnboardModal.tsx
@@ -1,0 +1,111 @@
+import dynamic from 'next/dynamic';
+import React, { memo } from 'react';
+import { useTranslation } from 'react-i18next';
+import styled from 'styled-components';
+
+import BaseModal from 'components/BaseModal';
+import Error from 'components/ErrorView';
+import { FlexDivRowCentered } from 'components/layout/flex';
+import Spacer from 'components/Spacer';
+import { selectSusdBalance } from 'state/balances/selectors';
+import { selectIsolatedTransferError } from 'state/futures/selectors';
+import { useAppSelector } from 'state/hooks';
+import { formatDollars } from 'utils/formatters/number';
+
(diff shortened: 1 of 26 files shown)


sections/futures/Trade/SmartMarginOnboardModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+};
+
+const SocketBridge = dynamic(() => import('../../../components/SocketBridge'), {
+	ssr: false,
+});
+
+const SmartMarginOnboardModal: React.FC<Props> = memo(({ onDismiss }) => {
+	const { t } = useTranslation();
+
+	const txError = useAppSelector(selectIsolatedTransferError);
```
avclarke, line 26: @LeifuChen this is a SmartMargin component but referencing isolatedTransfer error, is that expected?

utils/formatters/number.ts
```diff
@@ -106,11 +115,12 @@ export const formatNumber = (value: WeiSource, options?: FormatNumberOptions) =>
 	}
 
 	// specified truncation params overrides universal truncate
-	if (shouldTruncate && !truncation) {
-		if (weiValue.gt(1e6)) {
-			truncation = { divisor: 1e6, unit: 'M', decimals: 2 };
-		} else if (weiValue.gt(1e3)) {
-			truncation = { divisor: 1e3, unit: 'K', decimals: 0 };
+	if (truncateThreshold && !truncation) {
+		for (const threshold of thresholds) {
```
avclarke, line 119: I think we can use a cleaner way of doing this but can take a look in a follow up PR

sections/futures/Trade/SmartMarginOnboardModal.tsx
LeifuChen: No. It's an irrelevant selector as we don't have any deposit and withdraw logic in this component.

avclarke (approved):

PR #2380, opened 2023-05-23
Clean the code

LeifuChen:
## Description
This is the PR to resolve the following issues.
- Clean the unnecessary tx error in the `SmartMarginOnboard` component
- Add the TODO comment to the temporary solution of the prices and fix the non-serializable value issue
- Apply a cleaner solution to the truncation fix by removing the `break` statement

### DIFF
--- a/sdk/services/prices.ts
+++ b/sdk/services/prices.ts
@@ -188,11 +188,13 @@ export default class PricesService {
 		const prices = (response ? Object.values(response).flat() : []) as SynthPrice[];
 
 		//TODO: remove this once the rates are fixed
-		return prices.map((price) =>
+		const newPrices = prices.map((price) =>
 			price.synth === MarketAssetByKey[FuturesMarketKey.sPEPEPERP]
-				? { ...price, rate: wei(price.rate.toString()).div(1e10) }
+				? { ...price, rate: wei(price.rate.toString()).div(1e10).toString() }
 				: price
 		);
+
+		return newPrices;
 	}
 
 	public async getPythPriceUpdateData(marketKey: FuturesMarketKey) {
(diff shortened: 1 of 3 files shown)


sdk/services/prices.ts
```diff
@@ -187,11 +187,12 @@ export default class PricesService {
 		);
 		const prices = (response ? Object.values(response).flat() : []) as SynthPrice[];
 
+		//TODO: remove this once the rates are fixed
 		return prices.map((price) =>
 			price.synth === MarketAssetByKey[FuturesMarketKey.sPEPEPERP]
-				? { ...price, rate: wei(price.rate.toString()).div(1e10) }
+				? { ...price, rate: wei(price.rate.toString()).div(1e10).toString() }
```
avclarke, line 193: Good catch looks like this is an important one to get in a hotfix

LeifuChen: Yes. Agree. The error is also very annoying in the dev.

utils/formatters/number.ts
```diff
@@ -115,14 +115,12 @@ export const formatNumber = (value: WeiSource, options?: FormatNumberOptions) =>
...
-			if (weiValue.gt(threshold.value) && weiValue.gte(truncateThreshold)) {
-				truncation = threshold;
-				break;
-			}
-		}
-	}
+	truncation =
+		truncateThreshold && !truncation
+			? thresholds.find(
+					(threshold) => weiValue.gt(threshold.value) && weiValue.gte(truncateThreshold)
```
avclarke, line 121: Should this be `gte` in both places?

sdk/services/prices.ts
avclarke: I will also just try and fix on the subgraph now

utils/formatters/number.ts
LeifuChen: Yes. Correct. This should be `gte`. Otherwise, it will miss the threshold value. Will update this.

avclarke (approved):

platschi (approved):

PR #2385, opened 2023-05-24
Adjusted the height of the price chart after deselecting trading history

LeifuChen:
## Description
In the legacy futures, after deselecting the trading history box, the height of the price chart is not working as expected.

Reverted the following change in #2316 

image

Ref: https://github.com/Kwenta/kwenta/pull/2316/files#diff-75b78545b9e3fae40ffbea8e525c3b600e8a0c65dfbee2dd939021e19a3404caL83-L88

## Related issue
#2316

### DIFF
--- a/sections/futures/MarketInfo/MarketInfo.tsx
+++ b/sections/futures/MarketInfo/MarketInfo.tsx
@@ -16,7 +16,7 @@ const MarketInfo: React.FC = React.memo(() => {
 });
 
 const Container = styled.div`
-	height: 100%;
+	height: 100vh;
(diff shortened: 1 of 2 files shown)


platschi:
https://github.com/Kwenta/kwenta/assets/548702/38410dcd-9e96-4686-81c7-0285a2d3fb08

Minor issue for smaller screens (additional space below banner when switching trading history; but I think this we can neglect this given that legacy futures will be removed very soon. 

Approving PR for now.

platschi (approved):

Issue #2384, opened 2023-05-24
Banner: Site-wide visibility


platschi:
### Subject of the issue
We would like to show the announcement banner site-wide in order to announce the upcoming bedrock upgrade and inform regarding legacy deprecation

### Notes
- Make sure banner is visible on all pages
- Make sure pages do not break (see current isolated issue with trading history)
- Lower padding on banner / make more narrow to save space
- Make sure banner is also visible on mobile
- more tbd

PR #2387, opened 2023-05-25
Implement Banner with Site-wide Visibility

LeifuChen:
## Description
We want to show the announcement banner site-wide to announce the upcoming bedrock upgrade and inform regarding legacy deprecation.

- [X] Make sure the banner is visible on all pages
- [X] Make sure pages do not break (see current isolated issue with trading history)
- [X] Lower padding on the banner / make more narrow to save space
- [X] Make sure the banner is also visible on mobile
- [X] Implement the banner dismiss functionality. The user can dismiss the banner by clicking it, and the current timestamp will be stored locally. The banner will be visible again to the user if the current timestamp is three days or more after the stored value.

Extra: Also, fix the link of the kwenta option (from the portfolio page to trade page)

## Related issue
#2384 

image
image
image


constants/links.ts
```diff
@@ -17,7 +17,7 @@ export const EXTERNAL_LINKS = {
 		OptimismTokenBridge: 'https://gateway.optimism.io',
 	},
 	Options: {
-		Portfolio: 'https://options.kwenta.eth.limo/#/portfolio',
+		Trade: 'https://options.kwenta.eth.limo/#/trade',
```
platschi, line 20: Good catch

platschi (approved):

PR #2410, opened 2023-05-30
Fix the vertical scroll of conditional order table

LeifuChen:
## Description
- [X] Fix the vertical scroll of the conditional order table
- [X] Update the font family of the columns

## How Has This Been Tested?
1. Use wallet address: 0xC2ecD777d06FFDF8B3179286BEabF52B67E9d991
2. Scroll the conditional order table vertically

image

### DIFF
--- a/components/Text/CustomFontLabel.tsx
+++ b/components/Text/CustomFontLabel.tsx
@@ -1,0 +1,22 @@
+type Props = {
+	text: string;
+};
+
+export const CustomFontLabel: React.FC<Props> = ({ text }) => {
+	const chars = text.split('');
+
+	return (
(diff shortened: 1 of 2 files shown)


platschi:
Could you also fix change the font style in the columns to regular instead of monospaced for:

- Type
-  Size (except for numbers) 
- Reduce only (and capitalize)

LeifuChen:
> Could you also fix change the font style in the columns to regular instead of monospaced for:
> 
> * Type
> * Size (except for numbers)
> * Reduce only (and capitalize)

Updated.

platschi (approved):

PR #2418, opened 2023-06-02
Implement Smart Margin close method

avclarke:
## Description

- Switches from reducing position to close to using the smart margin function close function, which will be required to support close only mode
- Also adds back persistence


avclarke:
Opened another PR based off main

PR #2421, opened 2023-06-03
Show Close Only Prompt

LeifuChen:
## Description
1. Showing a blocking view of the trade panel for close-only mode with the following message during the optimism upgrade. 
`Positions are in close only mode while the Optimism network performs its Bedrock upgrade`
2. Set an env variable `NEXT_PUBLIC_CLOSE_ONLY=true` to enable the close-only mode.

image
image
image
image


platschi (approved):

sections/futures/MobileTrade/drawers/TradePanelDrawer.tsx
```diff
@@ -15,7 +16,11 @@ const TradePanelDrawer: FC<TradePanelDrawerProps> = ({ open, closeDrawer }) => {
 			<Background>
 				<Closer onClick={closeDrawer} />
 				<Foreground>
-					<TradeIsolatedMargin mobile />
+					{process.env.NEXT_PUBLIC_CLOSE_ONLY === 'true' ? (
+						<CloseOnlyPrompt mobile />
+					) : (
+						<TradeIsolatedMargin mobile />
```
avclarke, line 22: I think instead of having this duplicated in two places we could move it just render the close prompt inside the `TradePanel` instead.

LeifuChen: Good idea. Updated.

avclarke (approved):
Nice work, looks good!

PR #2426, opened 2023-06-05
Fix: Close full position and re-enable withdraw

avclarke:
## Description

- Fixes and issue where users are unable to close due to max market size error
- Re-enables withdraw


LeifuChen (approved):
Approved. Working as expected. Thanks for the quick fix.

A minor issue we can fix later: the layout has been changed because of one more component in the parent container. So we may need to change the CSS rule to re-position the prompt.
image

PR #2442, opened 2023-06-07
A better fix to the max price impact issue

LeifuChen:
## Description
1. Update the desired fill price selectors to use `preview.price` instead of `marketPrice` for trade, edit, close.
2. Increase the limit / stop buffers to 4 from 2

## Related issue
A better version of #2435

## Motivation and Context
After implementing #2435, the users will only be warned about price impact if above 4%, so most could go unnoticed on larger trades. This should be a better fix without too many changes.

### DIFF
--- a/state/futures/selectors.ts
+++ b/state/futures/selectors.ts
@@ -878,16 +878,70 @@ export const selectSlTpModalInputs = createSelector(
 export const selectCrossMarginOrderPrice = (state: RootState) =>
 	state.futures.crossMargin.orderPrice.price ?? '';
 
+export const selectTradePreview = createSelector(
+	selectFuturesType,
+	(state: RootState) => state.futures,
+	(type, futures) => {
+		const preview = futures[accountType(type)].previews.trade;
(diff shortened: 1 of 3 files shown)


avclarke (dismissed):
Changes look good but it's optional if you want to implement the other change to move desiredFill to preview as discussed. Otherwise we could look at that in a separate PR.

state/futures/selectors.ts
```diff
@@ -878,50 +878,109 @@ export const selectSlTpModalInputs = createSelector(
...
+				leverage: unserialized.margin.gt(0)
+					? unserialized.notionalValue.div(unserialized.margin).abs()
+					: wei(0),
+			};
+		}
+		return null;
 	}
 );
 
+export const selectDesiredTradeFillPrice = createSelector(
```
avclarke, line 969: I think we can delete all these desiredFillPrice selectors now and just read the value straight from preview selectors

platschi (approved):

PR #2445, opened 2023-06-08
4 decimal price for ARB in title

0xdapper:

### DIFF
--- a/utils/futures.ts
+++ b/utils/futures.ts
@@ -39,7 +39,7 @@ export const getSynthDescription = (synth: FuturesMarketAsset, t: TFunction) => 
 };
 
 export const isDecimalFour = (marketKeyOrAsset: string) =>
-	['sEUR', 'EUR', 'sDOGE', 'DOGE'].includes(marketKeyOrAsset);
+	['sEUR', 'EUR', 'sDOGE', 'DOGE', 'ARB'].includes(marketKeyOrAsset);
 
 export const orderPriceInvalidLabel = (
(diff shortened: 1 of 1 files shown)


0xdapper:
Maybe there should be a generic solution to show 4 decimals for any price < 10?

LeifuChen:
> Maybe there should be a generic solution to show 4 decimals for any price < 10?

Yes. We have a new solution to show the decimals based on the price. 
https://github.com/Kwenta/kwenta/blob/0f8af747026afacbab1b1eaf118ba90d97caee99/sdk/utils/number.ts#L208

LeifuChen:
I will close this PR. Thanks for your reminder. We need to replace the old hardcoded solution with the new one.

utils/futures.ts
```diff
@@ -39,7 +39,7 @@ export const getSynthDescription = (synth: FuturesMarketAsset, t: TFunction) =>
 };
 
 export const isDecimalFour = (marketKeyOrAsset: string) =>
-	['sEUR', 'EUR', 'sDOGE', 'DOGE'].includes(marketKeyOrAsset);
+	['sEUR', 'EUR', 'sDOGE', 'DOGE', 'ARB'].includes(marketKeyOrAsset);
```
LeifuChen, line 42: Can you revert this change, as it is unnecessary in this PR?

LeifuChen (approved):
Looks good to me.

platschi (approved):

PR #2464, opened 2023-06-12
Fix: Position table column width and data source

### LINKED ISSUE #2460: Bug: Can't access edit size button on larger positions
### Subject of the issue
When there is large value in the size field it is not possible to edit the size button.

### Solution
Make the size field expand enough to be able to click the button.

0fda286a86443184dd7803cb86db9c79

### LINKED ISSUE #2465: Bug: Market margin not updating after position change
### Subject of the issue
A user reported that the market margin doesn't update after modifying the position. 
image
image

### Your environment
* Network used (Mainnet? L2?)
* Which browser and its version

### Expected behaviour
The data on kwenta UI should be the same as the data on synthetix watcher.

### Actual behaviour
The data on kwenta UI also different from what synthetix watcher shows.

### Steps to reproduce
Tell us how to reproduce this issue.

LeifuChen:
## Description
1. Increase the column width in the position table to handle the large position size
2. Update data source for `market margin`: switch from `initial margin` to `remaining margin`

## Related issue
closes #2460 
closes #2465

image
image
image

### DIFF
--- a/sections/futures/UserInfo/PositionsTable.tsx
+++ b/sections/futures/UserInfo/PositionsTable.tsx
@@ -74,6 +74,7 @@ const PositionsTable: FC<FuturesPositionTableProps> = () => {
 				const markPrice = markPrices[market?.marketKey!] ?? ZERO_WEI;
 				return {
 					market: market!,
+					remainingMargin: position.remainingMargin,
 					position: position.position!,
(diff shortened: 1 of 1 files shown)


avclarke (commented):
I think we can just set this to `1fr` instead and should be ok

avclarke (approved):
Looks good

PR #2483, opened 2023-06-14
Split up multicall request for market params on Goerli

avclarke:
## Description

Fixes an issue where Multicall was failing to batch 40 requests since adding the new markets. Splitting up the request seemed to fix it.


platschi (approved):
Confirm, this works now on Goerli.

PR #2490, opened 2023-06-14
Add config for stETH market

avclarke:
## Description

In preparation for supporting the new stETH market.


platschi:
Do we have to set the decimal rule for the TradingView chart in `/components/TVChart/DataFeed.ts` ?

```ts
const getPriceScale = (asset: string | null) => {
        switch (asset) {
		case 'BTC':
		case 'BNB':
		case 'ETH':
		case 'stETH':
		case 'XAU':
			return 100;
```

platschi (approved):

PR #2515, opened 2023-06-21
Feat: Implementing Staking V2

### LINKED ISSUE #2393: Staking V2 Implementation Checklist
# Ticket Work in Progress

This issue will serve as an overview for tickets and todos with regard to Staking V2.

- [See here for reference document](https://github.com/tommyrharper/shared-notes/blob/main/stakingv2/overview/stakingv2-overview.md)
- [See here for the new smart contracts](https://github.com/Kwenta/token/pull/213)

## Todo
- [x] @tommyrharper: Set up [reference documentation](https://github.com/tommyrharper/shared-notes/blob/main/stakingv2/overview/stakingv2-overview.md)
- [x] @tommyrharper  set up v2 on testnet
- [x] @avclarke: Create first wireframes and user flows with regard to migration, new functionality etc.

## Testnet addresses

You only need to interact with the Proxies, not the implementations.

- RewardEscrowV2 Implementation: 0x0A34aee61770b3cE293Fb17CfC9d4a7F70945260
- RewardEscrowV2 Proxy: 0xf211F298C6985fF4cF6f9488e065292B818163F8
- StakingRewardsV2 Implementation: 0xE0aD43191312D6220DE64aFA54dbdD6982991A87
- StakingRewardsV2 Proxy: 0x3e5371D909Bf1996c95e9D179b0Bc91C26fb1279

## Staking

- Claimable Rewards: Display Claimable Rewards on V1+V2, claim should happen on V2

### Migration flow
- Users will still have tokens on V1, so for min. ~1y need to show both 
- Encourage users to unstake escrow on V1 and restake escrowed Kwenta on V2

// Note: Verify if this applies only to staked Kwenta, or also to escrowed Kwenta (which can't be moved to v2?)
// Note from Tom on the above ^ -> escrowed kwenta can only be moved to V2 after vesting, but users should still have the ability to stake their v1 escrowed kwenta so they can claim the associated rewards on V2.

- Encourage users to unstake liquid Kwenta on V1 and restake liquid Kwenta on V2

- Rewards in Escrow can't be migrated
- Rewards will be based on V1 and V2 funds
- Actual claiming will happen on V2

- V2 has view functions for V1

### Error Handling
- Errors have been changed and will be different, we should be aware of those

## New features
- Ability to offer permissions (delegated staking): Get rewards on behalf, stake escrow on behalf
- Staking Cooldown Period (unstake not possible for 2 weeks on total amount of staked Kwenta; even if you stake 1 Kwenta on top of 1000; cooldown is reset every time n tokens are staked)
- Escrowed entries are transferrable (NFTs): First step unstake escrowed Kwenta, then allow users to transfer escrowed Kwenta to other wallet
- Also possible to bulk transfer escrowed entry NFTs in a single transaction.
- We can fetch historical data now, so that can be utilized for a graph, rewards etc.

LeifuChen:
## Description
- [X] Implement the desktop layout
- [X] Implement the mobile layout
- [X] Add Staking V2 contracts
- [X] Implement the migration flow
- [X] Implement the staking flow
- [x] Buttons and status for contract interaction.
- [X] Data table for Escrow V2.
- [X] Displaying either Migrate to Staking V2 or Stake & Reward, depending on the user's wallet

## Related issue
closes #2393 

image
image
image


packages/app/src/pages/dashboard/migrate.tsx
```diff
@@ -0,0 +1,314 @@
...
+		} else {
+			dispatch(stakeKwentaV2(wei(kwentaBalance).toBN()))
+		}
+	}, [dispatch, kwentaBalance, kwentaStakingV2Approved])
+
+	const [, setCurrentTab] = useState(tabQuery ?? StakeTab.Staking)
+
+	useEffect(() => {
+		if (!!walletAddress) {
+			dispatch(fetchStakingData()).then(() => {
```
avclarke, line 98: I think a better approach would be to create an action `fetchStakeMigrateData` and have it perform these fetches under there. This way we can use `useFetchAction` hook here with walletAddress as a dep and we'll also be able to track state of the action and any errors by using `builder.addCase(fetchStakeMigrateData.rejected)` etc in the reducer.

packages/app/src/pages/dashboard/migrate.tsx
```diff
@@ -0,0 +1,314 @@
...
+			buttonLabel: t('dashboard.stake.tabs.migrate.unstake'),
+			onClick: handleUnstakeKwenta,
+			active: claimableBalance.lte(0) && stakedKwentaBalance.gt(0),
+		},
+		{
+			key: 'step-3',
+			copy: t('dashboard.stake.tabs.migrate.step-3-copy'),
+			label: t('dashboard.stake.tabs.migrate.staked'),
+			value: truncateNumbers(stakedKwentaBalanceV2, 2),
+			buttonLabel: kwentaStakingV2Approved ? t('dashboard.stake.tabs.migrate.stake') : 'Approve',
```
avclarke, line 131: We should be consistent with using i18n text, so be good to do this for 'approved' as well.

packages/app/src/pages/dashboard/migrate.tsx
```diff
@@ -0,0 +1,314 @@
...
+	]
+
+	return (
+		<>
+			<Head>
+				<title>{t('dashboard-stake.page-title')}</title>
+			</Head>
+			<StakingHeading>
+				<FlexDivCol rowGap="5px">
+					<StyledHeading variant="h4">Migrate to Staking V2</StyledHeading>
```
avclarke, line 215: I don't think migrate should be it's own page but it should just be a component of the staking page if the user has some actions to perform on migration.

packages/app/src/sections/dashboard/DashboardLayout.tsx
```diff
@@ -61,24 +60,24 @@ const DashboardLayout: FC<{ children?: ReactNode }> = ({ children }) => {
...
-				href: ROUTES.Dashboard.Rewards,
-			},
-			{
-				name: Tab.Governance,
-				label: t('dashboard.tabs.governance'),
-				active: activeTab === Tab.Governance,
-				href: EXTERNAL_LINKS.Docs.Governance,
-				external: true,
+				name: Tab.Migrate,
+				label: 'Migrate to V2',
```
avclarke, line 78: i18n

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,382 @@
...
+	const network = useAppSelector(selectNetwork)
+	const isL2 = useIsL2()
+	const epochData = useAppSelector(selectEpochData)
+	const selectedEpoch = useAppSelector(selectSelectedEpoch)
+	const walletAddress = useAppSelector(selectWallet)
+	const kwentaRewards = useAppSelector(selectKwentaRewards)
+	const opRewards = useAppSelector(selectOpRewards)
+	const snxOpRewards = useAppSelector(selectSnxOpRewards)
+
+	const futuresFeeQuery = useGetFuturesFeeForAccount(walletAddress!, start, end)
```
avclarke, line 63: Would be good if we refactored this to live in the sdk and fetch from redux the same way as other queries as we'd like to remove the use of react-query eventually.

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,382 @@
...
+			key: 'trading-rewards',
+			title: t('dashboard.rewards.trading-rewards.title'),
+			copy: t('dashboard.rewards.trading-rewards.copy'),
+			labels: [
+				{
+					label: t('dashboard.rewards.rewards'),
+					value: truncateNumbers(kwentaRewards, 4),
+				},
+				{
+					label: 'Fee Paid',
```
avclarke, line 122: i18n

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,382 @@
...
+					value: formatDollars(futuresFeePaid, { minDecimals: 2 }),
+				},
+				{
+					label: 'Fee Share',
```
avclarke, line 126: i18n

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,382 @@
...
+					value: formatPercent(ratio, { minDecimals: 2 }),
+				},
+			],
+			info: [
+				{
+					label: 'Period',
```
avclarke, line 132: i18n

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,382 @@
...
+					value: `Epoch ${period}`,
+				},
+				{
+					label: 'Total Pool Fees',
```
avclarke, line 136: i18n

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,382 @@
...
+			kwentaIcon: true,
+			linkIcon: true,
+		},
+		{
+			key: 'op-rewards',
+			title: t('dashboard.rewards.op-rewards.title'),
+			copy: t('dashboard.rewards.op-rewards.copy'),
+			labels: [
+				{
+					label: 'Rewards',
```
avclarke, line 153: i18n

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,382 @@
...
+			kwentaIcon: false,
+			linkIcon: false,
+		},
+		{
+			key: 'snx-rewards',
+			title: t('dashboard.rewards.snx-rewards.title'),
+			copy: t('dashboard.rewards.snx-rewards.copy'),
+			labels: [
+				{
+					label: 'Rewards',
```
avclarke, line 172: i18n

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,382 @@
...
+		kwentaRewards,
+	])
+
+	const ratio = useMemo(() => {
+		return futuresFeePaid.gt(0) && totalFuturesFeePaid.gt(0)
+			? futuresFeePaid.div(totalFuturesFeePaid)
+			: ZERO_WEI
+	}, [futuresFeePaid, totalFuturesFeePaid])
+
+	const REWARDS = [
```
avclarke, line 111: I think we can put this behind a useMemo instead as currently it will reformat this data on each render.

packages/app/src/sections/dashboard/Stake/EscrowTab.tsx
```diff
@@ -1,30 +1,172 @@
...
+
+	const timeLeft = useMemo(
+		() =>
+			stakedResetTime > new Date().getTime() / 1000
+				? formatTruncatedDuration(stakedResetTime - new Date().getTime() / 1000)
+				: NO_VALUE,
+		[stakedResetTime]
+	)
+
+	const DEFAULT_CARDS = [
```
avclarke, line 43: be good to change to useMemo and change name to regular casing

packages/app/src/sections/dashboard/Stake/StakingTab.tsx
```diff
@@ -4,75 +4,156 @@ import { useTranslation } from 'react-i18next'
...
-					fullWidth
-					variant="flat"
-					size="small"
-					disabled={!getReward || claimableBalance.eq(0)}
-					onClick={handleGetReward}
-				>
-					{t('dashboard.stake.tabs.staking.claim')}
-				</Button>
+				<StyledHeading variant="h4">Staking Rewards</StyledHeading>
+				<Body color="secondary">Stake your escrowed tokens to earn additional rewards.</Body>
```
avclarke, line 87: i18n

avclarke (changes requested):
Great work, it's coming together nicely! Just a few parts I think we can clean up a bit.

packages/app/src/pages/dashboard/migrate.tsx
```diff
@@ -0,0 +1,314 @@
...
+					<Body color="secondary">
+						Lorem ipsum dolor sit amet consectetur. Ut in nisl ut quam condimentum lacus.
+					</Body>
+				</FlexDivCol>
+				<StyledButton
```
avclarke, line 220: Can we have disabled / loading states on button

packages/app/src/pages/dashboard/migrate.tsx
```diff
@@ -0,0 +1,314 @@
...
+						<FlexDivRowCentered>
+							<FlexDivCol rowGap="5px">
+								<Body size="small" color="secondary">
+									{label}
+								</Body>
+								<Body size="large" color="preview">
+									{value}
+								</Body>
+							</FlexDivCol>
+							<Button
```
avclarke, line 246: Can we add disabled / loading

packages/app/src/components/Pill.tsx
```diff
@@ -53,7 +53,7 @@ const BasePill = styled.button<{
 			? '3.5px 8px'
 			: '10px 15px'};
 		height: ${props.$size === 'small' ? '20px' : props.$size === 'medium' ? '24px' : '36px'};
-		width: ${props.$fullWidth ? '100%' : 'auto'};
+		width: ${props.$fullWidth ? '100%' : props.$size === 'medium' ? '70px' : 'auto'};
```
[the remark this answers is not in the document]
LeifuChen: Good catch. I should revert this change as I used `Button` component instead. This is an unnecessary change now.

packages/app/src/components/StakeCard.tsx
```diff
@@ -112,25 +119,43 @@ const StakeCard: FC<StakeCardProps> = memo(
...
-						? t('dashboard.stake.tabs.stake-table.stake')
-						: t('dashboard.stake.tabs.stake-table.unstake')}
-				</Button>
+				<FlexDivCol>
+					<FlexDivCol>
+						<StakeInputHeader>
+							<Body color="secondary">{title}</Body>
+							<StyledFlexDivRowCentered>
+								<Body color="secondary">{t('dashboard.stake.tabs.stake-table.balance')}</Body>
+								<PointerBody color="primary" onClick={onMaxClick}>
```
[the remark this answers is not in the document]
LeifuChen: It's mainly for using the font Akkurat Mono LL to show the numerical value. I will replace it with `NumericValue` component

packages/app/src/components/Table/CustomizePagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+	pageCount: number
+	canNextPage: boolean
+	canPreviousPage: boolean
+	compact: boolean
+	setPage: (page: number) => void
+	previousPage: () => void
+	nextPage: () => void
+}
+
+const CustomizePagination: FC<PaginationProps> = React.memo(
```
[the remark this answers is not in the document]
LeifuChen: There is a big difference in the layout between `Pagination` (for the leaderboard and other tables) and `CustomizePagination` (for the escrow table). 
After implementing all the functionalities, I will check if it's possible to re-use sth. Thanks for the reminder.
image
image

packages/app/src/components/Table/Table.tsx
```diff
@@ -50,6 +50,17 @@ export function compareNumericString(rowA: Row<any>, rowB: Row<any>, id: string,
...
 
+function calculatePageSize(
+	showPagination: boolean,
+	showShortList: boolean | undefined,
+	pageSize: number | null
+): number {
+	if (showPagination) {
+		return pageSize ? pageSize : MAX_PAGE_ROWS
+	}
+	return showShortList ? pageSize ?? 5 : MAX_TOTAL_ROWS
```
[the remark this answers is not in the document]
LeifuChen: Nice catch. Added.

packages/app/src/pages/dashboard/migrate.tsx
LeifuChen: Yes, that's true. Updated.

packages/app/src/sections/dashboard/Stake/StakingTab.tsx
```diff
@@ -4,75 +4,156 @@ import { useTranslation } from 'react-i18next'
...
-	const apy = useAppSelector(selectAPY)
+	const claimableBalance = useAppSelector(selectClaimableBalanceV2)
+	const stakedKwentaBalance = useAppSelector(selectStakedKwentaBalanceV2)
+	const apy = useAppSelector(selectAPYV2)
 
 	const handleGetReward = useCallback(() => {
 		dispatch(getReward())
 	}, [dispatch])
 
+	const DEFAULT_CARDS = [
```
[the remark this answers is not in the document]
LeifuChen: Done. Thanks.

packages/app/src/sections/dashboard/Stake/EscrowTab.tsx
LeifuChen: Thanks. Updated.

packages/app/src/sections/dashboard/DashboardLayout.tsx
LeifuChen: Thanks for the reminder. This is a temporary label on the sidebar for preview purposes. I will remove it before the final release.

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Thanks. Updated

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Updated.

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Done. Thanks.

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Done

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Done. Thanks.

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Done.

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Done.

packages/app/src/sections/dashboard/Stake/StakingTab.tsx
LeifuChen: Done. Thanks.

packages/app/src/pages/dashboard/migrate.tsx
LeifuChen: Yes. This is only for preview purposes. Will update before the final release.

packages/app/src/components/Button/TabButton.tsx
```diff
@@ -21,6 +21,8 @@ export type TabButtonProps = {
 	isRounded?: boolean
 	onClick?: () => any
 	flat?: boolean
+	noOutline?: boolean
```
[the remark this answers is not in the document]
LeifuChen: Yes. That's a better solution. Updated.

packages/app/src/pages/dashboard/migrate.tsx
LeifuChen: Yes. That's true. The button state should be same as the `Step`.

packages/app/src/pages/dashboard/migrate.tsx
LeifuChen: This is a link to the docs so there is no need to disable it.
image

packages/app/src/components/StakeCard.tsx
```diff
@@ -112,25 +119,43 @@ const StakeCard: FC<StakeCardProps> = memo(
...
+					<Spacer height={25} />
+					<Button
+						fullWidth
+						variant="flat"
+						size="small"
+						disabled={isDisabled}
+						onClick={handleSubmit}
+					>
+						{!isApproved
+							? t('dashboard.stake.tabs.stake-table.approve')
```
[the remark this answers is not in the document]
LeifuChen: Yes. Good catch. `Approve` is unnecessary for `unstake`. Updated.

packages/app/src/pages/dashboard/migrate.tsx
LeifuChen: Good idea. Updated.

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Yes. Good idea. I will refactor this later.

packages/app/src/components/Table/CustomizePagination.tsx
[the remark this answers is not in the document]
avclarke: It seems it's just style and layout differences. If we could have one component with a variant of the two layouts that would be a better approach IMO. If not we should probably rename this something more specific, maybe `StakingPagination`

packages/app/src/queries/staking/utils.ts
```diff
@@ -59,8 +62,15 @@ export function getApy(totalStakedBalance: number, weekCounter: number) {
...
-	const endDate = formatShortDate(new Date(toJSTimestamp(epochEnd)))
-	const label = `Epoch ${index}: ${startDate} - ${endDate}`
+	const label = `Epoch ${index}`
 	return { period: index, start: epochStart, end: epochEnd, label }
 }
+
+export const useEstimatedReward = (fileName: string) => {
+	const network = useAppSelector(selectNetwork)
+	const walletAddress = useAppSelector(selectWallet)
+	const query = useGetFile(
```
avclarke, line 72: I think we can move this out of utils as it's own query hook. Also ideally we move away from these useQuery hooks and use the async functions in redux / store in state but can look at that as a refactor.

packages/app/src/queries/staking/utils.ts
```diff
@@ -59,8 +62,15 @@ export function getApy(totalStakedBalance: number, weekCounter: number) {
...
+		`trading-rewards-snapshots/${network === 420 ? 'goerli-' : ''}${fileName}`
+	)
+	return BigNumber.from(query?.data?.claims[walletAddress!]?.amount ?? 0)
```
avclarke, line 75: Might be better to use Wei to remain consistent with the rest of the app.

packages/app/src/sections/dashboard/RewardsTab.tsx
```diff
@@ -0,0 +1,352 @@
...
+import media from 'styles/media'
+
+type EpochValue = {
+	period: number
+	start: number
+	end: number
+	label: string
+}
+
+const RewardsTabs: FC<TradingRewardProps> = ({ period = 0 }) => {
```
avclarke, line 38: The files called RewardsTab but here it's named RewardsTabs and can see there's already a file called RewardsTabs. Also if this staking specific maybe it should be under dashboard/staking .

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
```diff
@@ -79,43 +90,96 @@ const EscrowTable = () => {
...
+								/>
+							),
 							Cell: (cellProps: CellProps<EscrowData>) => (
-								<input
+								<Checkbox
+									id={cellProps.row.index.toString()}
 									key={cellProps.row.index}
-									type="checkbox"
 									checked={checkedState[cellProps.row.index]}
 									onChange={handleOnChange(cellProps.row.index)}
```
avclarke, line 151: should this be `() => handleOnChange(cellProps.row.index)`?

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
```diff
@@ -79,43 +90,96 @@ const EscrowTable = () => {
...
 								<TableCell>{cellProps.row.original.date}</TableCell>
 							),
 							accessor: 'date',
 							width: 65,
 						},
+						{
+							Header: () => <TableHeader>{t('dashboard.stake.tabs.escrow.amount')}</TableHeader>,
+							Cell: (cellProps: CellProps<EscrowData>) => (
+								<FlexDivRowCentered columnGap="10px">
+									<TableCell>{truncateNumbers(cellProps.row.original.amount, 4)}</TableCell>
```
avclarke, line 172: Should we be using formatNumber instead of truncate so we get the proper formatting on thousands etc?

avclarke (changes requested):
Looking good, just a few suggestions. About half way through the PR will continue to review.

packages/app/src/sections/dashboard/RewardsTab.tsx
LeifuChen: Yes. Good catch. This is copied from the `RewardsTabs.tsx`. Need to clean this a bit.
image

packages/app/src/components/Media.tsx
```diff
@@ -15,6 +15,14 @@ export const DesktopOnlyView: FC<MediaProps> = memo(({ children }) => (
 	<MediaQuery minWidth={BREAKPOINTS.md}>{children}</MediaQuery>
 ))
 
+export const LargeScreenView: FC<MediaProps> = memo(({ children }) => (
```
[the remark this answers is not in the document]
LeifuChen: The new name is much better. Updated.

packages/app/src/components/StakeCard.tsx
```diff
@@ -112,25 +118,36 @@ const StakeCard: FC<StakeCardProps> = memo(
...
+						onClick={handleSubmit}
+					>
+						{activeTab === 0
+							? isApproved
+								? t('dashboard.stake.tabs.stake-table.stake')
+								: t('dashboard.stake.tabs.stake-table.approve')
+							: t('dashboard.stake.tabs.stake-table.unstake')}
+					</Button>
+					<ErrorView
+						message={'2 week unstaking cooldown lock'}
```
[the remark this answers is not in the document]
LeifuChen: Thanks for the reminder. Updated.

packages/app/src/constants/routes.ts
```diff
@@ -24,6 +24,7 @@ export const ROUTES = {
 		Stake: normalizeRoute('/dashboard', 'staking', 'tab'),
 		Rewards: normalizeRoute('/dashboard', 'rewards', 'tab'),
 		TradingRewards: formatUrl('/dashboard/staking', { tab: 'trading-rewards' }),
+		Migrate: normalizeRoute('/dashboard', 'migrate', 'tab'),
```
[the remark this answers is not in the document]
LeifuChen: Removed. Thanks.

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
```diff
@@ -1,24 +1,36 @@
...
 
 import VestConfirmationModal from './VestConfirmationModal'
 
 const EscrowTable = () => {
 	const { t } = useTranslation()
 	const dispatch = useAppDispatch()
-	const escrowData = useAppSelector(({ staking }) => staking.escrowData)
+	const escrowV1Data = useAppSelector(selectEscrowData)
+	const escrowV2Data = useAppSelector(selectEscrowV2Data)
+	const escrowData = useMemo(() => [...escrowV1Data, ...escrowV2Data], [escrowV1Data, escrowV2Data])
```
avclarke, line 32: Could be good to move to a selector `combinedEscrowData`

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
```diff
@@ -79,43 +91,98 @@ const EscrowTable = () => {
 	const openConfirmModal = useCallback(() => setConfirmModalOpen(true), [])
 	const closeConfirmModal = useCallback(() => setConfirmModalOpen(false), [])
 
+	const EscrowStatsContainer = () => (
+		<StatsContainer columnGap="25px" justifyContent="flex-end">
+			<LabelContainers columnGap="25px" justifyContent="center">
+				<LabelContainer alignItems="flex-end">
+					<Body color="secondary">{t('dashboard.stake.tabs.escrow.total')}</Body>
+					<Body color="primary">
+						{truncateNumbers(totalVestable, 4)} {t('dashboard.stake.tabs.stake-table.kwenta-token')}
```
avclarke, line 100: Can we use formatNumber in place of truncate in all places

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
```diff
@@ -79,43 +91,98 @@ const EscrowTable = () => {
...
+					</Body>
+				</LabelContainer>
+				<LabelContainer alignItems="flex-end">
+					<Body color="secondary">{t('dashboard.stake.tabs.escrow.fee')}</Body>
+					<Body color="primary">
+						{truncateNumbers(totalFee, 4)} {t('dashboard.stake.tabs.stake-table.kwenta-token')}
```
avclarke, line 106: we don't need a translation for this as it's just `KWENTA` in all languages.

packages/app/src/sections/dashboard/Stake/StakingPortfolio.tsx
```diff
@@ -1,145 +1,102 @@
...
+	Redeem = 'redeem',
 }
 
 type StakingPortfolioProps = {
-	setCurrentTab(tab: StakeTab): void
+	cards: StakingCards[]
 }
 
-const StakingPortfolio: FC<StakingPortfolioProps> = ({ setCurrentTab }) => {
+const StakingPortfolio: FC<StakingPortfolioProps> = ({ cards }) => {
```
avclarke, line 22: Will probably be worth wrapping this component in memo since cards prop will be memoized

packages/app/src/state/futures/reducer.ts
```diff
@@ -738,6 +744,36 @@ const futuresSlice = createSlice({
...
+		builder.addCase(fetchFuturesFees.rejected, (futuresState) => {
+			futuresState.queryStatuses.futuresFees = {
+				status: FetchStatus.Error,
+				error: 'Failed to fetch fee data',
+			}
+		})
+
+		// Trading Fees by given epoch and trader
+		builder.addCase(fetchFuturesFeesForAccount.pending, (futuresState) => {
+			futuresState.queryStatuses.futuresFees = LOADING_STATUS
```
avclarke, line 765: This is the same query status as above `futuresFees`, should probably use a new query status item for a separate action otherwise they would conflict.

packages/app/src/state/staking/actions.ts
```diff
@@ -58,9 +58,43 @@ export const fetchStakingData = createAsyncThunk<
...
+	const {
+		rewardEscrowBalance,
+		stakedNonEscrowedBalance,
+		stakedEscrowedBalance,
+		claimableBalance,
+		totalStakedBalance,
+		stakedResetTime,
+		kwentaStakingV2Allowance,
+	} = await sdk.kwentaToken.getStakingV2Data()
+
```
avclarke, line 83: Would be good to wrap in try catch and notify on error, similar to the pattern used in futures.

packages/app/src/state/staking/actions.ts
```diff
@@ -108,6 +142,35 @@ export const fetchEscrowData = createAsyncThunk<
...
 })
 
+export const fetchEscrowV2Data = createAsyncThunk<
+	{ escrowData: EscrowData<string>[]; totalVestable: string },
+	void,
+	ThunkConfig
+>('staking/fetchEscrowV2Data', async (_, { extra: { sdk } }) => {
+	const { escrowData, totalVestable } = await sdk.kwentaToken.getEscrowV2Data()
+
+	return {
```
avclarke, line 152: same here for handling error and notifying

packages/app/src/state/staking/actions.ts
```diff
@@ -108,6 +142,35 @@ export const fetchEscrowData = createAsyncThunk<
...
+			fee: e.fee.toString(),
+		})),
+		totalVestable: totalVestable.toString(),
+	}
+})
+
+export const fetchStakeMigrateData = createAsyncThunk<void, void, ThunkConfig>(
+	'stakeMigrateData/fetch',
+	async (_, { dispatch }) => {
+		await dispatch(fetchStakingData())
```
avclarke, line 166: I don't think you need to await these unless there is some reason for them to be synchronous?

packages/app/src/state/staking/actions.ts
```diff
@@ -118,7 +181,27 @@ export const vestEscrowedRewards = createAsyncThunk<void, number[], ThunkConfig>
...
+				onTxFailed: () => {
+					dispatch({ type: 'staking/setVestEscrowedRewardsStatus', payload: FetchStatus.Error })
+				},
+			})
+		}
+	}
+)
+
+export const vestEscrowedRewardsV2 = createAsyncThunk<void, number[], ThunkConfig>(
+	'staking/vestEscrowedRewards',
```
avclarke, line 195:
```suggestion
	'staking/vestEscrowedRewardsV2',
```

packages/app/src/state/staking/actions.ts
```diff
@@ -128,16 +211,52 @@ export const vestEscrowedRewards = createAsyncThunk<void, number[], ThunkConfig>
...
+			},
+			onTxFailed: () => {
+				dispatch({ type: 'staking/setGetRewardStatus', payload: FetchStatus.Error })
+			},
+		})
+	}
+)
+
+export const claimStakingRewardsV2 = createAsyncThunk<void, void, ThunkConfig>(
+	'staking/getReward',
```
avclarke, line 233:
```suggestion
	'staking/claimStakingRewardsV2',
```

packages/app/src/state/staking/actions.ts
```diff
@@ -128,16 +211,52 @@ export const vestEscrowedRewards = createAsyncThunk<void, number[], ThunkConfig>
 	}
 )
 
-export const getReward = createAsyncThunk<void, void, ThunkConfig>(
+export const claimStakingRewards = createAsyncThunk<void, void, ThunkConfig>(
 	'staking/getReward',
```
avclarke, line 215:
```suggestion
	'staking/claimStakingRewards',
```

packages/app/src/state/staking/actions.ts
```diff
@@ -351,7 +466,81 @@ export const unstakeKwenta = createAsyncThunk<void, BigNumber, ThunkConfig>(
...
+			},
+			onTxFailed: () => {
+				dispatch({ type: 'staking/setUnstakeStatus', payload: FetchStatus.Error })
+			},
+		})
+	}
+)
+
+export const stakeEscrowV2 = createAsyncThunk<void, BigNumber, ThunkConfig>(
+	'staking/stakeEscrow',
```
avclarke, line 479:
```suggestion
	'staking/stakeEscrowV2',
```

packages/app/src/state/staking/actions.ts
```diff
@@ -351,7 +466,81 @@ export const unstakeKwenta = createAsyncThunk<void, BigNumber, ThunkConfig>(
...
+			},
+			onTxFailed: () => {
+				dispatch({ type: 'staking/setStakeEscrowedStatus', payload: FetchStatus.Error })
+			},
+		})
+	}
+)
+
+export const unstakeEscrowV2 = createAsyncThunk<void, BigNumber, ThunkConfig>(
+	'staking/unstakeEscrow',
```
avclarke, line 497:
```suggestion
	'staking/unstakeEscrowV2',
```

packages/app/src/state/staking/actions.ts
```diff
@@ -351,7 +466,81 @@ export const unstakeKwenta = createAsyncThunk<void, BigNumber, ThunkConfig>(
...
+				dispatch({ type: 'staking/setUnstakeEscrowedStatus', payload: FetchStatus.Error })
+			},
+		})
+	}
+)
+
+// TODO: Consider merging this with the (stake|unstake)Escrow actions.
+
+export const stakeKwentaV2 = createAsyncThunk<void, BigNumber, ThunkConfig>(
+	'staking/stakeKwenta',
```
avclarke, line 517:
```suggestion
	'staking/stakeKwentaV2',
```

packages/app/src/state/staking/actions.ts
```diff
@@ -351,7 +466,81 @@ export const unstakeKwenta = createAsyncThunk<void, BigNumber, ThunkConfig>(
...
+	async (amount, { dispatch, extra: { sdk } }) => {
+		const { hash } = await sdk.kwentaToken.stakeKwentaV2(amount)
+
+		monitorTransaction({
```
avclarke, line 521: Would be good if we could use a similar pattern to `monitorAndAwaitTransaction` so we are consistent with transaction handling throughout the app but something we can discuss and look at in a separate PR.

packages/app/src/state/staking/actions.ts
```diff
@@ -351,7 +466,81 @@ export const unstakeKwenta = createAsyncThunk<void, BigNumber, ThunkConfig>(
...
+			},
+			onTxFailed: () => {
+				dispatch({ type: 'staking/setStakeStatus', payload: FetchStatus.Error })
+			},
+		})
+	}
+)
+
+export const unstakeKwentaV2 = createAsyncThunk<void, BigNumber, ThunkConfig>(
+	'staking/unstakeKwenta',
```
avclarke, line 535:
```suggestion
	'staking/unstakeKwentaV2',
```

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
LeifuChen: Yes. It's a better solution. Updated.

packages/app/src/state/staking/reducer.ts
```diff
@@ -10,31 +10,43 @@ import {
...
+	fetchStakingV2Data,
+	fetchEscrowV2Data,
+	fetchStakeMigrateData,
 } from './actions'
 import { StakingState } from './types'
 
 export const STAKING_INITIAL_STATE: StakingState = {
 	kwentaBalance: '0',
 	escrowedKwentaBalance: '0',
+	escrowedKwentaBalanceV2: '0',
```
avclarke, line 28: as this root object is growing quite big it might be nice to break it up a bit, potentially with a `v1` and `v2` field. Something like:

```
VersionedStakeData = {
  claimableBalance: string
  totalStakedBalance: string
  stakedEscrowedKwentaBalance: string
  escrowedKwentaBalance: string
}

StakingState = {
  ...
  v1: VersionedStakeData
  v2: VersionedStakeData
}
```

packages/sdk/src/services/kwentaToken.ts
```diff
@@ -344,20 +458,46 @@ export default class KwentaTokenService {
...
 	public stakeKwenta(amount: string | BigNumber) {
 		return this.performStakeAction('stake', amount)
 	}
 
 	public unstakeKwenta(amount: string | BigNumber) {
-		return this.performStakeAction('unstake', amount)
+		return this.performStakeAction('unstake', amount, { escrow: false, version: 1 })
 	}
 
 	public stakeEscrowedKwenta(amount: string | BigNumber) {
```
avclarke, line 479: should this be renamed to `stakeEscrowedKwentaV2`?

packages/app/src/pages/dashboard/migrate.tsx
```diff
@@ -0,0 +1,305 @@
...
+			<StakingPortfolio cards={migrationInfo} />
+			<TableContainer>
+				<EscrowTable />
+			</TableContainer>
+		</>
+	)
+}
+
+const StyledButton = styled(Button)`
+	border-width: 0px;
```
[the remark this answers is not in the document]
LeifuChen: No. We don't have this button variant. The current `noOutline` border is using the style `'1px solid rgba(255, 255, 255, 0.12)'`, which is different from what we need.

packages/app/src/pages/dashboard/staking.tsx
```diff
@@ -45,12 +95,101 @@ const StakingPage: StakingComponent = () => {
...
-	return (
+	const timeLeft = useMemo(
+		() =>
+			stakedResetTime > new Date().getTime() / 1000
+				? formatTruncatedDuration(stakedResetTime - new Date().getTime() / 1000)
+				: NO_VALUE,
+		[stakedResetTime]
+	)
+
+	const stakingInfo: StakingCards[] = [
```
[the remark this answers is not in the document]
LeifuChen: Yes. Good catch. Updated.

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
LeifuChen: Updated.

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
LeifuChen: Yes. Fixed.

packages/app/src/state/staking/actions.ts
LeifuChen: Updated.

packages/sdk/src/services/kwentaToken.ts
LeifuChen: `stakeEscrowedKwentaV2` has already been defined in the following lines of the code. However, I went through the code and found that I had passed the wrong version into this function as I figured out why you ask this question. Thanks for the reminder.

packages/app/src/state/staking/actions.ts
LeifuChen: Yes. That's a good idea. I will follow the pattern used in futures transactions.

packages/app/src/queries/staking/utils.ts
LeifuChen: Yes. Updated the code in https://github.com/Kwenta/kwenta/pull/2515/commits/26bf8ce8d171485eff34fe1811b3674450b575cc# `packages/sdk/src/services/kwentaToken.ts`.

packages/app/src/queries/staking/utils.ts
LeifuChen: Good idea. Updated.

packages/app/src/sections/dashboard/Stake/StakingPortfolio.tsx
LeifuChen: Good idea. Updated.

packages/app/src/pages/dashboard/staking.tsx
```diff
@@ -1,21 +1,74 @@
...
+	const stakedResetTime = useAppSelector(selectStakedResetTime)
+	const [rewardBalance, setRewardBalance] = useState(ZERO_WEI)
+	const opRewards = useAppSelector(selectOpRewards)
+	const snxOpRewards = useAppSelector(selectSnxOpRewards)
+	const opPrice = useAppSelector(selectOpPrice)
+	const kwentaPrice = useAppSelector(selectKwentaPrice)
+
+	useFetchStakeMigrateData()
+
+	const isMigrationActive = useMemo(
```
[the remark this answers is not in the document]
LeifuChen: Yes. Updated

packages/app/src/state/futures/reducer.ts
LeifuChen: Good catch. I already defined a new query status but mistyped it here. Thanks for pointing it out.

packages/app/src/state/staking/actions.ts
LeifuChen: Fixed.

packages/app/src/state/staking/actions.ts
LeifuChen: Fixed.

packages/app/src/state/staking/actions.ts
LeifuChen: Fixed.

packages/app/src/state/staking/actions.ts
LeifuChen: Fixed.

packages/app/src/state/staking/actions.ts
LeifuChen: Fixed.

packages/app/src/state/staking/actions.ts
LeifuChen: Fixed.

packages/app/src/state/staking/reducer.ts
LeifuChen: Good idea. Will follow your advice.

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
LeifuChen: Yes. Fixed.

packages/app/src/sections/dashboard/Stake/EscrowTable.tsx
LeifuChen: Yes. Updated. Thanks.

packages/app/src/state/staking/actions.ts
LeifuChen: Yes. Fixed.

packages/app/src/state/staking/actions.ts
LeifuChen: Yes. Good idea. Updated.

packages/app/src/state/staking/actions.ts
LeifuChen: Thanks. Updated.

packages/app/src/components/Table/Table.tsx
```diff
@@ -72,6 +84,8 @@ type TableProps = {
 	rowStyle?: Record<string, any>
 	rounded?: boolean
 	noBottom?: boolean
+	customizePagination?: boolean
+	children?: React.ReactNode
```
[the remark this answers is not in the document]
LeifuChen: Yes. Good idea. I created two props here to control the pagination layout in different cases.
1.`paginationVariant?: 'default' | 'staking'`: 
  - If true, we show the page number on the right of the control panel and keep a space between the pagination and table, e.g., the escrow table on the desktop
  - If false, we show the page number in the middle of the control panel and append the pagination directly to the table, e.g., the escrow table on the mobile, leaderboard table

2. `paginationExtra?: React.ReactNode`
  - If passing the component, we always show the pagination regardless of the data length and show extra buttons to interact with the data in the table, e.g., switching v1/v2 data and vesting escrow entries in the escrow table
  - If null, we always show the pagination if the data length is over 5 and hide any extra buttons

| props   | Leaderboard | Escrow Table on Mobile   | Escrow Table on Desktop  |
|---------|-------------|--------------------------|--------------------------|
| variant |     default |                  default |                  staking |
| extra   |        null | <EscrowStatsContainer /> | <EscrowStatsContainer /> |

packages/app/src/components/Button/Button.tsx
```diff
@@ -167,6 +168,13 @@ const BaseButton = styled.button<BaseButtonProps>`
 				display: none;
 			}
 		`}
+	
+	${(props) =>
+		props.$variant === 'yellow' &&
+		props.outlineColor === 'yellow' &&
```
avclarke, line 174: Would really like to avoid adding new props to this component, I know it's a bit messy currently anyway but the idea is that the parent should only need to pass a variant and then we have the different variant styles defined inside here. Would probably be better if we just created a variant for this particular style `staking-button` or something?

packages/app/src/components/Table/Pagination.tsx
```diff
@@ -30,42 +35,67 @@ const Pagination: FC<PaginationProps> = React.memo(
...
+						</ArrowButton>
+						<ArrowButton onClick={toLastPage} disabled={!canNextPage}>
+							<RightEndArrowIcon />
+						</ArrowButton>
+					</ArrowButtonContainer>
+				</PaginationContainer>
+				{extra}
+			</>
+		) : (
+			<StakingPagination
```
avclarke, line 76: This isn't really what I was hoping for but going to chat with Yashar about the differences here.

packages/app/src/components/Table/Table.tsx
```diff
@@ -118,76 +139,94 @@ const Table = <T,>({
...
+							setPage={table.setPageIndex}
+							previousPage={table.previousPage}
+							nextPage={table.nextPage}
+							variant={paginationVariant}
+							extra={paginationExtra}
+						/>
+					) : undefined}
+				</ReactTable>
+			</TableContainer>
+			{paginationVariant === 'staking' && (
```
avclarke, line 215: We shouldn't really be including parent specific logic inside here. Would be good if we can have a solution to keep this component generic.

packages/app/src/state/staking/reducer.ts
```diff
@@ -168,8 +217,45 @@ const stakingSlice = createSlice({
...
+			state.unstakeEscrowedStatus = FetchStatus.Loading
+		})
+		builder.addCase(unstakeEscrowV2.rejected, (state) => {
+			state.unstakeEscrowedStatus = FetchStatus.Idle
+		})
+		builder.addCase(claimStakingRewardsV2.pending, (state) => {
+			state.getRewardStatus = FetchStatus.Loading
+		})
+		builder.addCase(claimStakingRewardsV2.rejected, (state) => {
+			state.getRewardStatus = FetchStatus.Loading
```
avclarke, line 254:
```suggestion
			state.getRewardStatus = FetchStatus.Error
```

packages/app/src/sections/dashboard/Stake/RewardsTab.tsx
```diff
@@ -0,0 +1,387 @@
...
+						</RewardsContainer>
+					</CardGrid>
+				))}
+				<ButtonContainer>
+					<Button
+						variant="yellow"
+						size="small"
+						textTransform="uppercase"
+						isRounded
+						disabled={claimDisabledAll}
```
avclarke, line 254: Would be best to disable also when the transaction is pending and ideally show a loading state.

avclarke (commented):
Changes look great, my only last concerns are if the buttons properly handle disabled / loading states when a tx is pending and the structure of the pagination and table with custom staking logic in.

packages/app/src/state/staking/reducer.ts
LeifuChen: Good catch. Fixed.

packages/app/src/components/Button/Button.tsx
LeifuChen: Thanks for the advice. Fixed.

packages/app/src/sections/dashboard/Stake/RewardsTab.tsx
LeifuChen: Yes. Fixed. That's good practice. I will check the other buttons' state later.

packages/app/src/components/Table/Table.tsx
LeifuChen: Thanks. Updated the code according to #2583

platschi (approved):

PR #2525, opened 2023-06-22
chore: pyth-based conditional orders

platschi:
## Description
Update the copy for conditional orders to inform users that either the enchain Pyth price or the Chainlink price is used for execution. Also includes a link to the relevant blog post with further information.

Screenshot_2023-06-22_at_17 18 06


LeifuChen (approved):
Work like a charm. Thanks.

We could also manage the external link in a text like below.
image

PR #2527, opened 2023-06-26
Fix the application error on input for increasing the short position

LeifuChen:
## Description
On the Edit Position Szie input field, the application crashes at the backslash/deletion of input.

## How Has This Been Tested?
1. Open the short position
2. Edit the size
3. Type any numbers
4. Delete all the inputs and no crash

### DIFF
--- a/packages/app/src/state/futures/selectors.ts
+++ b/packages/app/src/state/futures/selectors.ts
@@ -920,6 +920,7 @@ export const selectEditPositionPreview = createSelector(
 	selectEditPositionInputs,
 	(state: RootState) => state.futures,
 	(type, { nativeSizeDelta }, futures) => {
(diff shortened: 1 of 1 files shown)


insulineru (approved):

packages/app/src/state/futures/selectors.ts
```diff
@@ -920,6 +920,7 @@ export const selectEditPositionPreview = createSelector(
 	selectEditPositionInputs,
 	(state: RootState) => state.futures,
 	(type, { nativeSizeDelta }, futures) => {
+		if (isNaN(Number(nativeSizeDelta))) return null
```
avclarke, line 923: I'm not sure if this is the best place to fix, there shouldn't be a preview when there is no position size so this is more a workaround and introduces some different logic to the other preview selectors.

Issue #2337, opened 2023-05-12
Bug: TP/SL denomination for tiny values too high 


platschi:
### Subject of the issue
For tiny value Perps, such as SHIBA, values for TP/SL on the `PositionCard` are too big to make any sense. See:

![tiny_shitcoin](https://github.com/Kwenta/kwenta/assets/548702/afc5e18d-f248-4528-9563-8d235ff31721)

A community member suggested to use scientific notation for tiny denominated TP/SL values.

platschi:
Use `suggestedDecimals` for now

PR #2531, opened 2023-06-28
Minor fixes and cleanup

koredefashokun:
## Description
This PR makes small changes to resolve minor post UI refresh issues.

## Related issue
#2337 

## Motivation and Context
Post UI Refresh

## How Has This Been Tested?
N/A

### DIFF
--- a/packages/app/src/sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
+++ b/packages/app/src/sections/dashboard/FuturesPositionsTable/FuturesPositionsTable.tsx
@@ -234,14 +234,20 @@ const FuturesPositionsTable: FC<FuturesPositionTableProps> = ({
 													<Body>{NO_VALUE}</Body>
 												) : (
 													<div>
-														<Currency.Price price={cellProps.row.original.takeProfit} />
+														<Currency.Price
+															price={cellProps.row.original.takeProfit}
+															formatOptions={{ suggestDecimals: true }}
+														/>
 													</div>
 												)}
 												{cellProps.row.original.stopLoss === undefined ? (
 													<Body>{NO_VALUE}</Body>
 												) : (
 													<div>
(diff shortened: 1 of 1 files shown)


packages/app/src/components/TVChart/TVChart.tsx
```diff
@@ -191,7 +197,8 @@ export function TVChart({
 				{ text: '5D', resolution: '15', description: '5 Days' },
 				{ text: '30D', resolution: '1H', description: '30 Days' },
 				{ text: '3M', resolution: '1H', description: '3 Months' },
-			],
+			] as TimeFrameItem[],
+			...(chartData ? { saved_data: JSON.parse(chartData) } : {}),
```
avclarke, line 201: I think it'd be cleaner if we moved the json parsing to the loadChartData function. We could always return an object there and then on need for conditional here.

packages/app/src/components/TVChart/TVChart.tsx
```diff
@@ -300,6 +307,12 @@ export function TVChart({
 		})
 	}, [marketAsset])
 
+	useEffect(() => {
+		_widget.current?.subscribe('onAutoSaveNeeded', () => {
```
avclarke, line 311: Wondering if we need to store a ref to this and unsubscribe when we clear the widget?

avclarke (commented):
Nice work, this will be a great addition! Just a couple of minor comments.

packages/app/src/components/TVChart/TVChart.tsx
koredefashokun: Literally made the change locally but hadn't pushed it. Pushed now.

packages/app/src/components/TVChart/TVChart.tsx
koredefashokun: Good point. Pushed.

platschi (approved):

PR #2537, opened 2023-06-29
fix: check for empty data in TVChart

insulineru:
## Description
Fix issue, when chart loading with empty object (crash chart)

## How Has This Been Tested?
1. Open application in incognito mode (with disabled localStorage)
2. Check TV Chart. It's works.

### DIFF
--- a/packages/app/src/components/TVChart/TVChart.tsx
+++ b/packages/app/src/components/TVChart/TVChart.tsx
@@ -8,7 +8,6 @@ import {
 	TimeFrameItem,
 	widget,
 } from 'charting_library/charting_library'
-import { isEmpty } from 'lodash/fp'
 import { useRouter } from 'next/router'
 import { useRef, useContext, useEffect, useCallback, useMemo } from 'react'
 import { ThemeContext } from 'styled-components'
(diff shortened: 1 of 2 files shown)


packages/app/src/components/TVChart/TVChart.tsx
```diff
@@ -198,7 +199,7 @@ export function TVChart({
 				{ text: '30D', resolution: '1H', description: '30 Days' },
 				{ text: '3M', resolution: '1H', description: '3 Months' },
 			] as TimeFrameItem[],
-			saved_data: chartData,
+			saved_data: isEmpty(chartData) ? undefined : chartData,
```
avclarke, line 202: I think it could be better to update `loadChartState ` to return an appropriate type so we don't have deal with the conditional here.

avclarke (commented):
Nice work on the quick fix, just one suggestion.

packages/app/src/components/TVChart/TVChart.tsx
koredefashokun: My bad for causing this in the first place. Pushed the fix. Nice catch @insulineru and thanks @avclarke.

platschi (approved):

PR #2557, opened 2023-07-04
Fix: Transaction notifier not displaying

avclarke:
## Description

Fixes an error within the transaction notifier which was causing it to fail to render.


LeifuChen (approved):
Thanks for the quick fix.

PR #2598, opened 2023-07-13
[Obsolete] Fix: Staking escrowed V1 rewards in the migration step 3

LeifuChen:
# We decided to revert to Staking V1, see #2611 

## Description
Implemented the migration step 3 to stake the V1 staking rewards.
- Update the migration step 3
- Remove the `stakingMigrationCompleted`
- Only use `stakingMigrationRequired` to check if the user needs to do the migration

Fixed the broken link of `Learn More` button. Now it will redirect the user to the rewards guide blog.

## Motivation and Context
In the new UI, there is no way for the user to stake escrowed Kwenta in V1

## How Has This Been Tested?
1. User has unclaimable rewards
- Go to the migration page and follow all the migration steps
3. User has done V2 migration before
- Go to the migration page once again since we added a new check with non-zero escrowed Kwenta balance
- Stake the escrowed Kwenta balance and then go to the staking V2 page
3. New User
- Go to the staking V2 page 

image


LeifuChen:
We decided to revert to Staking V1, see #2611

PR #2609, opened 2023-07-15
Fix: Remove the blog link from banner

LeifuChen:
## Description
1. Remove the blog link from the banner
2. Shor normal mouse cursor if the message is unclickable
3. Update the app version to v7.4.3


Marko-roy (approved):
Thanks

PR #2617, opened 2023-07-17
Fix: Hide step 2 if the user has no escrowed entries

LeifuChen:
## Description
1. Hide the step 2 in the reverse migration if the user has no escrowed entries.
2. Update the banner link to [Staking V2 Reverse Migration](https://mirror.xyz/kwenta.eth/xFomD0VE0H7o2sQ9zGeLyYPwCmtp8tLMXNTGtWy2UOQ)

## How Has This Been Tested?
Case 1: The migrated user has escrowed entries will see two steps
截屏2023-07-18 00 11 12

 
Case 2: The migrated user has no escrowed entries will see one step
截屏2023-07-18 00 10 00


tommyrharper (commented):
Looks good to me!

tommyrharper (approved):

PR #2624, opened 2023-07-18
Fix: Socket re-rendering issue upon network switch

### LINKED ISSUE #2576: 🐛  Socket: Switch to non-OP network removes Socket integration modal
### Subject of the issue
Users trying to swap from e.g. ETH to OP cannot once the modal requires the user to switch to L1, as this hides the modal from the UI, rendering the Socket integration useless:

https://github.com/Kwenta/kwenta/assets/548702/b202f1e1-ced7-4aca-b402-0f1afc2c719b

### Deliverables
- Make sure that the Socket modal is available on all chains

LeifuChen:
## Description
When the user is connected to the non-op network, the trade panel will be hidden, so the socket widget will be re-rendered. The PR is fixing this issue. Switching the network won't re-render the socket widget when the `Bridge & Swap` modal is open.

## Related issue
Closes #2576 

## How Has This Been Tested?
Case: User is connected to the Optimism Mainnet without any sUSD
1. Click on `Get sUSD`
2. Set the `Source` to `Ethereum` and `Token` to `Ether`
3. Set the `Destination` to `Optimism` and `Token` to `sUSD`
4. Click the `Review Quote` and `Confirm Bridge`
5. Click the `Switch chain to Ethereum`
6. Click the `Bridge`

Switching the network won't re-render the socket widget when the `Bridge & Swap` modal is open.
截屏2023-07-18 21 55 52


avclarke (approved):
Nice fix

PR #2631, opened 2023-07-19
Take deposit/withdraw modal back to smart margin

LeifuChen:
## Description
Add the `deposit/withdraw modal` back to the smart margin mode.

## Motivation and Context
It's easy for users to bridge more sUSD and trade more in one place.

## How Has This Been Tested?
Case 1. User has less than 50 sUSD
- Show `Get sUSD` to bridge more sUSD

Case 2. User has more than 50 sUSD
- Show `Deposit/Withdraw` modal. The user can expand `Bridge & Swap` to bridge more sUSD.

截屏2023-07-19 18 32 20

### DIFF
--- a/packages/app/src/components/SocketBridge/SocketBridge.tsx
+++ b/packages/app/src/components/SocketBridge/SocketBridge.tsx
@@ -6,8 +6,7 @@ import ArrowIcon from 'assets/svg/app/arrow-down.svg'
 import Connector from 'containers/Connector'
 import { chain } from 'containers/Connector/config'
 import { fetchBalances } from 'state/balances/actions'
-import { selectFuturesType } from 'state/futures/selectors'
-import { useAppDispatch, useAppSelector } from 'state/hooks'
+import { useAppDispatch } from 'state/hooks'
 import {
 	customizeSocket,
 	socketDefaultChains,
@@ -19,7 +18,6 @@ const SocketBridge = () => {
 	const { activeChain, signer } = Connector.useContainer()
 	const dispatch = useAppDispatch()
 	const customize = customizeSocket(useTheme())
-	const accountType = useAppSelector(selectFuturesType)
 	const onBridgeSuccess = useCallback(() => {
 		dispatch(fetchBalances())
 	}, [dispatch])
@@ -42,11 +40,9 @@ const SocketBridge = () => {
 				enableSameChainSwaps={true}
 				onBridgeSuccess={onBridgeSuccess}
 			/>
-			{accountType === 'isolated_margin' && (
-				<StyledDiv>
-					<ArrowIcon />
-				</StyledDiv>
-			)}
+			<StyledDiv>
+				<ArrowIcon />
+			</StyledDiv>
 		</BridgeContainer>
 	)
 }
--- a/packages/app/src/pages/market.tsx
+++ b/packages/app/src/pages/market.tsx
@@ -153,7 +153,12 @@ function TradePanelDesktop() {
 		[accountType, isolatedPositionsCount]
 	)
 
-	if (walletAddress && !isL2 && openModal !== 'futures_smart_margin_socket') {
+	if (
+		walletAddress &&
+		!isL2 &&
+		openModal !== 'futures_smart_margin_socket' &&
+		openModal !== 'futures_isolated_transfer'
+	) {
 		return <FuturesUnsupportedNetwork />
 	}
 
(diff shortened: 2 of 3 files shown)


packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+	)
+
+	useEffect(() => {
+		switch (balanceStatus) {
+			case 'no_balance':
+			case 'low_balance':
+				setOpenSocket(true)
+				break
+			case 'high_balance':
+				setOpenSocket(false)
```
avclarke, line 83: I think socket can always be open as it's the only function of deposit here.

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+		}
+		const amtWei = wei(amount)
+		if (amtWei.eq(0) || amtWei.gt(susdBal) || (transferType === 0 && amtWei.lt(minDeposit))) {
+			return true
+		}
+		return false
+	}, [amount, susdBal, minDeposit, transferType, submitting])
+
+	const computedWithdrawAmount = useMemo(
+		() => (availableMargin.eq(wei(amount || 0)) ? availableMargin : wei(amount || 0)),
```
avclarke, line 99: I think we need to use withdrawableMargin?

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+		}
+	}, [transferType, susdBal, totalWithdrawable])
+
+	const onChangeTab = (selection: number) => {
+		setTransferType(selection)
+		setAmount('')
+	}
+
+	const onDeposit = () => {
+		dispatch(depositCrossMargin(wei(amount)))
```
avclarke, line 119: There's no need to deposit to smart margin, this should just be about accessing the socket bridge.

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+					{openSocket ? <SocketBridge /> : <Spacer height={20} />}
+				</>
+			)}
+			<BalanceContainer>
+				<BalanceText>{t('futures.market.trade.margin.modal.balance')}:</BalanceText>
+				<BalanceText>
+					<span>{formatDollars(susdBal)}</span> sUSD
+				</BalanceText>
+			</BalanceContainer>
+			<NumericInput
```
avclarke, line 162: We should only display this on withdraw and just show socket on deposit.

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+		() =>
+			availableMargin.gt(ZERO_WEI) || susdBalance.gt(minDeposit)
+				? 'high_balance'
+				: susdBalance.eq(ZERO_WEI)
+				? 'no_balance'
+				: 'low_balance',
+		[availableMargin, minDeposit, susdBalance]
+	)
+
+	useEffect(() => {
```
avclarke, line 76: I'm not sure we need to worry about this, can just keep it simple with withdraw / deposit (socket).

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
LeifuChen: Sure. Updated.

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
LeifuChen: Yes. It's unnecessary. Removed.

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
LeifuChen: True. Updated.

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
LeifuChen: Removed. We don't need it since the socket is always open now.

packages/app/src/sections/futures/Trade/TransferSmartMarginModal.tsx
LeifuChen: Good catch. It's an unnecessary check. The `withdraw` button should be disabled if the input amount is zero or the input amount is greater than the `withdrawableMargin`.

avclarke (approved):

PR #2687, opened 2023-07-27
Fix: Update the prefix of forex

LeifuChen:
## Description
Updated the prefix of `GBP`, `EUR` and `AUD` to `FX` instead of Crypto

## Related issue
#2680
#2683 

截屏2023-07-27 14 59 06

### DIFF
--- a/packages/app/src/queries/rates/useCandlesticksQuery.ts
+++ b/packages/app/src/queries/rates/useCandlesticksQuery.ts
@@ -6,8 +6,7 @@ import { getSupportedResolution } from 'components/TVChart/utils'
 import { DEFAULT_NETWORK_ID } from 'constants/defaults'
 import logError from 'utils/logError'
 
-import { DEFAULT_PYTH_TV_ENDPOINT } from './constants'
-import { AssetTypes } from './types'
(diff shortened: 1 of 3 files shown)


packages/app/src/queries/rates/useCandlesticksQuery.ts
```diff
@@ -18,8 +19,14 @@ export const requestCandlesticks = async (
 ) => {
 	const ratesEndpoint = getRatesEndpoint(networkId)
 	const pythTvEndpoint = DEFAULT_PYTH_TV_ENDPOINT
-	const metalAssets = ['XAU', 'XAG']
-	const prefix = metalAssets.includes(currencyKey!) ? 'Metal' : 'Crypto'
+	const nonCryptoTypes: AssetTypes = {
```
avclarke, line 22: I think we can lift this out of the component and maybe rename as `NON_CRYPTO_ASSET_TYPES`

avclarke (commented):
one minor other bit of clean up we could do

platschi (approved):

packages/app/src/queries/rates/useCandlesticksQuery.ts
LeifuChen: Good idea. Updated.

PR #2696, opened 2023-07-27
fix(sdk): use addresses from constants

insulineru:
In `ethcall@5` Contract instance has private address field. We were using this field, but now it returns `undefined`. Fixed this behavior.


avclarke (approved):
Nice work on the quick fix!

LeifuChen (approved):
It works. Thanks for the quick fix.

Issue #2672, opened 2023-07-25
✨ Implement swap deposit feature 


avclarke:
### Feature Proposal

Enable users to deposit to smart margin from assets other than sUSD by utilising the new swap function in smart margin accounts. 

### Implementation

- Users should be able to select which asset they wish to deposit from (We could just begin with popular stable coins)
- A users available balance should reflect balances from assets available to deposit
- To prepare the params for the deposit we should use the Uniswap api
- If we want to allow swap > deposit > trade we will need to set the `_amountOutMin` to ensure enough is returned to satisfy to the trade.

PR #2718, opened 2023-08-01
Swap deposits

koredefashokun:
## Description
This PR implements swap deposits, enabling users to deposit margin into their smart margin account from their USDC, USDT, DAI or LUSD balances.

## Related issue
#2672 

## Motivation and Context
This makes it easier to onboard users who do not have SUSD in their wallet.

## How Has This Been Tested?
N/A


packages/sdk/src/services/futures.ts
```diff
@@ -382,22 +391,40 @@ export default class FuturesService {
...
+			allowance,
+			usdcAllowance,
+			usdtAllowance,
+			daiAllowance,
+		] = await Promise.all([
 			smartMarginAccountContract.freeMargin(),
 			this.sdk.context.provider.getBalance(smartMarginAddress),
 			this.sdk.context.provider.getBalance(walletAddress),
 			SUSD.allowance(walletAddress, smartMarginAccountContract.address),
+			USDC.allowance(walletAddress, PERMIT2_ADDRESS),
```
avclarke, line 412: I think we need to convert this to multicall now we have this many requests.

PR #2741, opened 2023-08-03
Fix leverage check and sltp crash on mobile

avclarke:
## Description

- Fixes max leverage issue
- Fixes a crash when user has sl / tp on mobile market page


LeifuChen (approved):
Thanks for the quick fix.

PR #2744, opened 2023-08-04
Allow to close position when below 50 usd margin

avclarke:
## Description

Fixes a bug where users can't close a position when they have below $50 margin


LeifuChen (approved):
Thanks for the quick fix.

PR #2747, opened 2023-08-04
Fix max size button

avclarke:
## Description

Fixes an issue where clicking the max button on the trade panel was populating the input with zero.


LeifuChen (approved):
Thanks for the quick fix. The max button is working again.

PR #2773, opened 2023-09-03
export history to csv

### LINKED ISSUE #2764: feat: export account history into csv
### Subject of the issue
Add an option to account history to export data into a csv file (e.g. with `react-csv`)

Design (see Figma link)

### Deliverables
- Add header/subtitle according to Figma design
- Add Export button according to design for both dark and light theme
- Implement functionality to export csv for selected table (Positions or Trades)

### What to export?
Same data as in table, plus:

- include transaction hash on export as a line item (currently hidden behind click on the trade itself)

pldespaigne:
fix #2764

### DIFF
--- a/packages/app/src/sections/futures/TraderHistory.tsx
+++ b/packages/app/src/sections/futures/TraderHistory.tsx
@@ -15,8 +15,9 @@ import Table, { TableHeader, TableNoResults } from 'components/Table'
 import { Body } from 'components/Text'
 import ROUTES from 'constants/routes'
 import TimeDisplay from 'sections/futures/Trades/TimeDisplay'
+import { setCsvExportData } from 'state/futures/reducer'
 import { selectFuturesPositions, selectQueryStatuses } from 'state/futures/selectors'
-import { useAppSelector } from 'state/hooks'
+import { useAppDispatch, useAppSelector } from 'state/hooks'
 import { FetchStatus } from 'state/types'
 import { ExternalLink } from 'styles/common'
 import media from 'styles/media'
@@ -33,11 +34,12 @@ type TraderHistoryProps = {
 const TraderHistory: FC<TraderHistoryProps> = memo(
 	({ trader, traderEns, positionHistory, resetSelection, compact, searchTerm }) => {
 		const { t } = useTranslation()
+		const dispatch = useAppDispatch()
 		const positions = useAppSelector(selectFuturesPositions)
 		const { selectedTraderPositionHistory: queryStatus } = useAppSelector(selectQueryStatuses)
 
 		let data = useMemo(() => {
-			return positionHistory
+			const result = positionHistory
 				.sort((a, b) => b.timestamp - a.timestamp)
 				.map((stat, i) => {
 					const totalDeposit = stat.initialMargin.add(stat.totalDeposits)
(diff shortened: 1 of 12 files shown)


packages/app/src/state/wallet/selectors.ts
```diff
@@ -6,7 +6,8 @@ import type { RootState } from 'state/store'
 
 const SUPPORTED_NETWORKS = [1, 10, 5, 420]
 
-export const selectWallet = (state: RootState) => state.wallet.walletAddress ?? null
+export const selectWallet = (state: RootState) =>
+	'0xC2ecD777d06FFDF8B3179286BEabF52B67E9d991' ?? null
```
avclarke, line 10: Need to revert this

platschi, line 10: We need to revert this change for prod

platschi (changes requested):

packages/app/src/sections/futures/TraderHistory.tsx
```diff
@@ -73,7 +75,23 @@ const TraderHistory: FC<TraderHistoryProps> = memo(
...
+			result.forEach(
+				(row: (typeof result)[number]) =>
+					(csvData += `${new Date(row.timestamp).toISOString()},${row.marketShortName},${
+						row.status
+					},${row.trades},${row.totalVolume.toNumber()},${row.pnl.toNumber()},${row.pnlPct.slice(
+						1,
+						-2
+					)},${row.funding.toNumber()},${row.transactionHash}\n`)
+			)
+			dispatch(setCsvExportData(csvData))
```
avclarke, line 91: It would be nice if we could move all of this data formatting logic out of the component and into redux. So I think we could create a selector for the table data `selectPositionHistoryTableData` which we move the original table data formatting to and then `selectCsvExport` can just pull in the data from `selectPositionHistoryTableData` meaning we can also remove the `csvExport` field it's reducer.

avclarke (changes requested):
Looks good but with the added formatting introduced for the csv export I think it's a good time to move all this data formatting over to a redux selector and out of the component. This will also remove the need to add the `csvExport` field, we can just use a selector.

avclarke (commented):
Changes looks great, just a couple of clean up suggestions.

avclarke (approved):
Changes look good, nice work!

packages/app/src/sections/futures/TraderHistory.tsx
```diff
@@ -1,7 +1,5 @@
-import { ZERO_WEI } from '@kwenta/sdk/constants'
-import { FuturesPositionHistory } from '@kwenta/sdk/dist/types'
-import { getMarketName, MarketKeyByAsset } from '@kwenta/sdk/utils'
-import { wei, WeiSource } from '@synthetixio/wei'
+import { FuturesMarketKey, FuturesPositionHistory } from '@kwenta/sdk/dist/types'
```
avclarke, line 1:
```suggestion
import { FuturesMarketKey, FuturesPositionHistory } from '@kwenta/sdk/types'
```

packages/app/src/utils/futures.ts
```diff
@@ -1,3 +1,4 @@
+import { ZERO_WEI } from '@kwenta/sdk/dist/constants'
```
avclarke, line 1:
```suggestion
import { ZERO_WEI } from '@kwenta/sdk/constants'
```

avclarke (changes requested):
Just noticed there are a couple of places using `/dist` imports which cause build failures.

platschi (approved):

PR #2807, opened 2023-10-26
Custom display value for ETHBTC, so dropdown shows "ETH/BTC".

Ondkloss:
## Description
Added a custom handling of "ETHBTC" `getDisplayAsset` to return "ETH/BTC".
Value should appear in market dropdown.

## Related issue
None

## Motivation and Context
Easier to skim markets and identify this ratio market.

## How Has This Been Tested?
Crude testing in Chrome using local source override. Less than ideal.

Showing from and to:

Kwenta_From Kwenta_To


Ondkloss:
I can see from the fleek build that this is broken, apparently due to graph and 24h change dependence on this value.
Going for test environment to adapt.

PR #2826, opened 2024-02-02
feat: replaced getMarkets, getDailyVolumes on futures and getFuturesFee on kwentaToken

diamondyhand:


insulineru:
gm! 

That's a great idea. Can you come into kwenta discord and chat with us in the #dev-chat channel? It would be cool to synchronize

